PR #8, opened 2022-07-07
Issue 7 tests

skyerus:
Created to address [issue 7](https://github.com/open-feature/golang-sdk/issues/7)


skyerus:
Included https://github.com/golang/mock as a dependency to generate mocks of our interfaces for testing purposes. If including this is an issue we can perhaps just generate the mocks and not import it?

pkg/openfeature/provider.go
```diff
@@ -56,3 +56,8 @@ type NumberResolutionDetail struct {
 	Value float64
 	ResolutionDetail
 }
+
+// Metadata provides provider metadata
```
toddbaert, line 60: Might want to change this comment to `Metadata provides provider name`

pkg/openfeature/client_test.go
```diff
@@ -0,0 +1,149 @@
...
+		GetObjectValueDetails(flag string, defaultValue interface{}, evalCtx EvaluationContext, options ...EvaluationOption) (EvaluationDetails, error)
+	}
+
+	var clientI interface{} = client
+	if _, ok := clientI.(requirements); !ok {
+		t.Error("client returned by GetClient doesn't implement the 1.4.1 requirements interface")
+	}
+}
+
+// Requirement_1_4_2 has no suitable test as the provider implementation populates the EvaluationDetails value field
```
toddbaert, line 69: :+1:

pkg/openfeature/client_test.go
```diff
@@ -0,0 +1,149 @@
...
+// Requirement_1_4_5 has no suitable test as the provider implementation populates the EvaluationDetails variant field
+
+// Requirement_1_4_6 has no suitable test as the provider implementation populates the EvaluationDetails reason field
+
+// Requirement_1_4_7 has no suitable test as the provider implementation populates the EvaluationDetails error code field
+
+// Requirement_1_4_8 has no suitable test as the provider implementation populates the EvaluationDetails reason field
+
+// Requirement_1_4_9 is satisfied by the error included in the return signatures, as is idiomatic in Go. Errors aren't
+// fatal, the operations won't terminate as a result of an error.
```
toddbaert, line 141: It MAY be worth testing that in the case of abnormal execution of the provider, the default is returned as well as an error, but I don't feel too strongly about it.

pkg/openfeature/client_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+// TODO Requirement_1_4_10
+
+// TODO Requirement_1_4_11
```
toddbaert, line 145: I don't think we have to much worry about this one because of goroutines.

toddbaert (approved):
This looks good to me, thanks! Personally I would recommend waiting for an approval from @davejohnston or @thomaspoignant as well.

pkg/openfeature/provider.go
skyerus: Done

pkg/openfeature/client_test.go
skyerus: Done

pkg/openfeature/client_test.go
skyerus: Updated

PR #14, opened 2022-07-20
Hooks implementation

skyerus:
https://github.com/open-feature/golang-sdk/issues/9

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,7 @@ module github.com/open-feature/golang-sdk
 
 go 1.17
 
-require github.com/golang/mock v1.6.0
+require (
+	github.com/golang/mock v1.6.0
+	golang.org/x/text v0.3.7
+)
--- a/pkg/openfeature/client.go
+++ b/pkg/openfeature/client.go
@@ -1,6 +1,7 @@
 package openfeature
 
 import (
+	"errors"
 	"fmt"
 )
 
@@ -34,8 +35,8 @@ type Client struct {
 	hooks    []Hook
 }
 
-// GetClient returns a new Client. Name is a unique identifier for this client
-func GetClient(name string) *Client {
+// NewClient returns a new Client. Name is a unique identifier for this client
+func NewClient(name string) *Client {
 	return &Client{
 		metadata: ClientMetadata{name: name},
 		hooks:    []Hook{},
@@ -71,19 +72,46 @@ type EvaluationDetails struct {
 // GetBooleanValue return boolean evaluation for flag
 func (c Client) GetBooleanValue(flag string, defaultValue bool, evalCtx EvaluationContext, options EvaluationOptions) (bool, error) {
 	evalDetails, err := c.evaluate(flag, Boolean, defaultValue, evalCtx, options)
-	return evalDetails.Value.(bool), err
+	if err != nil {
+		return defaultValue, fmt.Errorf("evaluate: %w", err)
+	}
+
+	value, ok := evalDetails.Value.(bool)
+	if !ok {
+		return defaultValue, errors.New("evaluated value is not a boolean")
+	}
+
+	return value, nil
 }
 
 // GetStringValue return string evaluation for flag
 func (c Client) GetStringValue(flag string, defaultValue string, evalCtx EvaluationContext, options EvaluationOptions) (string, error) {
 	evalDetails, err := c.evaluate(flag, String, defaultValue, evalCtx, options)
-	return evalDetails.Value.(string), err
+	if err != nil {
+		return defaultValue, fmt.Errorf("evaluate: %w", err)
+	}
+
+	value, ok := evalDetails.Value.(string)
+	if !ok {
+		return defaultValue, errors.New("evaluated value is not a string")
+	}
+
+	return value, nil
 }
 
 // GetNumberValue return number evaluation for flag
 func (c Client) GetNumberValue(flag string, defaultValue float64, evalCtx EvaluationContext, options EvaluationOptions) (float64, error) {
 	evalDetails, err := c.evaluate(flag, Number, defaultValue, evalCtx, options)
-	return evalDetails.Value.(float64), err
+	if err != nil {
+		return defaultValue, fmt.Errorf("evaluate: %w", err)
+	}
+
+	value, ok := evalDetails.Value.(float64)
+	if !ok {
+		return defaultValue, errors.New("evaluated value is not a float64")
+	}
+
+	return value, nil
 }
 
 // GetObjectValue return object evaluation for flag
@@ -141,7 +169,7 @@ func (c Client) evaluate(
 	evalCtx, err = c.beforeHooks(hookCtx, apiClientInvocationHooks, evalCtx, options)
 	hookCtx.evaluationContext = evalCtx
 	if err != nil {
-		err = fmt.Errorf("failed to execute before hook: %w", err)
+		err = fmt.Errorf("execute before hook: %w", err)
 		c.errorHooks(hookCtx, invocationClientApiHooks, err, options)
 		return evalDetails, err
 	}
@@ -166,7 +194,7 @@ func (c Client) evaluate(
 
 	err = resolution.Error()
 	if err != nil {
-		err = fmt.Errorf("failed to evaluate the flag: %w", err)
+		err = fmt.Errorf("evaluate the flag: %w", err)
 		c.errorHooks(hookCtx, invocationClientApiHooks, err, options)
 		return evalDetails, err
 	}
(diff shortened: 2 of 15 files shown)


pkg/openfeature/hooks.go
```diff
@@ -3,11 +3,64 @@ package openfeature
 // Hook Hooks are a mechanism whereby application developers can add arbitrary behavior to flag evaluation. They operate similarly to middleware in many web frameworks.
 // https://github.com/open-feature/spec/blob/main/specification/flag-evaluation/hooks.md
```
james-milligan, line 4: this link seems to be broken

pkg/openfeature/client.go
```diff
@@ -65,126 +69,171 @@ type EvaluationDetails struct {
...
+func (c Client) GetNumberValueDetails(flag string, defaultValue float64, evalCtx EvaluationContext, options EvaluationOptions) (EvaluationDetails, error) {
+	return c.evaluate(flag, Number, defaultValue, evalCtx, options)
+}
+
+// GetObjectValueDetails return object evaluation for flag
+func (c Client) GetObjectValueDetails(flag string, defaultValue interface{}, evalCtx EvaluationContext, options EvaluationOptions) (EvaluationDetails, error) {
+	return c.evaluate(flag, Object, defaultValue, evalCtx, options)
+}
+
+func (c Client) evaluate(
```
james-milligan, line 115: I like this

pkg/openfeature/hooks.go
skyerus: updated

pkg/openfeature/client.go
```diff
@@ -27,14 +31,14 @@ func (cm ClientMetadata) Name() string {
 // Client implements the behaviour required of an openfeature client
 type Client struct {
 	metadata ClientMetadata
-	hooks []Hook
+	hooks    []Hook
 }
 
 // GetClient returns a new Client. Name is a unique identifier for this client
 func GetClient(name string) *Client {
```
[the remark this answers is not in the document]
skyerus: Updated. We should probably also update the naming of the various getters to removed the prefixed `Get` as is idiomatic in Go. I will create an issue for this

pkg/openfeature/client.go
```diff
@@ -65,126 +69,171 @@ type EvaluationDetails struct {
...
-	resolution := api.provider.GetBooleanEvaluation(flag, defaultValue, evalCtx, options...)
-	value := resolution.Value
-	err := resolution.Error()
-	if err != nil {
-		value = defaultValue
-	}
-	return value, err
+func (c Client) GetBooleanValue(flag string, defaultValue bool, evalCtx EvaluationContext, options EvaluationOptions) (bool, error) {
+	evalDetails, err := c.evaluate(flag, Boolean, defaultValue, evalCtx, options)
+	return evalDetails.Value.(bool), err
```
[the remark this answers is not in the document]
skyerus: Nice spot, updated

pkg/openfeature/client.go
```diff
@@ -65,126 +69,171 @@ type EvaluationDetails struct {
...
-// GetStringValueDetails return string evaluation for flag
-func (c Client) GetStringValueDetails(flag string, defaultValue string, evalCtx EvaluationContext, options ...EvaluationOption) (EvaluationDetails, error) {
-	resolution := api.provider.GetStringEvaluation(flag, defaultValue, evalCtx, options...)
-	value := resolution.Value
-	err := resolution.Error()
+	evalCtx, err = c.beforeHooks(hookCtx, apiClientInvocationHooks, evalCtx, options)
+	hookCtx.evaluationContext = evalCtx
 	if err != nil {
-		value = defaultValue
+		err = fmt.Errorf("failed to execute before hook: %w", err)
```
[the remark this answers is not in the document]
skyerus: Nice suggestion, updated

pkg/openfeature/hooks_test.go
```diff
@@ -0,0 +1,762 @@
...
+		flagType:          String,
+		defaultValue:      defaultValue,
+		clientMetadata:    client.metadata,
+		providerMetadata:  mockProvider.Metadata(),
+		evaluationContext: evalCtx,
+	}
+	hook1EvalCtxResult := &EvaluationContext{TargetingKey: "mockHook1"}
+	mockHook1.EXPECT().Before(hook1Ctx, gomock.Any()).Return(hook1EvalCtxResult, nil)
+
+	// assert that the evaluation context returned by the first hook is passed into the second hook
```
toddbaert, line 205: These comments on the primary assertion are appreciated.

pkg/openfeature/client.go
```diff
@@ -65,126 +70,198 @@ type EvaluationDetails struct {
...
+		res := api.provider.GetBooleanEvaluation(flag, defValue, evalCtx, options)
+		resolution = res.ResolutionDetail
+	case String:
+		defValue := defaultValue.(string)
+		res := api.provider.GetStringEvaluation(flag, defValue, evalCtx, options)
+		resolution = res.ResolutionDetail
+	case Number:
+		defValue := defaultValue.(float64)
+		res := api.provider.GetNumberEvaluation(flag, defValue, evalCtx, options)
+		resolution = res.ResolutionDetail
```
toddbaert, line 192: Because the caller already "knows" the type, Instead of a switch case here, I _think_ you can use a reference to the appropriate provider method passed from the caller. I did something similar in JS: https://github.com/open-feature/node-sdk/blob/66ed1b1a56e5fa1137ff66198ac245c5c8496347/src/client.ts#L146

The only thing that might be a challenge is defining the receiver function type generally and accurately without generics since they have different params. If you have to use `interface{}` for the function, it might not be worth it.

pkg/openfeature/hooks_test.go
```diff
@@ -0,0 +1,762 @@
...
+		t.Errorf("field %s doesn't exist in the HookHints struct", fieldName)
+	}
+
+	caser := cases.Title(language.English)
+	if caser.String(fieldName) == fieldName {
+		t.Errorf("field %s is uppercased and therefore mutable", fieldName)
+	}
+}
+
+func TestRequirement_4_2_2_2(t *testing.T) {
```
toddbaert, line 95: This might be tedious, but we might want to create some wrapper functions for tests or something that allow us to include the overall idea of the spec point (some text) instead of just the number. I only think this is important because the actual requirement represented by the numbers _could_ shift so it will help us keep things sane.

Probably something for another PR.

toddbaert (approved):
Nice work!

pkg/openfeature/client.go
```diff
@@ -27,14 +32,14 @@ func (cm ClientMetadata) Name() string {
 // Client implements the behaviour required of an openfeature client
 type Client struct {
 	metadata ClientMetadata
-	hooks []Hook
+	hooks    []Hook
 }
 
-// GetClient returns a new Client. Name is a unique identifier for this client
-func GetClient(name string) *Client {
+// NewClient returns a new Client. Name is a unique identifier for this client
```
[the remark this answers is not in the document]
skyerus: Exactly that, `New` is idiomatic in Golang

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/golang/mock v1.6.0 // indirect
 	golang.org/x/mod v0.4.2 // indirect
 	golang.org/x/sys v0.0.0-20210510120138-977fb7262007 // indirect
+	golang.org/x/text v0.3.7 // indirect
 	golang.org/x/tools v0.1.1 // indirect
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
```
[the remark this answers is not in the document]
skyerus: https://github.com/open-feature/golang-sdk/pull/21 tidies them up but I'll do the same here

pkg/openfeature/client.go
skyerus: Without generics the only way that approach would work is if the provider functions all had the exact same signature. Using `interface{}` as a match all wouldn't work in this sense, the provider functions themselves would have to define their parameters as `interface{}`

pkg/openfeature/hooks_test.go
skyerus: I like this as an idea, I guess we could add the specification requirement as a comment to the test. To avoid the issue created by numbers shifting we could tie the sdk to a version of the spec.

pkg/openfeature/client.go
toddbaert: Alas!

pkg/openfeature/hooks_test.go
skyerus: Created [an issue](https://github.com/open-feature/golang-sdk/issues/27) for this.

PR #15, opened 2022-07-20
Github CI build workflow

skyerus:
https://github.com/open-feature/golang-sdk/issues/13

Checked locally using https://github.com/nektos/act


.github/workflows/build.yaml
```diff
@@ -0,0 +1,40 @@
+name: build
```
beeme1mr, line 1: I would change the name to `ci` or something similar.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,40 @@
...
+    branches:
+      - main
+
+jobs:
+
+  lint:
+    runs-on: ubuntu-latest
+    env:
+      GOPATH: /home/runner/work/open-feature/golang-sdk
+      GOBIN: /home/runner/work/open-feature/golang-sdk/bin
```
beeme1mr, line 18: I'm not sure how consistent the runner path is. It may be worth considering doing it the same way OTel does. You can see their workflow [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/.github/workflows/ci.yml#L22).

.github/workflows/build.yaml
```diff
@@ -0,0 +1,40 @@
+name: build
+
+on:
+  push:
+    branches:
+      - feature/workflows
```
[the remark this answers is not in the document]
skyerus: I shamelessly took inspiration from https://github.com/open-feature/flagd/blob/main/.github/workflows/build.yaml#L6

.github/workflows/build.yaml
skyerus: Done, thanks for the tip

.github/workflows/build.yaml
skyerus: Done

pkg/openfeature/openfeature_test.go
```diff
@@ -61,10 +62,10 @@ func TestRequirement_1_1_5(t *testing.T) {
 
 func TestRequirement_1_1_6(t *testing.T) {
 	defer t.Cleanup(initSingleton)
-	type clientCreationFunc func(name string)*Client
+	type clientCreationFunc func(name string) *Client
 
-	var f clientCreationFunc
-	f = GetClient // asserting that our GetClient method matches this signature is enough to deduce that no error is returned
+	// asserting that our GetClient method matches this signature is enough to deduce that no error is returned
+	var f clientCreationFunc = GetClient
```
skyerus, line 68: fixed lint error

beeme1mr (approved):

PR #21, opened 2022-07-22
Automate the creation of a release after tag push

skyerus:


.goreleaser.yaml
```diff
@@ -0,0 +1,30 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
+# Make sure to check the documentation at https://goreleaser.com
```
toddbaert, line 2: Can you remove this comment?

.goreleaser.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - replacements:
+      darwin: Darwin
+      linux: Linux
+      windows: Windows
+      386: i386
+      amd64: x86_64
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
```
toddbaert, line 22: Does this do anything in our case, if we're only running this on `v*` tags?

toddbaert (approved):

.goreleaser.yaml
skyerus: Removed

.goreleaser.yaml
skyerus: Done

PR #35, opened 2022-08-02
Renamed Number retrieval to Float. Added func for retrieval of int64

skyerus:


skyerus:
Awaiting similar changes to be ready in flagd/flagd-provider @james-milligan

davejohnston (approved):
This looks good.  I wonder if we should also just support Number?  I always read the types in the spec as 'JSON' types rather than language primitives, though for users they likely just want access to an int or a float.

skyerus:
> This looks good. I wonder if we should also just support Number? I always read the types in the spec as 'JSON' types rather than language primitives, though for users they likely just want access to an int or a float.

Would `Number` not also just be returning a `float64`?

PR #42, opened 2022-08-09
Implemented evaluation context as defined in 0.2.0 spec

skyerus:


james-milligan (approved):

beeme1mr (approved):

pkg/openfeature/client.go
```diff
@@ -9,6 +9,8 @@ import (
 type IClient interface {
 	Metadata() ClientMetadata
 	AddHooks(hooks ...Hook)
+	SetEvaluationContext(evalCtx EvaluationContext)
+	EvaluationContext() EvaluationContext
```
toddbaert, line 13: Some other SDKs don't allow the setting of evaluation context directly on the client, but instead take it as a param to the `getClient` method on the top-level API.

Up to you which you think is better for the go SDK.

pkg/openfeature/client.go
```diff
@@ -55,6 +59,16 @@ func (c *Client) AddHooks(hooks ...Hook) {
 	c.hooks = append(c.hooks, hooks...)
 }
 
+// SetEvaluationContext sets the client's evaluation context
+func (c *Client) SetEvaluationContext(evalCtx EvaluationContext) {
+	c.evaluationContext = evalCtx
+}
```
toddbaert, line 65: see https://github.com/open-feature/golang-sdk/pull/42/files#r947913435

toddbaert (approved):

pkg/openfeature/client.go
skyerus: I think this allows for more configurability and it keeps the client constructor terse. Thoughts @james-milligan ?

james-milligan: I do nto have any strong opinions here, however im leaning towards @skyerus's changes, it keeps things declarative and prevents a user needing to pass in an empty `EvaluationContext` into `getClient` when one is not needed.

PR #45, opened 2022-08-17
Context ordering

skyerus:


skyerus:
Draft as dependent on https://github.com/open-feature/golang-sdk/pull/42

pkg/openfeature/client.go
```diff
@@ -304,6 +304,10 @@ func mergeContexts(evaluationContexts ...EvaluationContext) EvaluationContext {
 	mergedCtx := evaluationContexts[0]
 
 	for i := 1; i < len(evaluationContexts); i++ {
+		if mergedCtx.TargetingKey == "" && evaluationContexts[i].TargetingKey != "" {
+			mergedCtx.TargetingKey = evaluationContexts[i].TargetingKey
```
james-milligan, line 308: minor, but could add a break in here

skyerus: we want to keep merging the attributes, a break would end that

james-milligan: ah!

james-milligan (approved):

PR #51, opened 2022-08-30
refactor!: flatten evaluationContext object

james-milligan:
Refactor to flatten evaluation context before the struct is passed to providers. This brings the golang-sdk more inline with others.

https://github.com/open-feature/golang-sdk/issues/30

 I've opted to not introduce another type to wrap the `map[string]interface{}` to reduce confusion, but as the flattened context is returned by a function this would be very simple to implement.


pkg/openfeature/client.go
```diff
@@ -254,6 +255,17 @@ func (c Client) evaluate(
 	return evalDetails, nil
 }
 
+func flattenContext(evalCtx EvaluationContext) map[string]interface{} {
+	flatCtx := map[string]interface{}{}
+	if evalCtx.Attributes != nil {
+		flatCtx = evalCtx.Attributes
+	}
+	if evalCtx.TargetingKey != "" {
+		flatCtx["targetingKey"] = evalCtx.TargetingKey
```
skyerus, line 264: can we create a constant for `"targetingKey"`

pkg/openfeature/client.go
```diff
@@ -5,6 +5,10 @@ import (
 	"fmt"
 )
 
+const (
+	TARGETING_KEY string = "targetingKey" // flattened evaluation context key for the TargetingKey field
```
skyerus, line 9: camelcase is the convention in Go: `TargetingKey`

pkg/openfeature/client.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 const (
-	TARGETING_KEY string = "targetingKey" // flattened evaluation context key for the TargetingKey field
+	TARGETING_KEY string = "TargetingKey" // flattened evaluation context key for the TargetingKey field
```
skyerus, line 9: I meant the naming of the constant, `"targetingKey"` was correct for the value

skyerus (approved):

beeme1mr:
This looks like a good change to me. @toddbaert, could you please review this when you have a moment?

beeme1mr (changes requested):
Let's wait to merge this until the the release process has been configured. I should have a PR ready shortly.

beeme1mr (approved):
Release please have been configured.

pkg/openfeature/client.go
```diff
@@ -5,6 +5,10 @@ import (
 	"fmt"
 )
 
+const (
+	TargetingKey string = "targetingKey" // flattened evaluation context key for the TargetingKey field
```
toddbaert, line 9: I find this comment a bit confusing, Since it's exported we might want to explain (very high level) what the `TargetingKey` is.

toddbaert, line 9: We also may want to specifically reference in some tests, to make sure we "see" breaking changes around it. I hope that makes sense.

james-milligan: I've updated this comment, pulled a line from the spec - let me know if it is suitable

james-milligan: I've added a unit test to check for happy path plus a few other scenarios I could think of

beeme1mr:
@toddbaert, please review this again when you have a moment. If it looks good, please approve and merge.

toddbaert (approved):

PR #53, opened 2022-08-31
Updated HookContext to be passed as a pointer

james-milligan:
Currently each method (`Before`, `After`, `Finally` and `Error`) is called on a copy of any given 'Hook' struct, with the `hookCtx` argument being passed by value. This means that across the lifecycle of a hook there is no opportunity for persisting state between these methods, excluding overloading the `EvaluationContext`.

An example of where this persistence is important is in the implementation of the open-telemetry hook. For this hook to work a `span` needs to be created in the before hook, and attributes set in any of the downstream methods. With the current implementation this  is not possible.

This draft PR contains a proposed approach for allowing state to be carried across the hooks lifecycle.
`HookContext` is now passed as a pointer and contains the `Store` field (Would appreciate some input on this name), this means that all `Hooks` now have access to a single shared `map[string]interface{}`.

Another approach that could be taken to achieve the same outcome is for each Hook to be treated as a singleton, at which point state can be managed internally.

### DIFF
--- a/pkg/openfeature/hooks.go
+++ b/pkg/openfeature/hooks.go
@@ -4,10 +4,10 @@ package openfeature
 // They operate similarly to middleware in many web frameworks.
 // https://github.com/open-feature/spec/blob/main/specification/hooks.md
 type Hook interface {
-	Before(hookContext HookContext, hookHints HookHints) (*EvaluationContext, error)
-	After(hookContext HookContext, flagEvaluationDetails EvaluationDetails, hookHints HookHints) error
-	Error(hookContext HookContext, err error, hookHints HookHints)
-	Finally(hookContext HookContext, hookHints HookHints)
+	Before(hookContext *HookContext, hookHints HookHints) (*EvaluationContext, error)
+	After(hookContext *HookContext, flagEvaluationDetails EvaluationDetails, hookHints HookHints) error
+	Error(hookContext *HookContext, err error, hookHints HookHints)
+	Finally(hookContext *HookContext, hookHints HookHints)
 }
 
 // HookHints contains a map of hints for hooks
@@ -34,6 +34,9 @@ type HookContext struct {
 	clientMetadata    ClientMetadata
 	providerMetadata  Metadata
 	evaluationContext EvaluationContext
+	// Store provides a persistent state across Hook methods (Before, After, Finally, Error).
+	// This is shared between all hooks so it is strongly advised to use a hook specific prefix.
+	Store map[string]interface{}
 }
 
 // FlagKey returns the hook context's flag key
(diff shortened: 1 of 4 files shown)


beeme1mr:
> Another approach that could be taken to achieve the same outcome is for each Hook to be treated as a singleton, at which point state can be managed internally.

How difficult would this be to implement and would there be any overhead concerns? Having this isolation would be nice but not at the expensive of performance in my opinion.

beeme1mr:
Also, please remember to update the flagD provider when this is merged.

pkg/openfeature/hooks.go
```diff
@@ -34,6 +34,9 @@ type HookContext struct {
 	clientMetadata    ClientMetadata
 	providerMetadata  Metadata
 	evaluationContext EvaluationContext
+	// Store provides a persistent state across Hook methods (Before, After, Finally, Error).
+	// This is shared between all hooks so it is strongly advised to use a hook specific prefix.
+	Store map[string]interface{}
```
toddbaert, line 39: Can we avoid creating a specific store in the SDK itself? I'm not completely opposed to it, but I prefer the idea of leaving it to the hooks themselves to store their own data. This leaves the SDK out of it, and also prevents the sort of collisions you're alluding to here.

Could a hook feature a sort of naive weakmap implementation where the pointer address of the HookContext is used as the key, and where the associated value removed it after it's GC'd? Is there a reasonable way to do this without risking a memory leak?

If that's not possible, I'm OK with this option.

toddbaert:
> Also, please remember to update the flagD provider when this is merged.

I'm not 100% sure, but I think since we're de-referencing the hookContext, no changes in the provider are required: https://github.com/open-feature/golang-sdk/pull/53/files#diff-1520c2162931fb780e1961a5cd73b7a724cbbbecea232fccb9384db38f4957c7R209

james-milligan:
> > Another approach that could be taken to achieve the same outcome is for each Hook to be treated as a singleton, at which point state can be managed internally.
> 
> How difficult would this be to implement and would there be any overhead concerns? Having this isolation would be nice but not at the expensive of performance in my opinion.

ill take a look into this tomorrow - I've been thinking through the solution and this would be a favourable approach, it would certainly make for a better development experience. I don't think it should be too difficult.

pkg/openfeature/hooks.go
james-milligan: I agree, I was thinking over this option and going down the singleton hook approach would be the cleanest, both in terms of garbage collection and development experience.

skyerus:
A developer could implement their hook like this:
```go
type ExampleHook struct {
    State map[string]interface{}
}

func (h *ExampleHook) Before(hookContext HookContext, hookHints HookHints) (*EvaluationContext, error) {
     h.State["flagKey"] = hookContext.FlagKey()
     return nil, nil
}

```

then pass it by reference and it will always use the same hook instance (therefore the same state).

Is this an option? Perhaps I've misunderstood the problem.

james-milligan:
you're right - apologies this was a misunderstanding on my behalf

PR #55, opened 2022-09-02
feat: introduce UnimplementedHook to avoid authors having to define empty functions

skyerus:


pkg/openfeature/hooks.go
```diff
@@ -65,3 +65,26 @@ func (h HookContext) ProviderMetadata() Metadata {
...
+// UnimplementedHook implements all hook methods with empty functions
+// Include UnimplementedHook in your hook struct to avoid defining empty functions
+// e.g.
+// type MyHook struct {
+//   UnimplementedHook
+// }
+type UnimplementedHook struct{}
+
+func (u UnimplementedHook) Before(hookContext HookContext, hookHints HookHints) (*EvaluationContext, error) {
+	return nil, nil
```
james-milligan, line 81: I could be wrong - but will this set the `EvaluationContext` to `nil` before it is used for flag evaluation?

beeme1mr: Could you add a test for this?

skyerus: Incorrect. The resultant context is only used if not `nil`
Will add a test

james-milligan:
looks good!

james-milligan (approved):

pkg/openfeature/hooks.go
skyerus: Added test

PR #58, opened 2022-09-13
fix!: remove duplicate Value field from ResolutionDetail structs

### LINKED ISSUE #57: [BUG] Typed ResolutionDetail structs have mulitple different `Value` values
### Observed behavior

Looking at https://github.com/open-feature/golang-sdk/blob/main/pkg/openfeature/provider.go#L65-L68 and running the code, it is possible for a provider to return a struct for `BoolResolutionDetail` where there are two different `Value` elements, with different values.

eg
image

It's been a while since I've done go coding, and this may be idiomatic, but to me it's confusing

### Expected Behavior

You should only be able to set one `Value` value

### Steps to reproduce

```go
	thing := openfeature.BoolResolutionDetail{
		Value: true,
		ResolutionDetail: openfeature.ResolutionDetail{
			Value: false,
		},
	}
```

skyerus:
Fixes #57

### DIFF
--- a/pkg/openfeature/provider.go
+++ b/pkg/openfeature/provider.go
@@ -20,7 +20,7 @@ type FeatureProvider interface {
 	StringEvaluation(flag string, defaultValue string, evalCtx map[string]interface{}) StringResolutionDetail
 	FloatEvaluation(flag string, defaultValue float64, evalCtx map[string]interface{}) FloatResolutionDetail
 	IntEvaluation(flag string, defaultValue int64, evalCtx map[string]interface{}) IntResolutionDetail
-	ObjectEvaluation(flag string, defaultValue interface{}, evalCtx map[string]interface{}) ResolutionDetail
+	ObjectEvaluation(flag string, defaultValue interface{}, evalCtx map[string]interface{}) InterfaceResolutionDetail
 	Hooks() []Hook
 }
 
@@ -30,7 +30,6 @@ type FeatureProvider interface {
 // N.B we could use generics but to support older versions of golang for now we will have type specific resolution
 // detail
 type ResolutionDetail struct {
-	Value     interface{}
 	ErrorCode string
 	Reason    string
 	Variant   string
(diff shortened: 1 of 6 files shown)


pkg/openfeature/provider.go
```diff
@@ -16,10 +16,10 @@ const (
...
 type FeatureProvider interface {
 	Metadata() Metadata
-	BooleanEvaluation(flag string, defaultValue bool, evalCtx map[string]interface{}) BoolResolutionDetail
-	StringEvaluation(flag string, defaultValue string, evalCtx map[string]interface{}) StringResolutionDetail
-	FloatEvaluation(flag string, defaultValue float64, evalCtx map[string]interface{}) FloatResolutionDetail
-	IntEvaluation(flag string, defaultValue int64, evalCtx map[string]interface{}) IntResolutionDetail
+	BooleanEvaluation(flag string, defaultValue bool, evalCtx map[string]interface{}) ResolutionDetail
+	StringEvaluation(flag string, defaultValue string, evalCtx map[string]interface{}) ResolutionDetail
+	FloatEvaluation(flag string, defaultValue float64, evalCtx map[string]interface{}) ResolutionDetail
+	IntEvaluation(flag string, defaultValue int64, evalCtx map[string]interface{}) ResolutionDetail
```
agentgonzo, line 22: It was nice having the typed ResolutionDetail where the `Value` was typed. That way, when you called `getBooleanValue(...)` (or getBooleanDetail) you got a `boolean` value back, rather than a generic `interface{}`. It would be nice if we could keep this.

Wolud a possible fix be to make `ResolutionDetail` abstract (if that's even a thing in go) and remove the `Value: interface{}` from there, and keep the typed version in BoolResolutionValue etc?

skyerus: Good suggestion, done :)

james-milligan (approved):

agentgonzo (approved):

PR #67, opened 2022-09-16
fix: return error code from client given by provider

rgrassian-split:
[Requirement 147 of the spec says that the error code must be provided upon abnormal execution](https://docs.openfeature.dev/docs/specification/sections/flag-evaluation#requirement-147). The current setup returns an `EvaluationDetails` with no error code, regardless of what error code the Provider returned.
This populates the `ResolutionDetail` of the return value with the error code, reason, and variant values returned by the provider.
Since this just populates the `ResolutionDetail` and not the `InterfaceResolutionDetail`, the default treatment will remain the value regardless of what the provider returned.

### DIFF
--- a/pkg/openfeature/client.go
+++ b/pkg/openfeature/client.go
@@ -303,6 +303,7 @@ func (c Client) evaluate(
 		)
 		err = fmt.Errorf("error code: %w", err)
 		c.errorHooks(hookCtx, providerInvocationClientApiHooks, err, options)
+		evalDetails.InterfaceResolutionDetail.ResolutionDetail = resolution.ResolutionDetail
 		return evalDetails, err
 	}
 	if resolution.Value != nil {


beeme1mr:
Thanks @rgrassian-split! Would you be able to add a test for this?

Could you also prefix the PR with "fix: " so the lint pr action is happy?

rgrassian-split:
> Thanks @rgrassian-split! Would you be able to add a test for this?
> 
> Could you also prefix the PR with "fix: " so the lint pr action is happy?

@beeme1mr We would need to mock the behavior of the NoOpProvider to return an error. I don't have a ton of experience in Go, is there a preferred way to go about mocking for this sdk? If not, the easiest way I see is to create a error_noop_provider.go and use this instead, but requires another file. Thoughts?

beeme1mr:
@rgrassian-split, no problem! I'll create an issue to add a test.

skyerus (approved):
Thanks @rgrassian-split, nice spot!
I think this was missed as in Go the idiomatic way of error handling is to check whether `err != nil` and handle `err` directly rather than looking through the `EvaluationDetails`. This change allows consumers the option to do either which is nice

james-milligan (approved):
Nice

Issue #69, opened 2022-09-16
Provider context type clarification


beeme1mr:
The `FeatureProvider` interface uses a `map[string]interface{}` instead of a `EvaluationContext`. Was this done intentionally?

This question was asked by @rgrassian-split on [Slack](https://cloud-native.slack.com/archives/C0344AANLA1/p1663356409146239).

james-milligan:
Changes were made following [this issue ](https://github.com/open-feature/golang-sdk/issues/30) happy to reopen

james-milligan:
I will add comments to references to make this change seem more intentional

beeme1mr:
@rgrassian-split, does this make sense to you? Do you have any concerns?

rgrassian-split:
@beeme1mr I softly disagree and think the provider having to look up the exact string to fetch the targeting key from the map is just as troublesome as / worse than having to make an additional fetch to attributes before accessing custom attributes. It's just a soft disagree though, overall it's pretty minor and fine with either.

james-milligan:
One benefit to decoupling the 2 types (context passed by the client, and context passed to a provider) is that we can make changes to the client facing api without requiring all providers to be updated with the new types, although this does nto necessarily have to come in the form of the `FlattenedContext`, and of course if only relevant if the `EvaluationContext` type is subject to change

PR #71, opened 2022-09-20
update context_type

### LINKED ISSUE #30: Evaluation Context Handling
In previous versions of the Golang-sdk `EvaluationContext` has been defined as an `interface{}`. In [this](https://github.com/open-feature/golang-sdk/commit/f0312d81766b1ac9849d4f3c31acf450b452e553) commit it has been updated to a `struct` containing required fields (currently just `TargetingKey` but likely to grow) with a nested `map[string]interface{}` under the `Attributes` field, to better comply to the spec.

I believe this may cause issues down the line for client applications needing to know if their properties are on the base struct or in the additional properties.
```
if evCtx.Yellow == true {
    return 'yellow'
}
```
turns into 
```
if evCtx.Attributes.Yellow == true {
    return 'yellow'
}
```

Following conversations with @skyerus we have identified 3 possible ways to resolve this, though this list likely isn't exhaustive:

1. Do nothing and handle the nesting within providers
2. Flatten the structure into a `map[string]interface{}` as standard and update the [provider interface](https://github.com/open-feature/golang-sdk/blob/91d68f3ffa9f698ffc313f2da3858e6ea33befb1/pkg/openfeature/provider.go#L15) method signatures to reflect this
3. Offer a `before hook` that will flatten the structure ([would require provider hooks](https://github.com/open-feature/research/pull/22)), similar outcome to doing nothing, but the hook would be directly linked to the sdk

Please let me know your thoughts

EDIT by @toddbaert :

Seems like we have a consensus, I will take this issue and action it.

Akshit42-hue:
Signed-off-by: Akshit42-hue <patelakshit2025@gmail.com>
Fixes #30

### DIFF
--- a/pkg/openfeature/provider_test.go
+++ b/pkg/openfeature/provider_test.go
@@ -41,11 +41,11 @@ func TestRequirement_2_2(t *testing.T) {
 	mockProvider := NewMockFeatureProvider(ctrl)
 
 	type requirements interface {
-		BooleanEvaluation(flag string, defaultValue bool, evalCtx map[string]interface{}) BoolResolutionDetail
-		StringEvaluation(flag string, defaultValue string, evalCtx map[string]interface{}) StringResolutionDetail
-		FloatEvaluation(flag string, defaultValue float64, evalCtx map[string]interface{}) FloatResolutionDetail
-		IntEvaluation(flag string, defaultValue int64, evalCtx map[string]interface{}) IntResolutionDetail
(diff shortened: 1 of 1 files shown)


Akshit42-hue:
cc @beeme1mr @james-milligan

beeme1mr:
Hi @Akshit42-hue, thanks for your contribution. In this case, the current interface is correct based on [this discussion](https://github.com/open-feature/go-sdk/issues/30). If you have any questions, please comment on [this issue](https://github.com/open-feature/go-sdk/issues/69). Thanks again!

PR #74, opened 2022-09-22
refactor!: defined type for provider interface evaluation context

james-milligan:
Signed-off-by: James-Milligan <james@omnant.co.uk>

### DIFF
--- a/pkg/openfeature/provider.go
+++ b/pkg/openfeature/provider.go
@@ -12,19 +12,19 @@ const (
 	TargetingKey string = "targetingKey" // evaluation context map key. The targeting key uniquely identifies the subject (end-user, or client service) of a flag evaluation.
 )
 
-// TransformedContext contains metadata for a given flag evaluation in a flattened structure.
+// FlattenedContext contains metadata for a given flag evaluation in a flattened structure.
 // TargetingKey ("targetingKey") is stored as a string value if provided in the evaluation context.
-type TransformedContext map[string]interface{}
+type FlattenedContext map[string]interface{}
 
 // FeatureProvider interface defines a set of functions that can be called in order to evaluate a flag.
 // vendors should implement
 type FeatureProvider interface {
 	Metadata() Metadata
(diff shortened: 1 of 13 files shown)


james-milligan:
Intention is to reduice confusion and ensure the change looks more deliberate

pkg/openfeature/provider.go
```diff
@@ -12,15 +12,17 @@ const (
 	TargetingKey string = "targetingKey" // evaluation context map key. The targeting key uniquely identifies the subject (end-user, or client service) of a flag evaluation.
 )
 
+type TransformedContext map[string]interface{}
```
skyerus, line 15: Consider adding a comment explaining how the targeting key is stored within this map

beeme1mr (approved):

pkg/openfeature/provider.go
toddbaert: Nitpick:

`transformedContext` is used in a lot of SDKs as a variable name internally because it was the context provided after the `contextTransformer` function ran.

I'm fine with it here, but I wonder if `FlattenedContext` might be a bit more helpful.

pkg/openfeature/provider.go
```diff
@@ -12,15 +12,19 @@ const (
 	TargetingKey string = "targetingKey" // evaluation context map key. The targeting key uniquely identifies the subject (end-user, or client service) of a flag evaluation.
 )
 
+// TransformedContext contains metadata for a given flag evaluation in a flattened structure.
+// TargetingKey ("targetingKey") is stored as a string value if provided in the evaluation context.
+type TransformedContext map[string]interface{}
```
toddbaert, line 17: Nitpick:

transformedContext is used in a lot of SDKs as a variable name internally because it was the context provided after the contextTransformer function ran.

I'm fine with it here, but I wonder if FlattenedContext might be a bit more helpful.

james-milligan: updated

skyerus (approved):

PR #75, opened 2022-09-22
feat!: introduced context.Context to client and provider api

skyerus:
Signed-off-by: Skye Gill <gill.skye95@gmail.com>

### DIFF
--- a/pkg/openfeature/client.go
+++ b/pkg/openfeature/client.go
@@ -1,6 +1,7 @@
 package openfeature
 
 import (
+	"context"
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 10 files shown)


markphelps (approved):
lgtm 🚀

pkg/openfeature/client.go
```diff
@@ -13,16 +14,16 @@ type IClient interface {
...
-	FloatValue(flag string, defaultValue float64, evalCtx EvaluationContext, options EvaluationOptions) (float64, error)
-	IntValue(flag string, defaultValue int64, evalCtx EvaluationContext, options EvaluationOptions) (int64, error)
-	ObjectValue(flag string, defaultValue interface{}, evalCtx EvaluationContext, options EvaluationOptions) (interface{}, error)
-	BooleanValueDetails(flag string, defaultValue bool, evalCtx EvaluationContext, options EvaluationOptions) (EvaluationDetails, error)
-	StringValueDetails(flag string, defaultValue string, evalCtx EvaluationContext, options EvaluationOptions) (EvaluationDetails, error)
-	FloatValueDetails(flag string, defaultValue float64, evalCtx EvaluationContext, options EvaluationOptions) (EvaluationDetails, error)
-	IntValueDetails(flag string, defaultValue int64, evalCtx EvaluationContext, options EvaluationOptions) (EvaluationDetails, error)
-	ObjectValueDetails(flag string, defaultValue interface{}, evalCtx EvaluationContext, options EvaluationOptions) (EvaluationDetails, error)
+	BooleanValue(ctx context.Context, flag string, defaultValue bool, evalCtx EvaluationContext, options EvaluationOptions) (bool, error)
```
beeme1mr, line 17: It may be confusing to have both `ctx` and `evalCtx`. I think we need to clearly document the purpose of each argument.

beeme1mr (commented):
it would be nice to have additional documentation describing the value `context.Context` provides. The code itself looks great.

toddbaert:
I just want to chime in and say I think this is a good change, as evidenced by the fact some of vendors take the context as well, so it will make integration with them easier.

pkg/openfeature/client.go
skyerus: Added a description of each parameter to the comment above each func (similar approach to https://github.com/open-feature/js-sdk/blob/main/src/types.ts#L56)

beeme1mr (approved):
Looks good, thanks!

PR #85, opened 2022-10-03
Client uses value returned by provider

rgrassian-split:
Signed-off-by: Robert Grassian <robert.grassian@split.io>

## This PR

- Sets the Evaluation Detail's `Value` to be the returned value from the provider.

### How to test

Added test


beeme1mr (approved):

skyerus (approved):
nice catch

PR #91, opened 2022-10-04
feat!: made EvaluationContext fields unexported with a constructor and setters to enforce immutability

skyerus:


james-milligan (approved):
Nice work

skyerus:
> This looks good.
> 
> I think the only thing I'm missing here is the lock on global providers and hooks, as well as client hooks. We probably want to read-lock all of these during evaluation, and write lock them during mutation.

Great points. I've tackled this in another PR: https://github.com/open-feature/go-sdk/pull/93

Issue #90, opened 2022-10-04
[FEATURE] Ensure thread safety of sdk


skyerus:
### Requirements

Inspired by https://github.com/open-feature/dotnet-sdk/issues/56

The global singletons (e.g. provider, evaluationContext, logger) all share a mutex to ensure thread safety, however a client doesn't use a mutex so there's a concern if two processes want to set an evaluation context for example.

The `EvaluationContext` type has the field `Attributes` of type `map[string]interface{}`. In go, maps are always passed by reference, this means that in theory one could pass an `EvaluationContext` to a client's flag evaluation call and continue to alter the `Attributes`' map in another process.
This could be avoided by making the `Attributes` field unexported with an exported constructor for `EvaluationContext`, this would mean that an application author couldn't change the field once constructed. However, an author could still alter the map they passed as a parameter to the constructor, which is the same map used in struct (passed by reference). In order to avoid this the constructor would need to initialise a new map and copy the map passed as a parameter.

Is this overkill? Is there a better solution we can come up with?

cc @beeme1mr @toddbaert @james-milligan

james-milligan:
i personally lean more towards @toddbaert's approach, it feels cleaner to have an `EvaluationContext` type with a `SetAttribute` / `SetAttributes` method(s)

beeme1mr:
If this is the approach, it must be a priority. To stay on schedule, this change would need to be made available tomorrow.

https://github.com/open-feature/spec/issues/146

toddbaert:
> i personally lean more towards @toddbaert's approach, it feels cleaner to have an `EvaluationContext` type with a `SetAttribute` / `SetAttributes` method(s)

I kinda hate my approach but I worry that it might be the best one.

skyerus:
Would allowing the the map to be modified after construction violate [this spec requirement](https://github.com/open-feature/spec/blob/main/specification/sections/04-hooks.md#requirement-414)?
> Requirement 4.1.4
The evaluation context MUST be mutable only within the before hook.

toddbaert:
> Would allowing the the map to be modified after construction violate [this spec requirement](https://github.com/open-feature/spec/blob/main/specification/sections/04-hooks.md#requirement-414)?
> 
> > Requirement 4.1.4
> > The evaluation context MUST be mutable only within the before hook.

That point is probably a bit misleading. We should improve the wording.

[...]

PR #93, opened 2022-10-06
fix: locks on singleton and client state to ensure thread safety

skyerus:
## This PR

- locks on singleton and client state to ensure thread safety

This PR adds multi-read/single-write locks to both the global singleton and the client. Particularly it addresses problems relating to the providers/hooks being mutated while an evaluation is occurring. This is especially important since some providers have provider hooks, and if we don't lock, it's possible hooks from provider1 run on provider2 if an interleaving thread switched the provider.

I've opted for the approach of locking a provider and its hooks to an evaluation transaction. Another thread is free to set a new provider but any ongoing transactions will carry on using the existing provider and its hooks.

### Related Issues

Fixes https://github.com/open-feature/go-sdk/issues/90


pkg/openfeature/client.go
```diff
@@ -531,8 +573,8 @@ func (c Client) evaluate(
 		},
 	}
 
-	apiClientInvocationProviderHooks := append(append(append(api.hooks, c.hooks...), options.hooks...), api.provider.Hooks()...) // API, Client, Invocation, Provider
-	providerInvocationClientApiHooks := append(append(append(api.provider.Hooks(), options.hooks...), c.hooks...), api.hooks...) // Provider, Invocation, Client, API
+	apiClientInvocationProviderHooks := append(append(append(api.hooks(), c.hooks...), options.hooks...), api.provider().Hooks()...) // API, Client, Invocation, Provider
+	providerInvocationClientApiHooks := append(append(append(api.provider().Hooks(), options.hooks...), c.hooks...), api.hooks()...) // Provider, Invocation, Client, API
```
toddbaert, line 577: I think I see one nasty problem here. If between this line and the evaluation, I switch providers, the hooks from the first provider will run with the evaluation of the second.

I think the only way to resolve it would be to expose a lock method on the API which could be read-locked locking at the point the hooks are retrieved, and unlocking after evaluation.

I don't know if adding that sort of locking is worth it, when you consider that providers are switched infrequently.

toddbaert: Perhaps we could defer getting provider hooks until the lock is active? I know that would make a mess.

skyerus: Great observation, I've opted for the approach of locking a provider and its hooks to an evaluation transaction. Another thread is free to set a new provider but any ongoing transactions will carry on using the existing provider and its hooks.
If there are any concerns we can revise this solution.

james-milligan (approved):

beeme1mr (approved):

pkg/openfeature/client.go
```diff
@@ -514,22 +547,30 @@ func (c Client) ObjectValueDetails(ctx context.Context, flag string, defaultValu
...
 		"evaluationContext", evalCtx, "evaluationOptions", options,
 	)
-	evalCtx = mergeContexts(evalCtx, c.evaluationContext, api.evaluationContext) // API (global) -> client -> invocation
+	evalCtx = mergeContexts(evalCtx, c.evaluationContext, api.evaluationContext()) // API (global) -> client -> invocation
+
+	// ensure that the same provider & hooks are used across this transaction to avoid unexpected behaviour
+	api.RLock()
+	provider := api.prvder
+	globalHooks := api.hks
+	api.RUnlock()
```
toddbaert, line 563: If I comment out api.RLock() and api.RUnlock here, no tests fail.

Ideally, tests would help us ensure this lock isnt removed, but I'm finding this EXTREMELY hard to test in Java.

pkg/openfeature/client_test.go
```diff
@@ -806,3 +806,63 @@ func TestErrorCodeFromProviderReturnedInEvaluationDetails(t *testing.T) {
...
+	mockProvider1Hook.EXPECT().Finally(gomock.Any(), gomock.Any())
+
+	client := NewClient("test")
+	_, err := client.BooleanValue(context.Background(), "foo", true, EvaluationContext{})
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestClientProviderLock(t *testing.T) {
```
skyerus, line 841: This is the best I could do @toddbaert
The test ensures that the RLock & RUnlock is called at that point (if the RLock/RUnlock are removed or moved the test fails).

pkg/openfeature/client.go
skyerus: Done now (whitebox though :( )

PR #96, opened 2022-10-07
fix: resolution error only includes the code

### LINKED ISSUE #94: [FEATURE] Improve default error logging
### Requirements

By default the leveled logger is only outputting errors, via [log.Println(err)](https://github.com/open-feature/go-sdk/blob/9fbed881195a5e1801a0157309db855bbad2b356/pkg/openfeature/logger.go#L24), which in turn calls the [ProviderResolutionDetail.Error()](https://github.com/open-feature/go-sdk/blob/9fbed881195a5e1801a0157309db855bbad2b356/pkg/openfeature/provider.go#L68).

This combination means you end up with logs that look like `2022/10/07 09:09:08 GENERAL` because only the error code is returned by `Error()`.

My preference would be to not log anything like the other levels do; `Error(...)` would be noop, leaving this to userland to decide what they want logging using `SetLogger`. If not, perhaps the message should include more details & make it clear(er) where it originates from, as when interleaved with other logs it makes hunting the cause down tricky.

Wanted to gather any thoughts & opinions first, but I'd be happy to attempt a PR depending on the outcome of this issue. Felt like a sharp edge and we could improve the out-of-the-box experience.

rcrowe:
## This PR
Adds the error message to the formatted error string. 
Makes it clear when logged that the error originates from the OpenFeature SDK.

### Related Issues

Fixes #94

### DIFF
--- a/pkg/openfeature/logger.go
+++ b/pkg/openfeature/logger.go
@@ -21,7 +21,9 @@ func (l logger) Enabled(level int) bool { return true }
 
 func (l logger) Info(level int, msg string, keysAndValues ...interface{}) {}
 
-func (l logger) Error(err error, msg string, keysAndValues ...interface{}) { log.Println(err) }
+func (l logger) Error(err error, msg string, keysAndValues ...interface{}) {
+	log.Println("openfeature:", err)
+}
 
 func (l logger) WithValues(keysAndValues ...interface{}) logr.LogSink { return l }
 
--- a/pkg/openfeature/provider.go
+++ b/pkg/openfeature/provider.go
@@ -47,8 +47,8 @@ type FeatureProvider interface {
 // N.B we could use generics but to support older versions of go for now we will have type specific resolution
 // detail
 type ProviderResolutionDetail struct {
-	Reason          Reason
 	ResolutionError ResolutionError
+	Reason          Reason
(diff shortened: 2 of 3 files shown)


pkg/openfeature/logger.go
```diff
@@ -21,7 +21,9 @@ func (l logger) Enabled(level int) bool { return true }
 
 func (l logger) Info(level int, msg string, keysAndValues ...interface{}) {}
 
-func (l logger) Error(err error, msg string, keysAndValues ...interface{}) { log.Println(err) }
+func (l logger) Error(err error, msg string, keysAndValues ...interface{}) {
+	log.Println("openfeature:", err)
```
skyerus, line 25: It looks like consensus has been that we'll default to no op even in error cases so I think this line can be removed entirely.

skyerus (commented):
Nice work, thanks!

pkg/openfeature/logger.go
beeme1mr: I think this should be done in a separate PR. That way, we can weigh the pros/cons of not logging anything out of the box.

beeme1mr (approved):

skyerus (approved):

Issue #88, opened 2022-10-03
Implement gherkin integration test suite


toddbaert:
- Implement gherkin suite [here](https://github.com/open-feature/test-harness/blob/main/features/evaluation.feature)
- run in CI
- See examples PRs in [Java](https://github.com/open-feature/java-sdk/pull/77) and [JS](https://github.com/open-feature/js-sdk/pull/225) SDKs

PR #99, opened 2022-10-13
chore: integration tests using gherkin test suite

skyerus:
## This PR

- integration with [the gherkin test suite](https://github.com/open-feature/test-harness)
- runs gherkin suite as part of the PR checks

### Related Issues

Fixes https://github.com/open-feature/go-sdk/issues/88


beeme1mr (approved):

Makefile
```diff
@@ -4,6 +4,8 @@ mockgen:
 	mockgen -source=pkg/openfeature/hooks.go -destination=pkg/openfeature/hooks_mock_test.go -package=openfeature
 	mockgen -source=pkg/openfeature/mutex.go -destination=pkg/openfeature/mutex_mock_test.go -package=openfeature
 test:
+	go test --short -cover ./...
+integration-test: # dependent on `docker run -p 8013:8013 ghcr.io/open-feature/flagd-testbed:latest`
```
toddbaert, line 8: Not necessary, but in the Java and JS repos, I added the git submodule retrieval into the integration test goal/script. This has the benefit of "just working" when somebody runs it locally.

integration/evaluation_test.go
```diff
@@ -0,0 +1,658 @@
...
+) (context.Context, error) {
+	got, err := client.StringValue(ctx, flagKey, defaultValue, openfeature.EvaluationContext{})
+	if err != nil {
+		return ctx, fmt.Errorf("openfeature client: %w", err)
+	}
+
+	return context.WithValue(ctx, ctxStorageKey{}, got), nil
+}
+
+func theResolvedStringValueShouldBe(ctx context.Context, expectedValue string) error {
```
toddbaert, line 79: There's a few functions here like this one and `theResolvedIntegerValueShouldBe` that might be able to be made generic so they can work for multiple gherkin tests. Probably not possible with the boolean version wherein you'll need to parse.

This is entirely not-necessary though.

toddbaert:
Added minor comments. Works (and fails) as expected for me locally. Nice job, this adds a lot of confidence.

toddbaert (approved):
Added minor comments. Works (and fails) as expected for me locally. Nice job, this adds a lot of confidence.

integration/evaluation_test.go
skyerus: Do you mean locating any functions where behaviour is being duplicated and merging them into a singular function?
I can't see any instances where this is happening. `theResolvedStringValueShouldBe` and `theResolvedStringResponseShouldBe` are similar in what they're asserting but the execution is different due to the way the data is store/retrieved.

Maybe I've misunderstood.

Makefile
skyerus: Nice suggestion, done

integration/evaluation_test.go
toddbaert: I mean functions like `theResolvedStringValueShouldBe` and `theResolvedIntegerValueShouldBe`. These functions are the same, except the interface casting and string format token. which are associated with the type of the `expectedValue`. That's why I think (possibly incorrectly) they could be made generic.

This is absolutely not a necessary change though, even if I'm correct.

skyerus: I think due to the way we're storing the typed data and retrieving it for the assertion the complexity of doing this wouldn't be worth the benefits

PR #118, opened 2022-12-09
fix: authorize nil value for object evaluation

### LINKED ISSUE #117: [BUG] When calling client.ObjectValueDetails with a value as nil the ResolutionDetail are not accessible.
### Observed behavior

I am using `nil` as a default value while calling `client.ObjectValueDetails` but when doing this everytime if the default value is used we receive no `ResolutionDetail` from the SDK.

After looking a bit at the SDK it seems that we add the `ResolutionDetail` only if the `value` is not nil.
https://github.com/open-feature/go-sdk/blob/fd678bd263eb532942c4777f4eed8e26a6e52694/pkg/openfeature/client.go#L642

It means that we cannot use `nil` as `value` in the SDK.

### Expected Behavior

I would like to be able to use `nil` as value for my flag.

### Steps to reproduce

SDK return a flag with a `nil` value.

thomaspoignant:
## This PR
Allow using a `nil` value when calling the `ObjectValueDetails` method.

### Related Issues
Fixes #117

### DIFF
--- a/pkg/openfeature/client_test.go
+++ b/pkg/openfeature/client_test.go
@@ -873,6 +873,7 @@ func TestObjectEvaluationShouldSupportNilValue(t *testing.T) {
 
 	variant := "variant1"
 	reason := TargetingMatchReason
+	var value interface{} = nil
 
 	mockProvider := NewMockFeatureProvider(ctrl)
 	mockProvider.EXPECT().Metadata().AnyTimes()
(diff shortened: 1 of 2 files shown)


pkg/openfeature/client_test.go
```diff
@@ -866,3 +866,43 @@ func TestClientProviderLock(t *testing.T) {
...
+		Return(InterfaceResolutionDetail{
+			Value: value,
+			ProviderResolutionDetail: ProviderResolutionDetail{
+				Variant: variant,
+				Reason:  reason,
+			},
+		})
+
+	client := NewClient("test")
+	evDetails, _ := client.ObjectValueDetails(context.Background(), "foo", nil, EvaluationContext{})
```
skyerus, line 892: Can we have a check on the error?

thomaspoignant: Done

toddbaert (approved):
This makes sense to me.

I think once the errors are fixed it can be merged as long as all the tests pass.

thomaspoignant:
I tried to fix everything from my phone, but it was not a good idea 🤦.
It should be fine now.

james-milligan (approved):

toddbaert:
> I tried to fix everything from my phone, but it was not a good idea facepalm. It should be fine now.

Honestly I'm impressed you even attempted it!

toddbaert (approved):

PR #140, opened 2023-01-17
feat!: upgrade Go to 1.18

skyerus:
## This PR

Upgrades Go to 1.18

### Related Issues

Fixes #12

PR #141, opened 2023-01-19
fix: validate that a flag key is valid UTF-8 & implemented fuzzing tests

### LINKED ISSUE #127: [TESTS] Introduce fuzzing tests
### Requirements

Fuzzing is a [repository requirement](https://github.com/open-feature/.github/blob/main/CONTRIBUTING.md#repository-requirements).

Go supports fuzzing natively since 1.18, `go-sdk` is still using version 1.17. Either upgrade to 1.18 or use a third-party library.

skyerus:
## This PR

- Fuzzing tests.
- Validate that a flag key is valid UTF-8 (issue arose from fuzzing 🎉 )

### Related Issues

Fixes #127

### DIFF
--- a/pkg/openfeature/client.go
+++ b/pkg/openfeature/client.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/go-logr/logr"
 )
@@ -563,6 +564,18 @@ func (c *Client) evaluate(
 		"evaluationContext", evalCtx, "evaluationOptions", options,
 	)
 
+	evalDetails := InterfaceEvaluationDetails{
+		Value: defaultValue,
+		EvaluationDetails: EvaluationDetails{
+			FlagKey:  flag,
+			FlagType: flagType,
+		},
+	}
+
+	if !utf8.Valid([]byte(flag)) {
+		return evalDetails, NewParseErrorResolutionError("flag key is not a UTF-8 encoded string")
+	}
+
 	// ensure that the same provider & hooks are used across this transaction to avoid unexpected behaviour
 	api.RLock()
 	provider := api.prvder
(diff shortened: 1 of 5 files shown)


skyerus:
Will rebase once 1.18 is merged: https://github.com/open-feature/go-sdk/pull/140

beeme1mr (approved):

pkg/openfeature/client.go
```diff
@@ -563,6 +564,18 @@ func (c *Client) evaluate(
...
+		Value: defaultValue,
+		EvaluationDetails: EvaluationDetails{
+			FlagKey:  flag,
+			FlagType: flagType,
+		},
+	}
+
+	if !utf8.Valid([]byte(flag)) {
+		return evalDetails, NewParseErrorResolutionError("flag key is not a UTF-8 encoded string")
+	}
```
toddbaert, line 577: Besides flag key, is there any other places we should validate this?

Is it worth doing it for context keys and values, for instance (that could incur some performance penalties... Just a thought!

skyerus: Perhaps there is value to that, should we create a separate issue to raise the question?

Kavindu-Dodan:
Overall looks good. One suggestion, let's add few more fuzz seeds. For example, flag keys with mix cases, include numbers (A-z0-9) plus special characters if needed. Plus the same for default variant to match the type.

pkg/openfeature/client.go
skyerus: https://github.com/open-feature/go-sdk/issues/147

skyerus:
> Overall looks good. One suggestion, let's add few more fuzz seeds. For example, flag keys with mix cases, include numbers (A-z0-9) plus special characters if needed. Plus the same for default variant to match the type.

Thanks! Good suggestion, done :)

Kavindu-Dodan (approved):

PR #151, opened 2023-02-01
feat: introduced v2 module

skyerus:
## This PR

- Creates the new module github.com/open-feature/go-sdk/v2 for v2.0.0 release

### Notes

Anywhere that uses a v1 interface (e.g. providers) will be forced to make changes to use the v2 interface. For example, the following won't work if an app author is using v2 of the go-sdk with a non-updated flagd provider.
```
provider := flagd.NewProvider()
openfeature.SetProvider(provider)
```


toddbaert:
There's some concerns about whether this is the right choice. The main reason we are doing this is because of the go version update, which though technically is breaking, isn't an API change.

toddbaert:
Closing this. We've opted to not consider the 1.18 upgrade a breaking change.

Issue #177, opened 2023-05-16
Changes to support spec `v0.6.0` compliance


toddbaert:
## Implement specification v0.6.0 changes

See [changes since last release](https://github.com/open-feature/spec/compare/v0.5.2...v0.6.0).

**_Please note there's no need to complete all these tasks in a single PR (in fact, that's not recommended)._**

### Functional requirements:

- [x] implement flag metadata: https://github.com/open-feature/spec/commit/74c373e089ad77bf8cac84f3d93c00c945ff3a8a
- [x] #183
- [x] #184
- [x] #186

### Non-functional requirements:

- [x] add relevant tests
- [x] add examples and documentation, see template README
- [x] update specification badge to `v0.6.0`: [![Specification](https://img.shields.io/static/v1?label=Specification&message=v0.6.0&color=yellow)](https://github.com/open-feature/spec/tree/v0.6.0)

### Keep in mind:

- backwards compatibility is _highly encouraged_
  - take every precaution to ensure existing code written by _application authors_ and existing providers implementations, do not break
- it's recommended these features are released together

Kavindu-Dodan:
I will continue to work on **implement initialization/shutdown: https://github.com/open-feature/spec/commit/a4ffec3d4416b8ad6d12d1ebfa2faf81ad153f65**

toddbaert:
@Kavindu-Dodan I've created sub-issues and assigned to you. Thanks!

Kavindu-Dodan:
All functional and non-functional goals have been met with the linked PRs.

PR #178, opened 2023-05-23
feat: add flag metadata field

james-milligan:
## This PR

- implement flag metadata: https://github.com/open-feature/spec/commit/74c373e089ad77bf8cac84f3d93c00c945ff3a8a

### Related Issues

https://github.com/open-feature/go-sdk/issues/177

### Notes

If the flag metadata is not in the provider response, then the returned evaluation details will have a nil value, this can be updated to have an empty map, but IMO its a cleaner implementation to leave this value as nil ([nil slices have length 0](https://go.dev/play/p/_m0EwwCvJ65))

### DIFF
--- a/pkg/openfeature/client.go
+++ b/pkg/openfeature/client.go
@@ -218,7 +218,7 @@ func (f FlagMetadata) GetInt(key string) (int64, error) {
 	case int64:
 		return v.(int64), nil
 	default:
-		return 0, fmt.Errorf("wrong type for key %s, expected integer, got %T", key, t)return 0, fmt.Errorf("wrong type for key %s, expected string, got %T", key, t)
+		return 0, fmt.Errorf("wrong type for key %s, expected integer, got %T", key, t)
 	}
 }
 
--- a/pkg/openfeature/client.go
+++ b/pkg/openfeature/client.go
@@ -218,7 +218,7 @@ func (f FlagMetadata) GetInt(key string) (int64, error) {
 	case int64:
 		return v.(int64), nil
 	default:
-		return 0, fmt.Errorf("wrong type for key %s, expected string, got %T", key, t)
+		return 0, fmt.Errorf("wrong type for key %s, expected integer, got %T", key, t)return 0, fmt.Errorf("wrong type for key %s, expected string, got %T", key, t)
 	}
 }
 
@@ -234,7 +234,7 @@ func (f FlagMetadata) GetFloat(key string) (float64, error) {
 	case float64:
 		return v.(float64), nil
 	default:
-		return 0, fmt.Errorf("wrong type for key %s, expected string, got %T", key, t)
+		return 0, fmt.Errorf("wrong type for key %s, expected float, got %T", key, t)
 	}
 }
 
--- a/pkg/openfeature/client.go
+++ b/pkg/openfeature/client.go
@@ -164,11 +164,69 @@ type ResolutionDetail struct {
 	Reason       Reason
 	ErrorCode    ErrorCode
 	ErrorMessage string
-	FlagMetadata *FlagMetadata
+	FlagMetadata FlagMetadata
 }
 
-// FlagMetadata provides a mechanism for providers to surface additional information about a feature flag
+// A structure which supports definition of arbitrary properties, with keys of type string, and values of type boolean, string, int64 or float64.
+//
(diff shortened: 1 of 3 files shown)


pkg/openfeature/client.go
```diff
@@ -164,6 +164,68 @@ type ResolutionDetail struct {
...
+}
+
+// Fetch int64 value from FlagMetadata, returns an error if the key does not exist, or, the value is of the wrong type
+func (f FlagMetadata) GetInt(key string) (int64, error) {
+	v, ok := f[key]
+	if !ok {
+		return 0, fmt.Errorf("key %s does not exist in FlagMetadata", key)
+	}
+	switch t := v.(type) {
+	case int64:
```
thisthat, line 210: what about the remaining integer primitives, namely int, int8, int16, int32? I would expect these to be coerced to int64

toddbaert:
This all makes sense to me.

I'm particularly interested in other's opinions on this:

> If the flag metadata is not in the provider response, then the returned evaluation details will have a nil value, this can be updated to have an empty map, but IMO its a cleaner implementation to leave this value as nil ([nil slices have length 0](https://go.dev/play/p/_m0EwwCvJ65))

I'll defer to those with more familiarity with go semantics.

Kavindu-Dodan:
> This all makes sense to me.
> 
> I'm particularly interested in other's opinions on this:
> 
> > If the flag metadata is not in the provider response, then the returned evaluation details will have a nil value, this can be updated to have an empty map, but IMO its a cleaner implementation to leave this value as nil ([nil slices have length 0](https://go.dev/play/p/_m0EwwCvJ65))
> 
> I'll defer to those with more familiarity with go semantics.

From pure spec point of view [1]

> If the flag metadata field in the flag resolution structure returned by the configured provider is set, the evaluation details structure's flag metadata field MUST contain that value. Otherwise, it MUST contain an empty record.

IMO, to be spec compliant, we should have an empty record here 🤔  

[1] - https://openfeature.dev/specification/sections/flag-evaluation/#requirement-1413

beeme1mr:
How do we want to handle releasing a breaking change? If we're going to release this as a minor version, we need to make sure Release Please is configured properly and that the changes are clearly documented.

pkg/openfeature/client.go
```diff
@@ -164,6 +164,78 @@ type ResolutionDetail struct {
...
+	case int8:
+		return int64(v.(int8)), nil
+	case int16:
+		return int64(v.(int16)), nil
+	case int32:
+		return int64(v.(int32)), nil
+	case int64:
+		return v.(int64), nil
+	default:
+		return 0, fmt.Errorf("wrong type for key %s, expected string, got %T", key, t)
```
Kavindu-Dodan, line 221:
```suggestion
		return 0, fmt.Errorf("wrong type for key %s, expected integer, got %T", key, t)return 0, fmt.Errorf("wrong type for key %s, expected string, got %T", key, t)
```

pkg/openfeature/client.go
```diff
@@ -164,6 +164,78 @@ type ResolutionDetail struct {
...
+	if !ok {
+		return 0, fmt.Errorf("key %s does not exist in FlagMetadata", key)
+	}
+	switch t := v.(type) {
+	case float32:
+		return float64(v.(float32)), nil
+	case float64:
+		return v.(float64), nil
+	default:
+		return 0, fmt.Errorf("wrong type for key %s, expected string, got %T", key, t)
```
Kavindu-Dodan, line 237:
```suggestion
		return 0, fmt.Errorf("wrong type for key %s, expected float, got %T", key, t)
```

toddbaert:
> How do we want to handle releasing a breaking change? If we're going to release this as a minor version, we need to make sure Release Please is configured properly and that the changes are clearly documented.

I didn't notice the `!` in the title. Is this actually breaking? I assumed from

> If the flag metadata is not in the provider response

That this was non-breaking? @james-milligan - can you clarify?

james-milligan:
yes this is non breaking, apologies should have caught this sooner!

toddbaert:
I've updated the title so it reads a bit better in release notes. Approved!

Kavindu-Dodan (approved):

beeme1mr (approved):

Issue #184, opened 2023-06-13
implement initialization/shutdown: open-feature/spec@a4ffec3


toddbaert:
See parent: https://github.com/open-feature/go-sdk/issues/177

PR #185, opened 2023-06-13
feat: Initialize and shutdown behavior 

Kavindu-Dodan:
## This PR

Fixes https://github.com/open-feature/go-sdk/issues/184 by introducing Initialization & Shutdown support.

### Initialization and Shutdown

This is a non-breaking change 🚀 

Initialization and Shutdown along with Status accessor is introduced through the new interface `StateHandler`.

OpenFeature API implementation undertakes the init and shutdown handling based on interface implementation (type assertion)

### Implementation considerations 

Provider initialization is done asynchronously to avoid blocking the caller. Provider implementations should update their status to `Ready` as soon as initialization is done. 

This implementation does not consider Go `context` to facilitate shutdown. This is to reduce complexity and take over full control of the provider's initialization & shutdown methods.

### DIFF
--- a/pkg/openfeature/api.go
+++ b/pkg/openfeature/api.go
@@ -17,6 +17,18 @@ type evaluationAPI struct {
 	mu              sync.RWMutex
 }
 
+// newEvaluationAPI is a helper to generate an API. Used internally
+func newEvaluationAPI() evaluationAPI {
+	return evaluationAPI{
+		defaultProvider: NoopProvider{},
+		namedProviders:  map[string]FeatureProvider{},
+		hks:             []Hook{},
+		evalCtx:         EvaluationContext{},
+		logger:          logr.New(internal.Logger{}),
+		mu:              sync.RWMutex{},
+	}
+}
+
 // setProvider sets the default provider of the evaluationAPI. Returns an error if FeatureProvider is nil
 func (api *evaluationAPI) setProvider(provider FeatureProvider) error {
 	api.mu.RLock()
--- a/pkg/openfeature/api.go
+++ b/pkg/openfeature/api.go
@@ -40,16 +40,20 @@ func (api *evaluationAPI) setProvider(provider FeatureProvider) error {
 	}
 
 	// Initialize new default provider and shutdown the old one
+	// Provider update must be non-blocking, hence initialization & shutdown happens concurrently
+	oldProvider := api.defaultProvider
+	api.defaultProvider = provider
+
 	go func() {
-		api.mu.Lock()
-		defer api.mu.Unlock()
+		v, ok := provider.(StateHandler)
+		if ok {
+			v.Init(api.evalCtx)
+		}
 
-		oldProvider := api.defaultProvider
-
-		api.defaultProvider = provider
-		api.defaultProvider.Init(api.evalCtx)
-
-		oldProvider.Shutdown()
+		v, ok = oldProvider.(StateHandler)
+		if ok {
+			v.Shutdown()
+		}
 	}()
 
 	return nil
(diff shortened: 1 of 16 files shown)


pkg/openfeature/provider.go
```diff
@@ -43,6 +47,59 @@ type FeatureProvider interface {
...
+}
+
+// StateHandler is the contract of initialization & shutdown
+type StateHandler interface {
+	Init(evaluationContext EvaluationContext)
+	Shutdown()
+	Status() State
+}
+
+// NoopStateHandler is an always Ready, otherwise noop implementation of StateHandler
```
toddbaert, line 60: This comment seems possibly incomplete.

pkg/openfeature/provider.go
```diff
@@ -34,7 +38,7 @@ type FlattenedContext map[string]interface{}
 type Reason string
 
 // FeatureProvider interface defines a set of functions that can be called in order to evaluate a flag.
-// vendors should implement
+// This should be implemented by vendors.
```
toddbaert, line 41:
```suggestion
// This should be implemented by flag management systems.
```

This term is more neutral. flagd isn't a vendor, for instance.

pkg/openfeature/provider.go
Kavindu-Dodan: Updated with the latest changeset

pkg/openfeature/provider.go
Kavindu-Dodan: Updated with the latest changeset

pkg/openfeature/provider.go
```diff
@@ -43,6 +47,59 @@ type FeatureProvider interface {
 	IntEvaluation(ctx context.Context, flag string, defaultValue int64, evalCtx FlattenedContext) IntResolutionDetail
 	ObjectEvaluation(ctx context.Context, flag string, defaultValue interface{}, evalCtx FlattenedContext) InterfaceResolutionDetail
 	Hooks() []Hook
+	StateHandler
```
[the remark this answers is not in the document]
Kavindu-Dodan: Updated the implementation to be non-breaking. Only catch is that we can't enforce a `status` accessor from the provider

pkg/openfeature/util_test.go
```diff
@@ -0,0 +1,21 @@
+package openfeature
+
+// Test Utils
+
+type StateHandlerForTests struct {
```
thisthat, line 5: I would not expose this since it is testing code.

pkg/openfeature/util_test.go
```diff
@@ -0,0 +1,21 @@
+package openfeature
```
thisthat, line 1: maybe worth to move into a testing package?

pkg/openfeature/api.go
```diff
@@ -0,0 +1,170 @@
...
+	defer api.mu.Unlock()
+
+	if provider == nil {
+		return errors.New("default provider cannot be set to nil")
+	}
+
+	// Initialize new default provider and shutdown the old one
+	// Provider update must be non-blocking, hence initialization & shutdown happens concurrently
+	oldProvider := api.defaultProvider
+	go func() {
```
thisthat, line 45: maybe worth having a dedicate `func switchProvider(old, new FeatureProvider)` function that is shared with setNamedProvider

thisthat (approved):
I like moving this into a new interface so we don't break the existing contract. Providers can then add these 3 new functions to their struct receivers in a separate version keeping compatibility with any version of OF SDK.

thisthat (commented):
I like moving this into a new interface so we don't break the existing contract. Providers can then add these 3 new functions to their struct receivers in a separate version keeping compatibility with any version of OF SDK.

thisthat (commented):
I like moving this into a new interface so we don't break the existing contract. Providers can then add these 3 new functions to their struct receivers in a separate version keeping compatibility with any version of OF SDK.

pkg/openfeature/api.go
Kavindu-Dodan: Good suggestion. I will add this with the next commit

pkg/openfeature/util_test.go
Kavindu-Dodan: At least not for now. but this can be considered if we see this grow. Right now, this is small enough to live in this level

pkg/openfeature/util_test.go
Kavindu-Dodan: Fixed with the latest commit

pkg/openfeature/api.go
Kavindu-Dodan: Logic is now isolated and shared with callers

Kavindu-Dodan:
> I like moving this into a new interface so we don't break the existing contract. Providers can then add these 3 new functions to their struct receivers in a separate version keeping compatibility with any version of OF SDK.

A big thanks go to @toddbaert for proposing to go with a non-breaking contract. And yes, the implementation is flexible in this regards :)

toddbaert (approved):

beeme1mr (approved):

Issue #186, opened 2023-06-14
Implement event support 


Kavindu-Dodan:
### Requirements

see parent: https://github.com/open-feature/go-sdk/issues/177

Spec change - https://github.com/open-feature/spec/commit/bf9e45b1d51c316ee2d6340bc2b807294995d0ad

- [ ] Initial implementation 
- [ ] 1 to n support between provider and clients & Emiting of initialize success/error
```[tasklist]
### Tasks
```

Kavindu-Dodan:
Completed and released with v1.5.x release

PR #188, opened 2023-06-20
feat: eventing implementation

Kavindu-Dodan:
## This PR

Covers the initial implementation of https://github.com/open-feature/go-sdk/issues/186

⚠️  Note that, this implementation does not cover the 1 to N relationship requirement of provider to named client requirement (i.e - single instance of a provider can be bound to multiple names, thus allowing that single provider to be bound by multiple clients). A follow-up PR is ready  for this - https://github.com/open-feature/go-sdk/pull/190 

This PR introduces OpenFeature eventing spec [1] to Go SDK. The implementation is non-breaking and providers which plan to support eventing can opt-in by implementing `EventHandler` interface.

### Implementation 

`EventHandler` interface contains a single method `EventChannel() <- chan Event`. OpenFeature core implementation performs a type assertion on the provider for `EventHandler` and listen to `Event` channel. It is provider implementation responsibility to emit spec-compliant events (ex:- ProviderReady when initialization is complete OR ProviderError when initialization failed)

The core event handling logic is concentrated on `event_executor.go`. OpenFeature core implementation maintains a single `EventExecutor` instance that handles both global (api) level events and client level (scoped) events.

Client and API both add and remove event handlers (callbacks) through their spec client accessors and these get stored at the global `EventExecutor`

```mermaid
flowchart TD
    A[OpenFeature] -->|Register Global Handler|D
    D[API Singleton] -->  B(EventExecutor)
    C[Client Instance] -->|Register Client Handler|D

```

When handling events from providers, all global-level handlers are always invoked. But, the client (scoped) handlers only run if they are associated with the specific provider. This is done by comparing event payload's provider name & registered client name of the handler

```mermaid
sequenceDiagram
    Provider->>+EventExecutor: Emit Event: ProviderReady
    EventExecutor-->>EventExecutor: Run Global Handlers
    EventExecutor-->>EventExecutor: Run Client Handlers - filter with provider name
```

### Testing 

Consider checking `event_executor_test.go` [2] to validate spec compliance as well as the correctness of the implementation 

[1] - https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md#5-events
[2] - https://github.com/open-feature/go-sdk/blob/fe93b46b4afc11533b0530314cc71a823abfe705/pkg/openfeature/event_executor_test.go


pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+
+	group, gCtx := errgroup.WithContext(ctx)
+
+	group.Go(func() error {
+		// first run API handlers
+		for _, c := range e.apiRegistry[event.EventType] {
+			f := *c
+			func() {
```
bacherfl, line 168: this function and the one handling the client handlers are mostly the same, except for the client name that's being added in the latter. To improve readability of this function I suggest to make one reusable function (smth. like `e.callEventHandlerFunc`) and make use of that one for the API and Client event handlers

pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	} else {
+		e.apiRegistry[t] = append(e.apiRegistry[t], c)
+	}
+}
+
+func (e *eventHandler) removeApiHandler(t EventType, c EventCallBack) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	entrySlice, ok := e.apiRegistry[t]
```
bacherfl, line 69: can you also add a test for removing a handler that was not registered before?

pkg/openfeature/event_handler.go
Kavindu-Dodan: Thanks, yes will refactor to extract this out.

pkg/openfeature/event_handler.go
Kavindu-Dodan: good suggestion, done :)

pkg/openfeature/api.go
```diff
@@ -15,17 +14,21 @@ type evaluationAPI struct {
 	hks             []Hook
 	evalCtx         EvaluationContext
 	logger          logr.Logger
-	mu              sync.RWMutex
+	eventHandler
+	mu sync.RWMutex
```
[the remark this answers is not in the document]
Kavindu-Dodan: Here my intension was to embed `eventHandler` struct into the api. This is why it's not a member of the struct. Let me know if you have different opinion on this :)

pkg/openfeature/api.go
```diff
@@ -41,9 +44,16 @@ func (api *evaluationAPI) setProvider(provider FeatureProvider) error {
...
-	api.initAndShutdown(provider, api.defaultProvider)
+	oldProvider := api.defaultProvider
+	api.initAndShutdown(provider, oldProvider)
 	api.defaultProvider = provider
 
+	api.registerEventingProvider(provider)
+	err := api.unregisterEventingProvider(oldProvider)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Kavindu-Dodan: This is only done at two places, so I think I would avoid for thistime. And this could be a refactoring effort and can be a follow up.

pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	"sync"
+	"time"
+)
+
+// event handler is registry to connect API and Client event handlers to Providers
+
+// handlerExecutionTime defines the maximum time event handler will wait for its handlers to complete
+const handlerExecutionTime = 500 * time.Millisecond
+
+type eventHandler struct {
```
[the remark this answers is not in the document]
Kavindu-Dodan: Good point, how about `eventExecutor` . I will update this and please feel free to give your opinion

pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	return eventHandler{
+		providerShutdownHook: map[string]chan interface{}{},
+		apiRegistry:          map[EventType][]EventCallBack{},
+		clientRegistry:       map[string]clientHolder{},
+		logger:               logger,
+	}
+}
+
+// clientHolder is a helper struct to hold client specific callbacks
+type clientHolder struct {
```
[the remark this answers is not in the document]
Kavindu-Dodan: As client is a OpenFeature specific terminology, renaming this to `scopedCallbacks`

pkg/openfeature/event_handler.go
Kavindu-Dodan: This is done - https://github.com/open-feature/go-sdk/pull/188/files#diff-5088e3f82f28056163763bfb774c1b8e91bb585580b4b1add90b133445b8bae6R191-R206

pkg/openfeature/event_handler.go
Kavindu-Dodan: Tests added to see if this panic

pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func newClientHolder(client string) clientHolder {
+	return clientHolder{
+		name:   client,
+		holder: map[EventType][]EventCallBack{},
+	}
+}
+
+func (e *eventHandler) updateLogger(l logr.Logger) {
```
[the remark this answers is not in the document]
Kavindu-Dodan: This is connected to the struct embedding 🤔  so I would have to keep this as it is

pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
+package openfeature
+
+import (
+	"context"
+	"fmt"
+	"github.com/go-logr/logr"
+	"golang.org/x/sync/errgroup"
```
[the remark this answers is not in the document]
Kavindu-Dodan: done

pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
...
+func (e *eventHandler) registerClientHandler(clientName string, t EventType, c EventCallBack) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	_, ok := e.clientRegistry[clientName]
+	if !ok {
+		e.clientRegistry[clientName] = newClientHolder(clientName)
+	}
+
+	v := e.clientRegistry[clientName]
```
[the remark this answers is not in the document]
Kavindu-Dodan: renamed to `callback` ;)

pkg/openfeature/event_handler.go
```diff
@@ -0,0 +1,243 @@
...
+		apiRegistry:          map[EventType][]EventCallBack{},
+		clientRegistry:       map[string]clientHolder{},
+		logger:               logger,
+	}
+}
+
+// clientHolder is a helper struct to hold client specific callbacks
+type clientHolder struct {
+	name   string
+	holder map[EventType][]EventCallBack
```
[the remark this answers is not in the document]
Kavindu-Dodan: done

bacherfl (approved):

pkg/openfeature/event_executor_test.go
```diff
@@ -0,0 +1,529 @@
...
+// including PROVIDER_READY, PROVIDER_ERROR, PROVIDER_CONFIGURATION_CHANGED and PROVIDER_STALE,
+// with a provider event details payload.
+func TestEventHandler_RegisterUnregisterEventProvider(t *testing.T) {
+
+	t.Run("Ignored non-eventing providers", func(t *testing.T) {
+		handler := newEventHandler(logger)
+		handler.registerEventingProvider(NoopProvider{})
+
+		if len(handler.providerShutdownHook) != 0 {
+			t.Errorf("implementation should ignore non eventing provider")
```
toddbaert, line 30: :sunglasses:

pkg/openfeature/event_executor_test.go
```diff
@@ -0,0 +1,529 @@
...
+func init() {
+	logger = logr.New(internal.Logger{})
+}
+
+// Requirement 5.1.1 The provider MAY define a mechanism for signaling the occurrence of one of a set of events,
+// including PROVIDER_READY, PROVIDER_ERROR, PROVIDER_CONFIGURATION_CHANGED and PROVIDER_STALE,
+// with a provider event details payload.
+func TestEventHandler_RegisterUnregisterEventProvider(t *testing.T) {
+
+	t.Run("Ignored non-eventing providers", func(t *testing.T) {
```
toddbaert, line 25: :sunglasses:

pkg/openfeature/event_executor_test.go
```diff
@@ -0,0 +1,529 @@
...
+
+		eventingProvider := struct {
+			FeatureProvider
+			EventHandler
+		}{
+			NoopProvider{},
+			eventingImpl,
+		}
+
+		handler := newEventHandler(logger)
```
toddbaert, line 50: `hander` and `newEventHandler` are both confusing to me here. This is not a handler, and the function doesn't seem to return a `handler`... but an `eventExecutor`, which is very different conceptually from a handler.

Kavindu-Dodan: This is something missed from renaming [1] . will fix in the next commits 

[1] - https://github.com/open-feature/go-sdk/pull/188/files#diff-5088e3f82f28056163763bfb774c1b8e91bb585580b4b1add90b133445b8bae6R26-R33

pkg/openfeature/openfeature_test.go
```diff
@@ -329,6 +330,122 @@ func TestRequirement_1_6_1(t *testing.T) {
 	}
 }
 
+func TestRequirement_EventCompliance(t *testing.T) {
```
toddbaert, line 333: The `event_executor` test suite covers a lot, but I think we could also use a couple "integration" (not unit) tests here as well that actually run handlers and test some high level behavior.

A couple scenarios I can think of:

- client handler fires when provider is set after client is created.
- client handler fires when provider is set before client is created.
- named client (bound to default provider) handler fires when the default provider emits an event
- named client (bound to named provider) handler fires when the default provider emits an event

pkg/openfeature/provider.go
```diff
@@ -60,6 +65,62 @@ type StateHandler interface {
...
+	EventType
+	ProviderEventDetails
+}
+
+type EventDetails struct {
+	client string
+	ProviderEventDetails
+}
+
+type EventCallBack *func(details EventDetails)
```
toddbaert, line 114:
```suggestion
type EventCallback *func(details EventDetails)
```
I'd consider `handler` but callback is fine too. I would do whatever _feels_ more right for Go users.

If you do use callback though, note that it's one word (https://en.wikipedia.org/wiki/Callback_(computer_programming)).

pkg/openfeature/event_executor_test.go
Kavindu-Dodan: Rename completed

pkg/openfeature/openfeature_test.go
Kavindu-Dodan: Good suggestion. Some of these are spec requirements this implementation is missing. I am working to implement them.

pkg/openfeature/provider.go
Kavindu-Dodan: Thanks for lang improvement. Updated to `EventCallback `

pkg/openfeature/openfeature_test.go
Kavindu-Dodan: All of these are covered. But I think I will add more. Eventing has lots of corner cases and we need lots of tests to validate them.

odubajDT (approved):

pkg/openfeature/event_executor.go
```diff
@@ -0,0 +1,319 @@
...
+	for i, f := range entrySlice {
+		if f == c {
+			entrySlice = append(entrySlice[:i], entrySlice[i+1:]...)
+		}
+	}
+
+	e.scopedRegistry[name].callbacks[t] = entrySlice
+}
+
+// registerDefaultProvider register the default FeatureProvider and remove the old default provider if available
```
toddbaert, line 167:
```suggestion
// registerDefaultProvider registers the default FeatureProvider and remove the old default provider if available
```

pkg/openfeature/event_executor.go
```diff
@@ -0,0 +1,319 @@
...
+		(*c)(EventDetails{
+			provider: provider.metadata.Name,
+			ProviderEventDetails: ProviderEventDetails{
+				Message: "provider is at ready state",
+			},
+		})
+	}
+}
+
+// removeClientHandler remove client level handler
```
toddbaert, line 141:
```suggestion
// removeClientHandler removes client level handler
```

pkg/openfeature/event_executor.go
```diff
@@ -0,0 +1,319 @@
...
+
+// updateLogger updates the executor's logger
+func (e *eventExecutor) updateLogger(l logr.Logger) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	e.logger = l
+}
+
+// registerApiHandler add API(global) level handler
```
toddbaert, line 66:
```suggestion
// registerApiHandler adds an API(global) level handler
```

I've seen this in a couple places.

I don't know if there's a convention around this, but if you keep referencing the name in the godoc, you'll have to conjugate all these verbs with an "s". The other option is just to drop the method name in the godoc, I guess:

`registerApiHandler adds an API(global) level handler`

vs

`add an API(global) level handler`

Not sure which is better.

pkg/openfeature/event_executor.go
```diff
@@ -0,0 +1,319 @@
...
+		for _, c := range defaultHandlers {
+			e.executeHandler(*c, event)
+		}
+
+		return nil
+	})
+
+	// wait for completion or timeout
+	select {
+	case <-time.After(handlerExecutionTime):
```
toddbaert, line 296: This is not spec'd, and honestly it could be changed later, but I think a per-handler timeout would be more useful.

pkg/openfeature/provider.go
```diff
@@ -60,6 +65,62 @@ type StateHandler interface {
...
+	FlagChanges   []string
+	EventMetadata map[string]interface{}
+}
+
+// Event is an event emitted by a FeatureProvider.
+type Event struct {
+	ProviderName string
+	EventType
+	ProviderEventDetails
+}
```
toddbaert, line 107: Does this type need to be exported? Unlike `ProviderEventDetails` and `EventDetails`, it's not in the spec.

pkg/openfeature/provider.go
```diff
@@ -60,6 +65,62 @@ type StateHandler interface {
...
+
+type EventDetails struct {
+	provider string
```
toddbaert, line 110:
```suggestion
	providerName string
```

toddbaert (approved):
Looks good to me overall, with a few nits (assuming the 1:N issue is fixed elsewhere).

pkg/openfeature/openfeature_test.go
```diff
@@ -329,6 +330,122 @@ func TestRequirement_1_6_1(t *testing.T) {
...
+
+	// The client MUST provide a function for associating handler functions with a particular provider event type.
+	// The API and client MUST provide a function allowing the removal of event handlers.
+	t.Run("requirement_5_2_1 & requirement_5_2_1", func(t *testing.T) {
```
toddbaert, line 337: 5_2_1 & 5_2_1 seems like a typo?

pkg/openfeature/event_executor.go
Kavindu-Dodan: Referring the function/struct name as the first word (especially for exposed ones) is a convention in go [1]

I will revisit the comments and fix the grammar 

[1] - https://go.dev/doc/comment

pkg/openfeature/openfeature_test.go
Kavindu-Dodan: Good one, it should have been 5_2_1 & 5_2_2

pkg/openfeature/provider.go
Kavindu-Dodan: For go-sdk implementation, events are emitted from providers. And it's the OF implementation that handles this event. This event emitting is done through a channel that eventing providers expose [1] and this is the value type channel emits. This is why this needs to be exposed. 

[1] - https://github.com/open-feature/go-sdk/blob/8ee7d3ce4b1dc81669b2bdf0fe74e6ac6cf795c3/pkg/openfeature/provider.go#L89

pkg/openfeature/provider.go
Kavindu-Dodan: done with the latest change

pkg/openfeature/event_executor.go
Kavindu-Dodan: Updated with the latest commit

Kavindu-Dodan:
> Looks good to me overall, with a few nits (assuming the 1:N issue is fixed elsewhere).

correct, 1:N support was missing in this implementation and I am actively working on enabling this.

pkg/openfeature/event_executor.go
Kavindu-Dodan: I removed the timeout handling as this is not specified in the spec. further, seems other sdk implementations do not follow a similar approach, which make go-sdk to function differently

beeme1mr (approved):

PR #190, opened 2023-06-27
feat: provider client 1:n binding support

Kavindu-Dodan:
## This PR

A follow up PR to introduce 1:N support - [Check diff from main ](https://github.com/Kavindu-Dodan/go-sdk/compare/feat/event-impl...Kavindu-Dodan:go-sdk:chore/1-to-n-handling)

Fixes and completes https://github.com/open-feature/go-sdk/issues/186

A follow-up for the PR https://github.com/open-feature/go-sdk/pull/188

This PR brings,

- 1 to N binding of provider (a single provider to multiple client binding capability)
- completes eventing specification by adding API-level events for success/error initialization
- Improves the eventing implementation and adds more tests


pkg/openfeature/provider.go
```diff
@@ -70,8 +70,9 @@ type StateHandler interface {
 type NoopStateHandler struct {
 }
 
-func (s *NoopStateHandler) Init(e EvaluationContext) {
+func (s *NoopStateHandler) Init(e EvaluationContext) error {
 	// NOOP
+	return nil
 }
 
 func (s *NoopStateHandler) Shutdown() {
```
toddbaert, line 78: Does it makes sense to include an error return from shutdown as well? I guess the question would be what the SDK would do with such errors... just log them and shutdown anyway I suppose?

If that's the case, it might not help much. Just food for thought.

pkg/openfeature/api.go
```diff
@@ -145,21 +166,56 @@ func (api *evaluationAPI) forTransaction(clientName string) (FeatureProvider, []
...
-			if ok {
-				v.Shutdown()
+// initNewAndShutdownOld is a helper to initialise new FeatureProvider and shutdown the old FeatureProvider.
+// Operations happen concurrently.
+func (api *evaluationAPI) initNewAndShutdownOld(newProvider FeatureProvider, oldProvider FeatureProvider) {
+	v, ok := newProvider.(StateHandler)
+	if ok && v.Status() == NotReadyState {
+		go func(provider FeatureProvider, stateHandler StateHandler, evalCtx EvaluationContext, eventChan chan eventPayload) {
+			err := stateHandler.Init(evalCtx)
+			if err != nil {
```
toddbaert, line 176:
```suggestion
                        // emit ready/error event once initialization is complete
			if err != nil {
```

pkg/openfeature/event_executor_test.go
```diff
@@ -0,0 +1,1307 @@
...
+			},
+		}
+
+		// callback
+		rsp := make(chan EventDetails, 1)
+		callback := func(e EventDetails) {
+			rsp <- e
+		}
+
+		addClientHandler("someClient", ProviderReady, &callback)
```
toddbaert, line 412: These new tests are very comprehensive, they add a lot of confidence. I've noticed that with the client tests, we're always testing the private client methods. Do you think it might be better to extract the client tests into `client_test.go` and instead of calling the private `addClientHandler`, call the public equivalent on the client? Would that cause a lot of difficulty?

toddbaert (approved):
This is great. I don't see any issues from a spec perspective. The testing is very robust. I had a nit about some of the tests [here](https://github.com/open-feature/go-sdk/pull/190#discussion_r1252267385), but if this is more effort than it's worth I'm fine with leaving it as is, though perhaps with a comment we're testing a private API as a proxy for a public one.

pkg/openfeature/provider.go
Kavindu-Dodan: Good point. From SDK point of view, we can't do much with an error return. Like you said, the only extra step we can take is to log the error. 

I will leave it as it is If there is no strong opinion to add error for shutdown method signature

pkg/openfeature/event_executor_test.go
Kavindu-Dodan: Good suggestion. I tried to aggregate all eventing spec-bound tests [1] under `event_executor_test.go`. But yes, using internal methods is not the best choice here. 

I will do the following - given that we have client API level test coverage [2], I will avoid internal methods and migrate this registration to be done via a Client. This way we validate both client API as well as eventing functionality 

[1] - https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md
[2] - https://github.com/open-feature/go-sdk/pull/190/files#diff-1520c2162931fb780e1961a5cd73b7a724cbbbecea232fccb9384db38f4957c7R92-R101

Kavindu-Dodan: @toddbaert  thank you for highlighting this. 

I migrated tests to use API contracts where possible. For example, I am using internals only where it's neede (ex:- to check if registrations are correct or validations are there). However, I kept tests in the same file. This is to aggregate all eventing spec tests at one location

toddbaert: > @toddbaert thank you for highlighting this.
> 
> I migrated tests to use API contracts where possible. For example, I am using internals only where it's neede (ex:- to check if registrations are correct or validations are there). However, I kept tests in the same file. This is to aggregate all eventing spec tests at one location

That sounds perfect to me.

pkg/openfeature/event_executor.go
```diff
@@ -0,0 +1,381 @@
...
+	// then run client handlers
+	var nameAssociations []string
+	for name, reference := range e.namedProviderReference {
+		if reference.featureProvider == handler {
+			nameAssociations = append(nameAssociations, name)
+		}
+	}
+
+	for _, nameAssociation := range nameAssociations {
+		for _, c := range e.scopedRegistry[nameAssociation].callbacks[event.EventType] {
```
RealAnna, line 314: instead of going again through nameAssociations here, we could directly run the for at 314 in line 309

pkg/openfeature/event_executor.go
```diff
@@ -0,0 +1,381 @@
...
+	if e.defaultProviderReference.featureProvider != handler {
+		return
+	}
+
+	// handling the default provider - invoke default provider bound (no provider associated) handlers by filtering
+	var defaultHandlers []EventCallback
+
+	for clientName, registry := range e.scopedRegistry {
+		if _, ok := e.namedProviderReference[clientName]; !ok {
+			defaultHandlers = append(defaultHandlers, registry.callbacks[event.EventType]...)
```
RealAnna, line 328: probably we could do the same here?

RealAnna, line 328: probably we could also do the same here

pkg/openfeature/api.go
```diff
@@ -145,21 +166,57 @@ func (api *evaluationAPI) forTransaction(clientName string) (FeatureProvider, []
...
+	if oldProvider == nil || !ok {
+		return
+	}
+
+	// check for multiple bindings
+	if oldProvider == api.defaultProvider || contains(oldProvider, maps.Values(api.namedProviders)) {
+		return
+	}
+
+	go func(forShutdown StateHandler) {
```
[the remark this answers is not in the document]
toddbaert: The spec doesn't require this. Switching providers after startup is a bit of an edge-case... but when you set a new one, the thing that prevents you from using it pre-maturely is waiting the ready event, which will fire when the new provider is initialized.

pkg/openfeature/event_executor.go
Kavindu-Dodan: Good suggestion. I went with this extract and execute it for easy reading. But yes, direct execution on the same loop is also fine.

Update with the latest commit :)

pkg/openfeature/event_executor.go
Kavindu-Dodan: Update with the latest commit - https://github.com/open-feature/go-sdk/pull/190/files#diff-5088e3f82f28056163763bfb774c1b8e91bb585580b4b1add90b133445b8bae6R322-R330

RealAnna (approved):
LGTM!

beeme1mr (approved):

PR #192, opened 2023-07-07
feat: in-memory provider for sdk testing

Kavindu-Dodan:
## This PR

Is related to OFEP  https://github.com/open-feature/ofep/pull/71 and shows the feasibility of an in-memory provider

### Features 

- Provider is constructed with map of pre-defined flags [1]
- Supports context evaluations based on callbacks [2]

[1] - https://github.com/open-feature/ofep/pull/71
[2] - https://github.com/open-feature/go-sdk/pull/192/files#diff-1e56a5f9a72a9694aeb6f3510e8a2e02359c3cbdb9bd5c906219e30756d857efR181


e2e/caching_test.go
```diff
@@ -1,36 +0,0 @@
-package e2e_test
```
Kavindu-Dodan, line 1: Removed the caching test suite as this is purely flagd behavior test

e2e/evaluation_test.go
```diff
@@ -1,20 +1,29 @@
 package e2e_test
```
Kavindu-Dodan, line 1: This is a port from go test suite [1]

[1] -https://github.com/open-feature/go-sdk-contrib/blob/main/tests/flagd/pkg/integration/evaluation.go

e2e/common_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	if customer != false {
+		return defaultValue, defaultResolution
+	}
+
+	return this.Variants["internal"], openfeature.ProviderResolutionDetail{
+		Reason:  openfeature.TargetingMatchReason,
+		Variant: "internal",
+	}
```
toddbaert, line 58: I'm not sure this all needs to be so verbose, particularly because there's no specific logging for each assertion failure. I think you could just do the type assertion and equality check in a single step for each, no?

e2e/common_test.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+var memoryFlags = map[string]mp.InMemoryFlag{
```
toddbaert, line 61: This is pretty cool and simple.

pkg/openfeature/testing/in_memory_provider.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func (i InMemoryProvider) BooleanEvaluation(ctx context.Context, flag string, defaultValue bool, evalCtx openfeature.FlattenedContext) openfeature.BoolResolutionDetail {
+	memoryFlag, ok := i.flags[flag]
+	if !ok {
+		return openfeature.BoolResolutionDetail{
+			Value: defaultValue,
+			ProviderResolutionDetail: openfeature.ProviderResolutionDetail{
+				ResolutionError: openfeature.NewFlagNotFoundResolutionError(fmt.Sprintf("flag for key %s not found", flag)),
```
toddbaert, line 36: you could possibly extract the repeated not-found stuff into a func.

toddbaert (approved):
This is very straightforward. It's pretty much convinced me we should do this everywhere.

e2e/common_test.go
Kavindu-Dodan: Yeah, will rework this :)

pkg/openfeature/testing/in_memory_provider.go
Kavindu-Dodan: Attempted to extract the common logic  - https://github.com/open-feature/go-sdk/pull/192/files#diff-1e56a5f9a72a9694aeb6f3510e8a2e02359c3cbdb9bd5c906219e30756d857efR132-R143

e2e/common_test.go
Kavindu-Dodan: Done, reduced the complexity - https://github.com/open-feature/go-sdk/pull/192/files#diff-25f457711915ad116e8b162b1c338a4eeaf032cbe097e20d34de53b5ec4812d3R9-R35 :)

RealAnna (approved):

beeme1mr (approved):

Kavindu-Dodan:
> Perhaps we should add the `e2e` folder to the Codecov ignore list.

For this PR, coverage complaint is on newly added in-memory providers. Some of the error paths are not covered with tests but IMO these are not essential (ex:- testing errors for each flag type)

PR #193, opened 2023-07-11
chore: update readme for inclusion in the docs

beeme1mr:
## This PR

- Update the readme so it renders nicely in the docs

PR #204, opened 2023-07-27
feat: EvaluationContext add constructor without TargetingKey

thomaspoignant:
## This PR
Based on this [discussion](https://github.com/open-feature/go-sdk/pull/193#discussion_r1276491478) we have discovered that in some cases we want to create an EvaluationContext without Targeting key.

With`NewEvaluationContext` it was mandatory to set a `targetingKey` forcing us to set an empty `targetingKey` sometimes.

In this PR we add a new constructor called `NewTargetlessEvaluationContext` that is doing exactly the same things as `NewEvaluationContext` but does not force you to add a `targetingKey`.

### DIFF
--- a/pkg/openfeature/evaluation_context.go
+++ b/pkg/openfeature/evaluation_context.go
@@ -46,3 +46,10 @@ func NewEvaluationContext(targetingKey string, attributes map[string]interface{}
 		attributes:   attrs,
 	}
 }
+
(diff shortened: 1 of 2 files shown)


beeme1mr (approved):
Thanks @thomaspoignant! I'll update the readme when this is live

toddbaert (approved):
nice

PR #232, opened 2023-11-15
feat: Move all files from pkg/openfeature to openfeature.

### LINKED ISSUE #227: [FEATURE] Follow official Go guidance on how to organize a module?
### Requirements

Go released an official guidance on how to organize a module here: https://go.dev/doc/modules/layout. Do you think it is a good idea to follow this guidance for this repo? What it means here is to just move everything in `pkg/openfeature` to the top-level. IMHO it would make importing this module a bit nicer:
```
import "github.com/open-feature/go-sdk/pkg/openfeature"
```
vs
```
import "github.com/open-feature/go-sdk"
```

The mismatch of the package name and the import is a bit jarring though. I suppose one could also "import as".

Unfortunately, this would be a breaking change.

paddycarver:
## This PR

Move the package from being
github.com/open-feature/go-sdk/pkg/openfeature, which has an unnecessary pkg in the import path, to github.com/open-feature/go-sdk/openfeature, without the unnecessary "pkg" in the import path.

The existing github.com/open-feature/go-sdk/pkg/openfeature package is now a compatibility shell; exported types, constants, variables, and functions are now aliased to the new
github.com/open-feature/go-sdk/openfeature equivalents in a way that will pass equality checks, and types are interchangeable between both packages. No logic or tests live in the package anymore, only the exported identifiers, which call through to the new package for behavior.

The memprovider package, similarly, has moved.

### Related Issues

Fixes #227 

### How to test

Any consumers of the package that already exist should be able to transparently use this version (`go mod edit -replace github.com/open-feature/go-sdk=github.com/paddycarver/open-feature@no-pkg`) with no code changes or changes in behavior.


Kavindu-Dodan:
@paddycarver thank you for the contribution and I like the approach proposed through this PR (compared to #233 with root level content)

Could you please sign off your commit ? This guide should help you - https://github.com/open-feature/go-sdk/pull/232/checks?check_run_id=18694944964

paddycarver:
Should be done. :)

toddbaert:
@Kavindu-Dodan @paddycarver thanks so much for this change. I agreed very much with @craigpastro that the imports were not ideal, and I had already accepted that this would be a breaking change, so this was a welcome surprise!

There's a `!` in the title. This implies this is breaking, which I believe it is not (I don't consider deprecations breaking). Since everything should work transparently with the exception of compiler warnings around deprecations, I guess we can remove the `!`? Otherwise our release automation will try to increment the package version to `v2.0.0`.

toddbaert (approved):

AlexsJones:
What problem does this solve? It's good practice to use `pkg` in golang

beeme1mr:
@AlexsJones, the reasoning was discussed in the [linked issue](https://github.com/open-feature/go-sdk/issues/227). @paddycarver was able to introduce this change in a nonbreaking way.

AlexsJones:
If we can make the change without breaking things 👍

Kavindu-Dodan (approved):

PR #239, opened 2023-12-13
fix: wait for concurrent operations to finish in initNewAndShutdownOld

### LINKED ISSUE #238: [BUG] Panic when following getting started tutorial
### Observed behavior

Panic from inside the library.

### Expected Behavior

No panic.

### Steps to reproduce

This code panics:
```golang
package main

import (
	"context"
	"fmt"
	flagd "github.com/open-feature/go-sdk-contrib/providers/flagd/pkg"
)

func main() {
	provider := flagd.NewProvider(flagd.WithHost("localhost"), flagd.WithPort(8013))
        // Super duper asynchronous provider initialization magic is starting to happen.

	err := openfeature.SetProvider(provider)
        // We start using the provider variable, but the super duper magic is still happening and the provider is not ready for usage yet!!!
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	client := openfeature.NewClient("app")

	ctx := context.Background()
	evalCtx := openfeature.EvaluationContext{}
        // Is the super duper magic over yet? Of course not. Panic.
	v2Enabled, err := client.BooleanValue(ctx, "v2_enabled", true, evalCtx) // This line of code panics.
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	if v2Enabled {
		fmt.Println("v2 is enabled")
	}
}
```

This code does not. The only difference is the added time.Sleep.
```golang
package main

import (
	"context"
	"fmt"
	flagd "github.com/open-feature/go-sdk-contrib/providers/flagd/pkg"
)

func main() {
	provider := flagd.NewProvider(flagd.WithHost("localhost"), flagd.WithPort(8013))
        // Super duper asynchronous provider initialization magic is starting to happen.

       time.Sleep(time.Second*3)
       // Actually wait for the super duper asynchronous initialization to finish, as if it was actually synchronous like it's supposed to be.

	err := openfeature.SetProvider(provider)
        // We start using the provider variable, and the super duper magic is actually over and nothing bad is going to happen.
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	client := openfeature.NewClient("app")

	ctx := context.Background()
	evalCtx := openfeature.EvaluationContext{}
        // Is the super duper magic over yet? Of course! We used time.Sleep to circumvent the library's buggy behaviour.
	v2Enabled, err := client.BooleanValue(ctx, "v2_enabled", true, evalCtx) // This line of code does not panic.
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	if v2Enabled {
		fmt.Println("v2 is enabled") // Output: "v2 is enabled"
	}
}
```

The issue is `github.com/open-feature/go-sdk@v1.8.0/pkg/openfeature/api.go:174`.

For some reason, library/provider initialization is asynchronous, meaning one can use the library/provider without any guarantee that it is ready for usage, which goes against all common sense. The fix is to not do async initializaiton. I will link the pull request shortly.

i-fix-retarded-bugs:
## This PR
- This PR fixes panic during library initialization by waiting for provider initialization to finish.

### Related Issues
Fixes #238

### How to test
Check that the panicking code listed in the issue no longer panics. It does not.

### DIFF
--- a/openfeature/api.go
+++ b/openfeature/api.go
@@ -167,11 +167,11 @@ func (api *evaluationAPI) forTransaction(clientName string) (FeatureProvider, []
 }
 
 // initNewAndShutdownOld is a helper to initialise new FeatureProvider and shutdown the old FeatureProvider.
-// Operations happen concurrently.
+// Operations happen NOT concurrently.
 func (api *evaluationAPI) initNewAndShutdownOld(newProvider FeatureProvider, oldProvider FeatureProvider) {
 	v, ok := newProvider.(StateHandler)
 	if ok && v.Status() == NotReadyState {
-		go func(provider FeatureProvider, stateHandler StateHandler, evalCtx EvaluationContext, eventChan chan eventPayload) {
+		func(provider FeatureProvider, stateHandler StateHandler, evalCtx EvaluationContext, eventChan chan eventPayload) {
 			err := stateHandler.Init(evalCtx)
 			// emit ready/error event once initialization is complete
 			if err != nil {
@@ -206,7 +206,7 @@ func (api *evaluationAPI) initNewAndShutdownOld(newProvider FeatureProvider, old
 		return
 	}
 
-	go func(forShutdown StateHandler) {
+	func(forShutdown StateHandler) {
 		forShutdown.Shutdown()
 	}(v)
 }


toddbaert:
Hey @i-fix-retarded-bugs , please see my comment [here](https://github.com/open-feature/go-sdk/issues/238#issuecomment-1854409704).

We definitely have some doc shortcomings here, and we certainly wouldn't expect any panics, so there may be a bug somewhere here. Additionally, we would expect the usage of a READY handler here to make sure the provider has initialized (that's the shortcoming in our tutorial).

However, we can't make this function sync. We could add a new function that behaves this way though (we have something imilar in other SDKs). If you'd like to implement that, we'd accept that change (pending that you update all references to your username as mentioned).

toddbaert:
Indeed, the panic here is the result of a bug in the flagd provider. I've opened an issue for that: https://github.com/open-feature/go-sdk-contrib/issues/407

toddbaert:
Closing. See https://github.com/open-feature/go-sdk/issues/238#issuecomment-1854409704 and linked issues.

PR #244, opened 2023-12-29
feat: Blocking provider mutator

### LINKED ISSUE #240: [FEATURE] Blocking provider mutator
### Requirements

As is recommended by the spec, we should add a blocking version of setProvider. This would wait for the provider to be initialized before it returns. It should be named in such a way that its blocking behavior is obvious, and should be documented appropriately. The README should be update to include it (close to the start) and highlight how it behaves differently from `setProvider`.

See a [PR for a similar feature in Java](https://github.com/open-feature/java-sdk/pull/563).

Please be sure to add tests!

Relates to: https://github.com/open-feature/go-sdk/issues/238

znonthedev:
## This PR
Adds setProviderAndWait method in the API that will set the provider synchronously  

- adds this new feature

### Related Issues

Fixes #240

### Follow-up Tasks
add test scenarios
update readme

### DIFF
--- a/openfeature/api.go
+++ b/openfeature/api.go
@@ -211,6 +211,46 @@ func (api *evaluationAPI) initNewAndShutdownOldAsync(newProvider FeatureProvider
 	}(v)
 }
 
+// initNewAndShutdownOldSync is a helper to initialise new FeatureProvider and shutdown the old FeatureProvider.
+// Operations happen synchronously.
+func (api *evaluationAPI) initNewAndShutdownOldSync(newProvider FeatureProvider, oldProvider FeatureProvider) {
+	v, ok := newProvider.(StateHandler)
+	if ok && v.Status() == NotReadyState {
+		err := v.Init(api.evalCtx)
+		if err != nil {
+			api.eventExecutor.eventChan <- eventPayload{
+				Event{
+					ProviderName:         newProvider.Metadata().Name,
+					EventType:            ProviderError,
+					ProviderEventDetails: ProviderEventDetails{},
+				}, newProvider,
+			}
+		} else {
+			api.eventExecutor.eventChan <- eventPayload{
+				Event{
+					ProviderName:         newProvider.Metadata().Name,
+					EventType:            ProviderReady,
+					ProviderEventDetails: ProviderEventDetails{},
+				}, newProvider,
+			}
+		}
+	}
+
+	v, ok = oldProvider.(StateHandler)
+
+	// oldProvider can be nil or without state handling capability
+	if oldProvider == nil || !ok {
+		return
+	}
+
+	// check for multiple bindings
+	if oldProvider == api.defaultProvider || contains(oldProvider, maps.Values(api.namedProviders)) {
+		return
+	}
+
+	v.Shutdown()
+}
+
 func contains(provider FeatureProvider, in []FeatureProvider) bool {
 	for _, p := range in {
 		if provider == p {
--- a/openfeature/api.go
+++ b/openfeature/api.go
@@ -58,6 +58,30 @@ func (api *evaluationAPI) setProvider(provider FeatureProvider) error {
 	return nil
 }
 
+// setProviderAndWait sets the default FeatureProvider of the evaluationAPI. Returns an error if FeatureProvider is nil
+// This is a blocking call and will wait for the provider to be ready
+func (api *evaluationAPI) setProviderAndWait(provider FeatureProvider) error {
+	api.mu.Lock()
+	defer api.mu.Unlock()
+
+	if provider == nil {
+		return errors.New("default provider cannot be set to nil")
+	}
+
+	// Initialize new default provider and shutdown the old one
+	// Provider update must be non-blocking, hence initialization & shutdown happens concurrently
+	oldProvider := api.defaultProvider
+	api.defaultProvider = provider
+
+	api.initNewAndShutdownOldSync(provider, oldProvider)
+	err := api.eventExecutor.registerDefaultProvider(provider)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // getProvider returns the default FeatureProvider
 func (api *evaluationAPI) getProvider() FeatureProvider {
 	api.mu.RLock()
--- a/openfeature/api.go
+++ b/openfeature/api.go
@@ -91,6 +91,7 @@ func (api *evaluationAPI) getProvider() FeatureProvider {
 }
 
 // setProvider sets a provider with client name. Returns an error if FeatureProvider is nil
+// Provider initialization is asynchronous and status can be checked from provider status
 func (api *evaluationAPI) setNamedProvider(clientName string, provider FeatureProvider) error {
 	api.mu.Lock()
 	defer api.mu.Unlock()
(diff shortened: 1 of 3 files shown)


znonthedev:
Hi @toddbaert, could you please review this PR? I am very new to Go and not sure if this implementation is correct. I will refactor it and keep it DRY if the current implementation meets the expectations outlined in the issue. Thanks in advance.

toddbaert:
Thank you @znonthedev !

I will review, and also add some reviewers who are more experienced with Go than me! Many people are on holiday however, so it may take a few days to get some approvals. Your patience is appreciated! :pray:

openfeature/api.go
```diff
@@ -211,6 +235,46 @@ func (api *evaluationAPI) initNewAndShutdownOld(newProvider FeatureProvider, old
...
+			}
+		} else {
+			api.eventExecutor.eventChan <- eventPayload{
+				Event{
+					ProviderName:         newProvider.Metadata().Name,
+					EventType:            ProviderReady,
+					ProviderEventDetails: ProviderEventDetails{},
+				}, newProvider,
+			}
+		}
```
Kavindu-Dodan, line 260: This is duplicate of the same calls of https://github.com/open-feature/go-sdk/blob/main/openfeature/api.go#L174-L194 

Instead of duplication, you can extract this code to a dedicated function (ex:- `func initHelper(...)`) . Then you can call this function as a go routine for async operation (L174-L194 ) and as a direct function call for sync operation.

openfeature/openfeature.go
```diff
@@ -23,6 +23,10 @@ func SetProvider(provider FeatureProvider) error {
 	return api.setProvider(provider)
 }
 
+func SetProviderAndWait(provider FeatureProvider) error {
```
Kavindu-Dodan, line 26: +1 

I am fine with this method of naming. Please add a doc comment explaining the functionality. Also, please update `SetProvider` (L20-L24) doc comment mention that it is asynchronous (non-blocking)

znonthedev: Added doc comments

openfeature/api.go
znonthedev: Refactored and avoids duplication

znonthedev:
I have updated the PR based on the comments. I have added two test cases also. I am not sure on the test cases. Do I need to add more test cases or change the current ones

toddbaert:
We need to add an equivalent function for setting a named provider and waiting (see [here](https://github.com/open-feature/go-sdk/pull/244/files#diff-434e8d62e6a56910999195bb04f7c78e5373efd76288dc4039cd58686c99ae8bR94)).

toddbaert:
@bacherfl when you get a moment, could you take a look at this as well?

znonthedev:
> We need to add an equivalent function for setting a named provider and waiting (see [here](https://github.com/open-feature/go-sdk/pull/244/files#diff-434e8d62e6a56910999195bb04f7c78e5373efd76288dc4039cd58686c99ae8bR94)).

Certainly, @toddbaert. I had the same thought. I will incorporate the necessary changes into the PR

openfeature/client_test.go
```diff
@@ -749,6 +749,87 @@ func TestRequirement_1_4_13(t *testing.T) {
 // Is tested by TestRequirement_4_4_2.
 
 // TODO Requirement_1_6_1
+
+func TestWaitForProvider(t *testing.T) {
```
toddbaert, line 753: It might be nice to add a test that works like this:

- set a provider which is ready after 1s (has an init function that returns after 1s)
- returns a valid flag value after that time only
- returns an error before 1s

This is consistent with how we'd expect many providers to work. Maybe you could add some options to `NewMockFeatureProvider` that would support this?

znonthedev: okay, I got the idea. I'll give it a try, but since I'm new to Go, it might take me some time 😇 .

openfeature/__debug_bin1610825419
Kavindu-Dodan, line 1: please remove this file

openfeature/api.go
```diff
@@ -166,32 +216,35 @@ func (api *evaluationAPI) forTransaction(clientName string) (FeatureProvider, []
...
+				ProviderEventDetails: ProviderEventDetails{},
+			}, provider,
+		}
+	}
+}
+
+// initNewAndShutdownOldAsync is a helper to initialise new FeatureProvider and shutdown the old FeatureProvider.
 // Operations happen concurrently.
-func (api *evaluationAPI) initNewAndShutdownOld(newProvider FeatureProvider, oldProvider FeatureProvider) {
+func (api *evaluationAPI) initNewAndShutdownOldAsync(newProvider FeatureProvider, oldProvider FeatureProvider) {
```
Kavindu-Dodan, line 244: You may refactor to combine `initNewAndShutdownOldAsync` & `initNewAndShutdownOldSync` as they perform similar set of operations with only difference being sync vs async.

openfeature/api.go
```diff
@@ -49,7 +49,31 @@ func (api *evaluationAPI) setProvider(provider FeatureProvider) error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// setProviderAndWait sets the default FeatureProvider of the evaluationAPI. Returns an error if FeatureProvider is nil
+// This is a blocking call and will wait for the provider to be ready
+func (api *evaluationAPI) setProviderAndWait(provider FeatureProvider) error {
```
bacherfl, line 63: Nitpick: to stay consistent with the naming i would suggest to call this function `setProviderSync`

toddbaert: Agreed. The spec doesn't dictate function names. I think your suggestion makes sense here.

openfeature/client_test.go
bacherfl: I just had a look at the test, and to make sure the `Init` function actually gets called, you will need to have a type that consolidates a `MockFeatureProvider` and `MockStateHandler` - otherwise the check [here](https://github.com/open-feature/go-sdk/pull/244/files#diff-434e8d62e6a56910999195bb04f7c78e5373efd76288dc4039cd58686c99ae8bR270) will fail and the function will proceed without calling the `Init` function of the provider.

Since gomock, which we use for creating the mocks seems to not support the generation of a type that implements multiple interfaces, i believe the simplest way to get a struct that implements both the FeatureProvider and the StateHandler interfaces is to consolidate them into a single struct, such as follows:

```go
type mockProviderAndStateHandler struct {
	*MockFeatureProvider
	*MockStateHandler
}
```

This type should be defined in the `client_test.go` file, and can then be used to pass it to the `SetProviderAndWait` function, such as in the following example:

```
...
		mockProvider.EXPECT().Metadata().AnyTimes()
		mockProvider.EXPECT().Hooks().AnyTimes()
		mockProvider.EXPECT().IntEvaluation(context.Background(), flagKey, defaultValue, flatCtx).
			Return(IntResolutionDetail{
				Value: 0,
				ProviderResolutionDetail: ProviderResolutionDetail{
					ResolutionError: NewGeneralResolutionError("test"),
				},
			}).Times(2)

		mockStateHandler := NewMockStateHandler(ctrl)

                // TODO add the desired behavior for the Init function of the MockStateHandler

		o := & mockProviderAndStateHandler{
			MockFeatureProvider: mockProvider,
			MockStateHandler:    mockStateHandler,
		}

		err := SetProviderAndWait(o)
...
```

Kavindu-Dodan:
@znonthedev are you planning to complete this contribution soon ? There are pending changes in some providers which require this feature.

toddbaert:
Closed with https://github.com/open-feature/go-sdk/pull/251

PR #251, opened 2024-02-02
feat: blocking provider mutator

### LINKED ISSUE #240: [FEATURE] Blocking provider mutator

Kavindu-Dodan:
## This PR

- Add `SetProviderAndWait` & `SetNamedProviderAndWait` with blocking provider mutators
- Improve non-eventing provider registration by emitting ready event : This can be considered as a fix as this is defined in spec but was missing in the implementation [^1]

> The provider MAY define a status field/accessor which indicates the readiness of the provider, with possible values NOT_READY, READY, STALE, or ERROR.

> Providers without this field can be assumed to be ready immediately.

### Related Issues

Fixes #240

### How to test

Tests were added to cover new APIs and old tests were updated to handle readiness emitted for non eventing providers 

[^1]: https://openfeature.dev/specification/sections/providers#requirement-242

### DIFF
--- a/openfeature/api.go
+++ b/openfeature/api.go
@@ -169,9 +169,9 @@ func (api *evaluationAPI) forTransaction(clientName string) (FeatureProvider, []
 // initNewAndShutdownOld is a helper to initialise new FeatureProvider and shutdown the old FeatureProvider.
 func (api *evaluationAPI) initNewAndShutdownOld(newProvider FeatureProvider, oldProvider FeatureProvider, async bool) {
 	if async {
-		go func() {
-			api.eventExecutor.triggerEvent(initializer(newProvider, api.apiCtx), newProvider)
-		}()
+		go func(executor *eventExecutor, ctx EvaluationContext) {
+			executor.triggerEvent(initializer(newProvider, ctx), newProvider)
+		}(api.eventExecutor, api.apiCtx)
 	} else {
 		api.eventExecutor.triggerEvent(initializer(newProvider, api.apiCtx), newProvider)
 	}
--- a/openfeature/api.go
+++ b/openfeature/api.go
@@ -49,8 +49,12 @@ func (api *evaluationAPI) setProvider(provider FeatureProvider, async bool) erro
 	oldProvider := api.defaultProvider
 	api.defaultProvider = provider
 
-	api.initNewAndShutdownOld(provider, oldProvider, async)
-	err := api.eventExecutor.registerDefaultProvider(provider)
+	err := api.initNewAndShutdownOld(provider, oldProvider, async)
+	if err != nil {
+		return err
+	}
+
+	err = api.eventExecutor.registerDefaultProvider(provider)
 	if err != nil {
 		return err
 	}
@@ -80,8 +84,12 @@ func (api *evaluationAPI) setNamedProvider(clientName string, provider FeaturePr
 	oldProvider := api.namedProviders[clientName]
 	api.namedProviders[clientName] = provider
 
-	api.initNewAndShutdownOld(provider, oldProvider, async)
-	err := api.eventExecutor.registerNamedEventingProvider(clientName, provider)
+	err := api.initNewAndShutdownOld(provider, oldProvider, async)
+	if err != nil {
+		return err
+	}
+
+	err = api.eventExecutor.registerNamedEventingProvider(clientName, provider)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


openfeature/api.go
```diff
@@ -211,6 +193,31 @@ func (api *evaluationAPI) initNewAndShutdownOld(newProvider FeatureProvider, old
...
+	handler, ok := provider.(StateHandler)
+	if !ok {
+		// Note - a provider without state handling capability can be assumed to be ready immediately.
+		return event
+	}
+
+	err := handler.Init(apiCtx)
+	if err != nil {
+		event.EventType = ProviderError
+		event.Message = fmt.Sprintf("Provider initialization error, %v", err)
```
thisthat, line 215: I think we should also clean up the `ProviderEventDetails` property otherwise it will be quite weird to fail successfully 🤣

toddbaert: image

Kavindu-Dodan: @thisthat  good concern & @toddbaert  nice one 😅 

Luckily in OF go-sdk, this is handled correctly as we use go struct embedding [1] where `message` updated in error is from `ProviderEventDetails`

image

I know this is little confusing but I wanted to avoid multiple event initializations and went ahead with this. Let me know if you think we need to change this :) 

[1] - https://github.com/open-feature/go-sdk/blob/main/openfeature/provider.go#L105-L109

thisthat: oohh gotcha! All good for me then 👍

thisthat (approved):

openfeature/api.go
toddbaert: I always forget about struct embedding!

openfeature/api.go
```diff
@@ -28,28 +29,27 @@ func newEvaluationAPI() evaluationAPI {
...
 		logger:          logger,
 		mu:              sync.RWMutex{},
 		eventExecutor:   newEventExecutor(logger),
 	}
 }
 
-// setProvider sets the default FeatureProvider of the evaluationAPI. Returns an error if FeatureProvider is nil
-func (api *evaluationAPI) setProvider(provider FeatureProvider) error {
+// setProvider sets the default FeatureProvider of the evaluationAPI.
+// Returns an error if provider registration cause an error
```
toddbaert, line 40:
```suggestion
// Returns an error if provider registration or initialization fails
```

openfeature/openfeature.go
```diff
@@ -20,13 +20,23 @@ func initSingleton() {
 // SetProvider sets the default provider. Provider initialization is asynchronous and status can be checked from
 // provider status
 func SetProvider(provider FeatureProvider) error {
-	return api.setProvider(provider)
+	return api.setProvider(provider, true)
+}
+
+// SetProviderAndWait sets the default provider and waits for its initialization.
+func SetProviderAndWait(provider FeatureProvider) error {
+	return api.setProvider(provider, false)
```
toddbaert, line 28: This function needs to return an error if init fails (and I think you should add documentation saying so)

https://openfeature.dev/specification/sections/flag-evaluation/#requirement-1124

toddbaert (changes requested):
This seems good, code is clean. I think my only issue is that the set+wait function should return errors if the provider init fails, that's a key difference between the blocking and non-blocking setProvider functions.

Returning an error allows users to completely sidestep setting up error handers (though they can still do so optionally and we'd expect them to fire).

I think you should implement this and add a couple tests:

- one that returns an error for an init after some delay
- one that returns an error and also ensures error handlers are run.

openfeature/openfeature.go
Kavindu-Dodan: Good catch, I will improve this

Kavindu-Dodan: Done - https://github.com/open-feature/go-sdk/pull/251/commits/54ef5b0457b26309d3d4215eac8abc148de1a615

Kavindu-Dodan:
> This seems good, code is clean. I think my only issue is that the set+wait function should return errors if the provider init fails, that's a key difference between the blocking and non-blocking setProvider functions.
> 
> Returning an error allows users to completely sidestep setting up error handers (though they can still do so optionally and we'd expect them to fire).
> 
> I think you should implement this and add a couple tests:
> 
> * one that returns an error for an init after some delay
> * one that returns an error and also ensures error handlers are run.
> 
> Relavant PR language improvement (editorial, not significant): [open-feature/spec#242](https://github.com/open-feature/spec/pull/242)

Done, addressed this concern in the latest commit. I added error return as well as test to validate error return as well as eventing

openfeature/openfeature_test.go
```diff
@@ -252,6 +253,146 @@ func TestRequirement_1_1_2_3(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// then - must not update variable as registration is async
+		if initialized != false {
+			t.Errorf("expected uninitialized as async, but got true")
+		}
+	})
+}
```
toddbaert, line 394: beautiful!

toddbaert (approved):

beeme1mr (approved):

PR #252, opened 2024-02-05
feat: update to go 1.19

Kavindu-Dodan:
## This PR

Update Go version from `1.18` to `1.19`. 

Go 1.22 rc2 is alrady out and this is an attempt to gradually upgrade to go 1.20, keeping support for the last two major versions.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -7,7 +7,7 @@ env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
   # Default minimum version of Go to support.
-  DEFAULT_GO_VERSION: 1.18
+  DEFAULT_GO_VERSION: 1.19
 
 jobs:
   release-please:
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@
 
 ### Requirements
 
-- Go 1.18+
+- Go 1.19+
 
 ### Install
 
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,12 @@
 module github.com/open-feature/go-sdk
 
-go 1.18
+go 1.19
 
 require (
 	github.com/cucumber/godog v0.14.0
 	github.com/go-logr/logr v1.4.1
 	github.com/golang/mock v1.6.0
-	golang.org/x/exp v0.0.0-20230811145659-89c5cff77bcb
+	golang.org/x/exp v0.0.0-20240205201215-2c58cdc269a3
 	golang.org/x/text v0.14.0
 )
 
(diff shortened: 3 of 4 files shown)


thomaspoignant:
Why GO 1.19 and not go 1.20?
End of life of 1.19 was 5 month ago.

https://endoflife.date/go

thisthat (approved):

toddbaert:
@Kavindu-Dodan we can't mark this as `feat!`, it will force a 2.0 which is really nasty to deal with. We can consider upgrading the min version OFF of unsupported go version non-breaking.

Kavindu-Dodan:
> @Kavindu-Dodan we can't mark this as `feat!`, it will force a 2.0 which is really nasty to deal with. We can consider upgrading the min version OFF of unsupported go version non-breaking.... but I agree we should go to 1.20 I think.

Good point. Yes I remember this happening before.

Still I think conventional commit messages and titles are valuble. Shall we attempt this strategy where we can enforce release version through main branch commit - https://github.com/googleapis/release-please/tree/main?tab=readme-ov-file#how-do-i-change-the-version-number ? 

Otherwise I will revert the title :)

Kavindu-Dodan:
> Why GO 1.19 and not go 1.20? End of life of 1.19 was 5 month ago.
> 
> https://endoflife.date/go

Agree. I did not want to bump two major versions and this is why the update to 1.19. Soon I think we will update to 1.20 :)

toddbaert:
> > @Kavindu-Dodan we can't mark this as `feat!`, it will force a 2.0 which is really nasty to deal with. We can consider upgrading the min version OFF of unsupported go version non-breaking.... but I agree we should go to 1.20 I think.
> 
> Good point. Yes I remember this happening before.
> 
> Still I think conventional commit messages and titles are valuble. Shall we attempt this strategy where we can enforce release version through main branch commit - https://github.com/googleapis/release-please/tree/main?tab=readme-ov-file#how-do-i-change-the-version-number ?
> 
> Otherwise I will revert the title :)

I think you should just revert the title, I think people will not be surprised to see this support drop and it will show up clearly in the release notes (just my 2c).

thomaspoignant:
People using GO are migrating easily to other version I am not sure a new major version is needed.
Especially because if GO version is not compatible go will not let you update it in your go.mod so no risk to build something not compatible.

Kavindu-Dodan:
> > > @Kavindu-Dodan we can't mark this as `feat!`, it will force a 2.0 which is really nasty to deal with. We can consider upgrading the min version OFF of unsupported go version non-breaking.... but I agree we should go to 1.20 I think.
> > 
> > 
> > Good point. Yes I remember this happening before.
> > Still I think conventional commit messages and titles are valuble. Shall we attempt this strategy where we can enforce release version through main branch commit - https://github.com/googleapis/release-please/tree/main?tab=readme-ov-file#how-do-i-change-the-version-number ?
> > Otherwise I will revert the title :)
> 
> I think you should just revert the title, I think people will not be surprised to see this support drop and it will show up clearly in the release notes (just my 2c).

Done ✅

toddbaert (approved):
I don't feel too strongly in any direction, but I'm OK with releasing this and then upgrading again to 1.20 when 1.22 is released in Feb

Kavindu-Dodan:
> People using GO are migrating easily to other version I am not sure a new major version is needed. Especially because if GO version is not compatible go will not let you update it in your go.mod so no risk to build something not compatible.

@thomaspoignant  agree and we are not making a major release. I have updated the title to avoid trigger this. 

Beside, I created https://github.com/open-feature/go-sdk/issues/253 to track 1.20 upgrade. I think we should do this towards end of February 2024.

PR #263, opened 2024-04-04
feat: Adding simplified evaluation methods

### LINKED ISSUE #262: [BUG] Awkward usage when trying to evaluate a boolean flag in a complex conditional
### Observed behavior

(Related to #259, but that one was closed, so filing this instead)

I just ran into a papercut when working on an experiment to move from an in-house feature management system to OpenFeature. The previous system allowed calling a simple `IsEnabled(flag string) bool` method, which I'd like to replace with a call to `client.BooleanValue`. Mostly this works fine, but I just ran into an awkward case for some code with a fairly convoluted conditional:

```go
if cond1 && (cond2 || (cond3 && features.IsEnabled(ctx, "myflag"))) {
...
```

In this case it's impossible to simply replace the `features.IsEnabled` call with `client.BooleanValue` due to the `err` return value. And I'd rather avoid evaluating the flag altogether if `cond1` is false.

This isn't insurmountable, but it does make it more awkward to use OpenFeature in this case, and will cause adoption friction.

### Expected Behavior

The package should be less awkward to integrate with an existing project.

Could an additional set of methods be considered like these?

```go
func (c *Client) MustBoolean(ctx context.Context, flag string, defaultValue bool, evalCtx EvaluationContext, options ...Option) bool

func (c *Client) MustFloat(ctx context.Context, flag string, defaultValue float64, evalCtx EvaluationContext, options ...Option) float64

func (c *Client) MustInt(ctx context.Context, flag string, defaultValue int64, evalCtx EvaluationContext, options ...Option) int64

func (c *Client) MustObject(ctx context.Context, flag string, defaultValue interface{}, evalCtx EvaluationContext, options ...Option) interface{}

func (c *Client) MustString(ctx context.Context, flag string, defaultValue string, evalCtx EvaluationContext, options ...Option) string
```

### Steps to reproduce

(See above)

hairyhenderson:
## This PR

- adds new client methods `Boolean`, `String`, `Float`, `Int`, and `Object` that enable simpler usage (no error handling required)

### Related Issues

Fixes #262

### Notes

There seems to be a soft consensus on #262, so this may need to change if the TC deems necessary, but I'm submitting this early for discussion's sake.

### How to test

See additional examples provided.

### DIFF
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -28,6 +28,12 @@ type IClient interface {
 	FloatValueDetails(ctx context.Context, flag string, defaultValue float64, evalCtx EvaluationContext, options ...Option) (FloatEvaluationDetails, error)
 	IntValueDetails(ctx context.Context, flag string, defaultValue int64, evalCtx EvaluationContext, options ...Option) (IntEvaluationDetails, error)
 	ObjectValueDetails(ctx context.Context, flag string, defaultValue interface{}, evalCtx EvaluationContext, options ...Option) (InterfaceEvaluationDetails, error)
+
+	Boolean(ctx context.Context, flag string, defaultValue bool, evalCtx EvaluationContext, options ...Option) bool
+	String(ctx context.Context, flag string, defaultValue string, evalCtx EvaluationContext, options ...Option) string
+	Float(ctx context.Context, flag string, defaultValue float64, evalCtx EvaluationContext, options ...Option) float64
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan (approved):
Approving as I think this adds good syntatic sugar to our client API.

But please consider opinion of our TC (cc - @toddbaert @thomaspoignant @lukas-reining )

thomaspoignant (approved):
Sounds good to me, it will simplify the usage of the SDK 🙌

PR #268, opened 2024-04-29
feat: isolate interfaces from SDK to improve testability 

### LINKED ISSUE #266: [FEATURE] New test-oriented client
### Requirements

Right now it's a bit awkward to write unit tests that depend on different values being injected. This is largely because of the singleton evaluation API - if a new default provider is set with some flag values set a certain way, that may cause a test running concurrently to fail.

One option for getting around this is to use named providers and clients, but that can be awkward, and adds the requirement of tracking arbitrary strings for specific tests.

I propose a new `test` package (`github.com/open-feature/go-sdk/openfeature/test`) that would contain a client (conforming to `openfeature.IClient`) with a directly-attached in-memory provider to make it simpler to write unit tests.

_(See related [Slack conversation](https://cloud-native.slack.com/archives/C06HAN9KP9Q/p1712934184385439))_

Kavindu-Dodan:
## This PR

Partially address #266 by providing interface contracts for the SDK. There are no breaking changes in this change

I have introduced `IEvaluation` interface, which represents the general OpenFeatuer API layer [1]. Before this change, there was no such interface so mocking API interactions was impossible. To obtain the singleton instance of `IEvaluation` interface, I have introduced `GetApiInstance()` helper.

Consider the example below,

```go
// get the global instance that may be injected into testable components 
apiInstance := openfeature.GetApiInstance()

// use for OpenFeature operations 
apiInstance.SetProvider(myProvider)
apiInstance.AddHandler(....)
```
 In the future, we can update our documentation to make API instance based operations the preferred interaction method. 

Further, `IEvaluation` exposes client creation methods `GetClient()` & `GetNamedClient(clientName string)` (similar to spec [2]). These methods return `IClient` instances, making it easy to test OpenFeature client related operations. 

```go
apiInstance := openfeature.GetApiInstance()

// IClient instance, which can be mocked when injected into preferred locations 
client := apiInstance.GetClient()
```

[1] - https://openfeature.dev/specification/sections/flag-evaluation
[2] - https://openfeature.dev/specification/sections/flag-evaluation#creating-clients


openfeature/interfaces.go
```diff
@@ -0,0 +1,50 @@
+package openfeature
+
+import "context"
+
+// IOFApi defines the OpenFeature API contract
+type IOFApi interface {
```
hairyhenderson, line 6: This name is not very idiomatic. Can we call it something like `Evaluator` (since it wants to model the Evaluation API?)

openfeature/openfeature_api.go
```diff
@@ -6,81 +6,70 @@ import (
...
 
 	"github.com/go-logr/logr"
-	"github.com/open-feature/go-sdk/openfeature/internal"
 	"golang.org/x/exp/maps"
 )
 
-// evaluationAPI wraps OpenFeature evaluation API functionalities
-type evaluationAPI struct {
+// ofApiImpl is an internal reference interface extending IOFApi
+type ofApiImpl interface {
```
hairyhenderson, line 13: I'm not clear on why this intermediate interface is necessary - can you clarify why it exists?

openfeature/openfeature_api.go
```diff
@@ -6,81 +6,70 @@ import (
...
+	IOFApi
+	GetProvider() FeatureProvider
+	GetNamedProviders() map[string]FeatureProvider
+	GetHooks() []Hook
+	SetLogger(l logr.Logger)
+	ForEvaluation(clientName string) (FeatureProvider, []Hook, EvaluationContext)
+}
+
+// EvaluationAPI wraps OpenFeature evaluation API functionalities
+type EvaluationAPI struct {
```
hairyhenderson, line 23: Is it necessary to export this type as public API? Maybe it would be better to leave it unexported so as not to create undesired dependencies

openfeature/openfeature_api.go
```diff
@@ -6,81 +6,70 @@ import (
...
-	eventExecutor   *eventExecutor
 }
 
-// newEvaluationAPI is a helper to generate an API. Used internally
-func newEvaluationAPI() evaluationAPI {
-	logger := logr.New(internal.Logger{})
-
-	return evaluationAPI{
+// NewEvaluationAPI is a helper to generate an API. Used internally
+func NewEvaluationAPI(eventExecutor *EventExecutor, log logr.Logger) *EvaluationAPI {
```
hairyhenderson, line 34: if this function is used internally, why is it exported?

openfeature/event_executor.go
```diff
@@ -42,42 +57,50 @@ func newEventExecutor(logger logr.Logger) *eventExecutor {
...
 }
 
+type eventPayload struct {
+	event   Event
+	handler FeatureProvider
+}
+
 // providerReference is a helper struct to store FeatureProvider with EventHandler capability along with their
-// shutdown semaphore
+// Shutdown semaphore
```
hairyhenderson, line 88: I see `shutdown` replaced with `Shutdown` in many comments (here and below) - I'm assuming a find/replace error?

openfeature/event_executor.go
```diff
@@ -110,8 +133,8 @@ func (e *eventExecutor) removeApiHandler(t EventType, c EventCallback) {
 	e.apiRegistry[t] = entrySlice
 }
 
-// registerClientHandler registers a client level handler
-func (e *eventExecutor) registerClientHandler(clientName string, t EventType, c EventCallback) {
+// RegisterClientHandler registers a client level handler
+func (e *EventExecutor) RegisterClientHandler(clientName string, t EventType, c EventCallback) {
```
hairyhenderson, line 137: maybe name this `AddClientHandler` to reflect the relationship to `AddHandler` (i.e. the same except scoped), or perhaps rename `AddHandler` above as `RegisterHandler`?

openfeature/interfaces.go
Kavindu-Dodan: Agree, I will change the naming

openfeature/openfeature_api.go
Kavindu-Dodan: I introduced this to support internal work and avoid direct access to singleton instance from other components like `Client`. For example, a client instance needs to obtain the provider associate with its domain, so it rely on `ForEvaluation` method exposed by this internal interface.

openfeature/event_executor.go
Kavindu-Dodan: yeah thank you

openfeature/interfaces.go
Kavindu-Dodan: I am changing this to `IEvaluation`

openfeature/openfeature_api.go
Kavindu-Dodan: Yup, this can be non-public, will change along with the constructor

openfeature/event_executor.go
Kavindu-Dodan: will rename to `AddClientHandler`. Regarding `AddHandler`, this is already public so I will avoid a breaking change and keep it there.

openfeature/openfeature_api.go
Kavindu-Dodan: This is done and I rechecked the same for other structs that should've been internal

toddbaert (approved):
This makes sense to me, and we've made similar choice regarding the interfaces in other implementations. I'm approving but I think other's with more Go experience than me should also comment (thanks to @hairyhenderson for his feedback so far :pray: ).

beeme1mr (approved):

openfeature/interfaces.go
hairyhenderson: I guess that's slightly better, the `I` prefix is still not very idiomatic though - that feels very "Java" and not at all like a Go interface name.

hairyhenderson:
sorry for the slow response on this, and thanks for merging it!

PR #294, opened 2024-10-22
chore: require go 1.21

toddbaert:
Require go 1.21.

1.20 has long been EOL: https://endoflife.date/go

This is consistent with our guidelines on runtime support.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -6,8 +6,6 @@ on:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
-  # Default minimum version of Go to support.
-  DEFAULT_GO_VERSION: "~1.21"
 
 jobs:
   release-please:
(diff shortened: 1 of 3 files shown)


liran2000 (approved):

.github/workflows/release-please.yml
```diff
@@ -7,7 +7,7 @@ env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
   # Default minimum version of Go to support.
-  DEFAULT_GO_VERSION: "~1.20"
+  DEFAULT_GO_VERSION: "~1.21"
```
beeme1mr, line 10: It might be better to use the version defined in the `go.mod` file instead of manually defining them per file.

Here's an example from the codegen repo.

https://github.com/open-feature/codegen/blob/main/.github/workflows/release.yaml#L50

toddbaert: nice, done

lukas-reining (approved):

.github/workflows/release-please.yml
toddbaert: @beeme1mr note that I had to reverse the order of some steps for this, since we need to checkout before we install go, in order to read this file.

beeme1mr (approved):

trungdlp-wolffun:
@toddbaert

**Did you forget to update README.md?**

image


**Also, Go 1.21 is no longer supported.**

image

Issue #293, opened 2024-10-13
[BUG] Race condition when testing with memprovider


natenho:
Hi! I'm not really sure if this can be considered a bug, but someone could be led to test the code this way.

Consider that I'm testing code and I want to cover both scenarios of a feature flag being enabled and disabled. So we don't want to rely on actual providers, just a stub. The memprovider is the way to go:

```
func TestParallelSingletonUsage(t *testing.T) {
	t.Parallel()

	type fields struct {
		givenFlagValue bool
	}

	tests := map[string]struct {
		fields fields
		want   bool
	}{
		"test when flag is true": {
			fields: fields{givenFlagValue: true},
			want:   true,
		},
		"test when flag is false": {
			fields: fields{givenFlagValue: false},
			want:   false,
		},
	}

	for name, tt := range tests {
		tt := tt
		name := name
		t.Run(name, func(t *testing.T) {
			t.Parallel()

			err := openfeature.GetApiInstance().SetProvider( // Here we have a race condition! We need to set the provider in parallel to ensure the test runs under the expected condition
				memprovider.NewInMemoryProvider(
					map[string]memprovider.InMemoryFlag{
						"some_cool_feature": {
							State:          memprovider.Enabled,
							DefaultVariant: "variant_1",
							Variants: map[string]any{
								"variant_1": tt.fields.givenFlagValue, // Make sure the provider always returns the value given by the test case
							},
							ContextEvaluator: nil,
						},
					},
				))

			require.NoError(t, err)

			got := theActualCodeUnderTest()

			require.Equal(t, tt.want, got, "uh oh, value is not as expected")
		})
	}
}

func theActualCodeUnderTest() bool {
	return openfeature.GetApiInstance().
		GetClient().
		Boolean(context.Background(), "some_cool_feature", false, openfeature.EvaluationContext{})
}

```

Running a single test it is not a issue:

```
$ go test -run ^TestParallelSingletonUsage$ local/goff-tests -count=1
ok      local/goff-tests        0.005s
```

Running more parallel tests, you can start to see some unpredictable results:
```
$ go test -run ^TestParallelSingletonUsage$ local/goff-tests -count=20
--- FAIL: TestParallelSingletonUsage (0.00s)
    --- FAIL: TestParallelSingletonUsage/test_when_flag_is_true (0.00s)
        parallel_test.go:57: 
                Error Trace:    /home/renato_lima/go/src/local/goff-tests/parallel_test.go:57
                Error:          Not equal: 
[...]

PR #295, opened 2024-10-22
feat: TestProvider for easy, parallel-safe testing

warber:
## This PR
- implements an idea how to improve possibility to write highly parallel unit tests.

### Related Issues

Fixes: #293


pkg/openfeaturetest/testprovider.go
```diff
@@ -0,0 +1,104 @@
+package openfeaturetest
```
warber, line 1: I guess we need to come up with a better place for this ? 🤔

pkg/openfeaturetest/testprovider.go
```diff
@@ -0,0 +1,104 @@
...
+	goroutineLocalData.Store(fmt.Sprintf("%d_%v", gID, key), value)
+}
+
+func getGoroutineLocal(key interface{}) interface{} {
+	gID := getGoroutineID()
+	value, _ := goroutineLocalData.Load(fmt.Sprintf("%d_%v", gID, key))
+	return value
+}
+
+func getGoroutineID() uint64 {
```
warber, line 97: not perfect but does the trick of getting the current gid from the runtime

pkg/openfeaturetest/testprovider.go
aepfli: the question is, if we want this to be part of the sdk or rather put this as a separate package in the sdk-contrib - like we did in java for the junit extension.

warber: then to me this sounds like contrib would be the better place. @toddbaert ?

pkg/openfeaturetest/testprovider.go
natenho: @warber to be honest it is a very hacky way of dealing with that.

https://go.dev/doc/faq#no_goroutine_id

pkg/openfeaturetest/testprovider.go
toddbaert: Since there's no external deps, I think we should include this in the SDK.

pkg/openfeaturetest/testprovider.go
toddbaert: > @warber to be honest it is a very hacky way of dealing with that. Maybe we can think of a more elegant solution [since the go routine id is intentionally not available](https://go.dev/doc/faq#no_goroutine_id).

I agree it's hacky, however, there are multiple relatively popular packages that use this same mechanism, and we are only using it for testing. I would not advocate for including this in production code, but I think it's acceptable to dodge our singleton issue for tests.

IMO this is a net improvement. I'm open to other solutions but at the end of the day, we need some transparent piece of data (at least an id) to propagate from the test to the production code; I'm not aware of any other alternative.

pkg/openfeaturetest/testprovider.go
```diff
@@ -0,0 +1,108 @@
...
+		providers: &sync.Map{},
+	}
+}
+
+// TestAwareProvider can be used in parallel unit tests. It holds a map of unit test name to `openfeature.FeatureProvider`s.
+// Before executing the test, specify the actual (in memory) provider that's going to  be used for the specific test using the
+// `SetProvider` method.
+type TestAwareProvider struct {
+	openfeature.NoopProvider
+	providers *sync.Map
```
toddbaert, line 26: My only current issue with this approach is cleaning up this map. It would be great if we could have some ergonomic mechanism to REMOVE things from this map, eventually. Perhaps something a user could clean-up with a `defer` statement in whatever scope they need.

toddbaert (commented):
I approve of this approach. [This](https://github.com/open-feature/go-sdk/pull/295/files#r1823045987) is my only blocker.

pkg/openfeaturetest/testprovider.go
natenho: ok, I can agree with that

toddbaert:
@warber Looks good, I have proposed one more additional ergonomic change I'd like your opinion on. Fundamentally it works the exact same way, it just changes the API a bit so that users are thinking in terms of _flags_ for a test scope, not a _provider_ for a test scope: https://github.com/warber/go-sdk/pull/1

toddbaert:
@warber after you pointed out the deprecated packaging, I moved this to `openfeature/testing`: https://github.com/open-feature/go-sdk/pull/295/commits/9c5de095715228595982b9a74776a25aa5bb668d

This is consistent with goes `testing` package naming, I think.

openfeature/testing/testprovider.go
```diff
@@ -0,0 +1,122 @@
...
+	providers *sync.Map
+}
+
+// UsingFlags sets flags for the scope of a test
+func (tp TestProvider) UsingFlags(test *testing.T, flags map[string]memprovider.InMemoryFlag) {
+	storeGoroutineLocal(test.Name())
+	tp.providers.Store(test.Name(), memprovider.NewInMemoryProvider(flags))
+}
+
+// Cleanup deletes the test provider bound to the current test and should be executed after each test execution
```
toddbaert, line 37:
```suggestion
// Cleanup deletes the flags provider bound to the current test and should be executed after each test execution
```

README.md
```diff
@@ -432,29 +432,71 @@ func (h MyHook) Error(context context.Context, hookContext openfeature.HookConte
...
 
-While you may use global methods to interact with the API, it is recommended to obtain the singleton API instance so that you can use appropriate mocks for your testing needs,
+Call `testProvider.UsingFlags(t, tt.flags)` to set flags for a test, and clean them up with `testProvider.Cleanup()`
 
 ```go
-// global helper
-openfeature.SetProvider(myProvider)
+import (
+ 	"github.com/open-feature/go-sdk/openfeature"
+  "github.com/open-feature/go-sdk/openfeature/testing"
```
beeme1mr, line 443:
```suggestion
  "github.com/open-feature/go-sdk/openfeature"
  "github.com/open-feature/go-sdk/openfeature/testing"
```

beeme1mr (approved):

PR #299, opened 2024-10-31
feat: make provider interface "stateless"

warber:
## This PR

TBD

### Related Issues

Fixes #1234523

### DIFF
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -41,6 +41,7 @@ type Client struct {
 	metadata          ClientMetadata
 	hooks             []Hook
 	evaluationContext EvaluationContext
+	name              string
 
 	mx sync.RWMutex
 }
@@ -56,12 +57,17 @@ func NewClient(name string) *Client {
 
 func newClient(name string, apiRef evaluationImpl, eventRef clientEvent) *Client {
 	return &Client{
+		name:              name,
 		api:               apiRef,
 		clientEventing:    eventRef,
 		metadata:          ClientMetadata{name: name},
 		hooks:             []Hook{},
 		evaluationContext: EvaluationContext{},
 	}
+}
+
+func (c *Client) State() State {
+	return c.clientEventing.State(c.name)
 }
 
 // Deprecated
--- a/openfeature/client_test.go
+++ b/openfeature/client_test.go
@@ -2,6 +2,7 @@ package openfeature
 
 import (
 	"context"
+	"errors"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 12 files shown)


openfeature/client.go
```diff
@@ -64,6 +66,10 @@ func newClient(name string, apiRef evaluationImpl, eventRef clientEvent) *Client
 	}
 }
 
+func (c *Client) State() State {
```
toddbaert, line 69:
```suggestion
// Returns the state of the associated provider.
func (c *Client) State() State {
```

openfeature/util_test.go
```diff
@@ -62,3 +62,18 @@ func (s ProviderEventing) Invoke(e Event) {
 func (s ProviderEventing) EventChannel() <-chan Event {
 	return s.c
 }
+
+func eventually(t *testing.T, condition func() bool, timeout, interval time.Duration, errMsg string) {
```
toddbaert, line 66: I saw you using this and I thought it was some library I didn't know about. We've built the same helper in other repos as well. :joy:

openfeature/event_executor_test.go
```diff
@@ -975,7 +988,7 @@ func TestEventHandler_HandlersRunImmediately(t *testing.T) {
 			rsp <- e
 		}
 
-		AddHandler(ProviderReady, &callback)
+		//AddHandler(ProviderReady, &callback) // TODO: hard to test, because callback is executed on handler registration
```
toddbaert, line 991: Is this TODO still valid?

openfeature/event_executor.go
```diff
@@ -215,6 +213,14 @@ func (e *eventExecutor) emitOnRegistration(
 	}
 }
 
+func (e *eventExecutor) State(domain string) State {
```
toddbaert, line 216: Question: What is the convention for casing of internal interfaces? This implements `clientEvent`, which is internal, but it's methods are capitalized. Is that convention? Since it's internal they could all be lower case. :thinking:

openfeature/client.go
```diff
@@ -41,6 +41,7 @@ type Client struct {
 	metadata          ClientMetadata
 	hooks             []Hook
 	evaluationContext EvaluationContext
+	name              string
```
toddbaert, line 44:
```suggestion
	domain              string
```

`name` has been renamed to `domain` in the most recent spec. I think there's a few places we need to update param names, etc, but if we're adding new fields, prefer `domain`.

openfeature/util_test.go
warber: 😄 yeah, actually i am used to [testify](https://github.com/stretchr/testify) which provides such functionalities.
This repository is not using any external testing library (which is also fine) so I just "borrowed" the idea and naming from there 😛

openfeature/client_test.go
```diff
@@ -1112,3 +1113,195 @@ func TestFlagMetadataAccessors(t *testing.T) {
...
+		NoopProvider{},
+		&ProviderEventing{
+			c: make(chan Event, 1),
+		},
+	}
+
+	if err := SetProviderAndWait(provider); err != nil {
+		t.Fatalf("failed to set up provider: %v", err)
+	}
+	eventually(t, func() bool {
```
toddbaert, line 1156: I think we can get rid of all the `eventually`'s here, since you are using `SetProviderAndWait`.

I also think it might reduce test interference (if you see any) if you pass `t.Name()` into all `GetClient` calls as the domain: `GetApiInstance().GetClient(t.Name())`

openfeature/client_test.go
```diff
@@ -1112,3 +1113,195 @@ func TestFlagMetadataAccessors(t *testing.T) {
...
+		return GetApiInstance().GetClient().State() == FatalState
+	}, time.Second, 100*time.Millisecond, "expected client to report ERROR state")
+}
+
+// The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider
+// is in NOT_READY.
+func TestRequirement_1_7_6(t *testing.T) {
+	defer t.Cleanup(initSingleton)
+	t.Skip("Test not yet implemented")
```
toddbaert, line 1239: Did you mean to implement these? (1.7.6 - 1.7.8)?

It seems to me that the short-circuiting logic is missing for NOT_READY and FATAL, at least. I think you'd need to implement that in `client.evaluate`.

openfeature/util_test.go
toddbaert: I don't mind adding testify at all, since it wouldn't be shipped. If you feel it's a good addition, go for it.

toddbaert:
Great progress! I have 2 concerns: [this](https://github.com/open-feature/go-sdk/pull/299#discussion_r1829554671) and [this](https://github.com/open-feature/go-sdk/pull/299#discussion_r1829566858).

openfeature/client_test.go
warber: yes, let's do that in a another PR

openfeature/client_test.go
warber: I think that's a good idea :) will do. And you are absolutely right, no need for `eventually` in a lot of places 👌

openfeature/client.go
warber: yes, thx i will rename that one. For the other ocurrences. I just saw that the old/wrong terminology is used in quite a lot of places. Wdyt about if I just open another PR that contains all the renaming?

openfeature/event_executor.go
warber: I don’t think there’s a clear convention here. Ultimately, it comes down to the simple question of whether you want others to implement this interface. To me, this (and similar interfaces) seems like internal code, so it makes sense for the methods to be in lowercase—I agree with that.

The issue is that the methods already include exported ones. (Keep in mind that it doesn’t matter if the interface itself is exported; what matters is whether the methods are.) Changing this now would introduce a breaking change.

openfeature/util_test.go
warber: Got it. To keep the test assertions consistent, I wouldn't bring in a new third-party library just for this tho - unless there's a really compelling reason.

openfeature/event_executor_test.go
warber: unfortunately, yes. The RDY callback comes from `emitOnRegistration`. Now that i cannot mock the state via the provider implementation the testing this becomes a bit harder.

openfeature/client.go
toddbaert: You can open a PR or even just an issue. :pray:

openfeature/event_executor.go
toddbaert: That makes sense!

openfeature/client_test.go
toddbaert: I think you can complete these now, yes? Or would you rather leave them to another PR?

toddbaert (approved):
Hey @warber ! This looks great to me now. I will merge this week unless I hear other objections from the TC or maintainers.

cc @kinyoklion @lukas-reining @thomaspoignant @justinabrahms

aepfli (approved):

PR #302, opened 2024-11-09
test: make client test independet from the api

aepfli:
Our client_test is not a unit test; we rely on the API to set the provider regularly, and within the code, we rely on the singleton. With these changes, we mock the API interface and provide it to the client's existing factory method. So each of the tests is independent from the others.

There are still some tests, which i would rather declare as e2e or integration tests, which i did not tackle, or where i do think it makes sense to keep them as is.


openfeature/client_test.go
```diff
@@ -27,13 +27,35 @@ func TestRequirement_1_2_1(t *testing.T) {
...
 }
 
+func newTestClientWithProvider(clientName string, t *testing.T) (*Client, *MockFeatureProvider, *MockevaluationImpl) {
+	client, evalApi := newTestClient(clientName, t)
+
+	mockProvider := NewMockFeatureProvider(evalApi.ctrl)
+	mockProvider.EXPECT().Metadata().AnyTimes()
+	mockProvider.EXPECT().Hooks().AnyTimes()
+
+	evalApi.EXPECT().ForEvaluation(client.metadata.name).Times(1).DoAndReturn(func(_ string) (*MockFeatureProvider, []Hook, EvaluationContext) {
```
aepfli, line 37: Now we're even checking if this call only happens once - i am not sure if i could return without a function, at least i did not make it work

openfeature/client_test.go
```diff
@@ -1058,6 +1018,7 @@ func TestErrorCodeFromProviderReturnedInEvaluationDetails(t *testing.T) {
 }
 
 func TestSwitchingProvidersMidEvaluationCausesNoImpactToEvaluation(t *testing.T) {
+	t.Skip("not sure how to handle this one now")
```
aepfli, line 1021: https://github.com/open-feature/go-sdk/pull/302/files#r1835318170—Based on this, we're sure we're not calling another provider, so I think we can remove this test. Wdyt?

warber:
hey @aepfli, It's excellent that we are no longer relying on the API instance in these tests. However, I have some concerns about the two functions, `newTestClient()` and `newTestClientWithProvider()`. While they provide a convenient one-liner setup for tests, they might raise some questions for users regarding their implementation and usage.

What exactly do these functions return? Is it a mocked client for testing? It seems more like a real client with mocked dependencies. Same for the `WithProvider` part... What are the return values of these functions? If another API needs to be mocked, will these functions return an additional value? In general i consider three return values handed back from a function already more than enough :D ... ...

I think it would be worth thinking about to separate the mocked dependencies from the creation of the client. For example, we could add a helper function in the util_test.go file to create all the mocks that might be useful for the tests.

Something like:
```go
type clientMocks struct {
	clientHandlerAPI *MockclientEvent
	evaluationAPI    *MockevaluationImpl
	providerAPI      *MockFeatureProvider
}

func mocksForClient(t *testing.T) clientMocks {
	ctrl := gomock.NewController(t)
	mockExecutor := NewMockclientEvent(ctrl)
	mockEvalApi := NewMockevaluationImpl(ctrl)
	mockProvider := NewMockFeatureProvider(ctrl)

	mockProvider.EXPECT().Metadata().AnyTimes()
	mockProvider.EXPECT().Hooks().AnyTimes()
	mockExecutor.EXPECT().State(gomock.Any()).AnyTimes().Return(ReadyState)
	mockEvalApi.EXPECT().ForEvaluation(gomock.Any()).Times(1).DoAndReturn(func(_ string) (*MockFeatureProvider, []Hook, EvaluationContext) {
		return mockProvider, nil, EvaluationContext{}
	})

	return clientMocks{
		clientHandlerAPI: mockExecutor,
		evaluationAPI:    mockEvalApi,
		providerAPI:      mockProvider,
	}
}
```



Note that I left the "basic" expectations in there, but I am generally questioning if they should be specified at this level. However, I think it's fine for now.

In the test, you would need to create mocks for your client, either via this helper or separately—the choice is up to the developer. Then, use the real constructor function, passing in the dependencies/mocks.

```go
mocks := mocksForClient(t)
mocks.providerAPI.EXPECT().BooleanEvaluation(gomockstuff...).
	Return(BoolResolutionDetail{
		Value: booleanValue,
		ProviderResolutionDetail: ProviderResolutionDetail{
			Variant: booleanVariant,
			Reason:  testReason,
		},
	})

client := newClient("test-client", mocks.evaluationAPI, mocks.clientHandlerAPI)
...
```

tell me wdyt 👋

aepfli:
@warber, I like the proposed type for "collecting" all the mocks, but I am not sure about splitting it into a different file. 

IMHO, there is currently just one usage within tests: this one. Furthermore, there won't be many other use cases where this might be beneficial, especially as the mocked methods are purely for the client's default behavior to work.
We need those default mocks throughout all tests for basic evaluation functionality, even if they're not always 100% required.

I will take your remarks into consideration and try to incorporate them into the next patch.

aepfli:
this pr is blocked by https://github.com/open-feature/go-sdk/issues/308

aepfli:
@toddbaert @warber as the blocking task got resolved, i would highly appreciate your review

warber (approved):

toddbaert (approved):

toddbaert:
@aepfli I tested locally to verify a few things, and everything looks good. I did change the title to `test:..` since this isn't a user-facing feature.

PR #305, opened 2024-11-25
chore: use correct term "domain" instead of name/clientName in code

warber:
## This PR
This PR ensures consistent usage of the term "domain" throughout the codebase, replacing instances of "name," "clientName," and similar terms.


aepfli (approved):

toddbaert (approved):
This looks right to me - my only concern with any of this is that it might be breaking, but I can't see anything that's exposed publicly which is changing.

lukas-reining (approved):

PR #319, opened 2025-02-04
feat: loosen test instance interface to support various test frameworks #316

### LINKED ISSUE #316: [FEATURE] Support for Ginkgo GinkgoT()
### Requirements

Add support for UsingFlags() variant using the testing.TB interface instead so that it works with for example GinkgoT()

ekarlso:
feat: loosen test instance interface to support various test frameworks (#317)

## This PR

Use interface instead of testing.BB since it has private() that blocks use.

### Related Issues

Fixes #316


openfeature/testing/testprovider.go
```diff
@@ -29,7 +28,7 @@ type TestProvider struct {
 }
 
 // UsingFlags sets flags for the scope of a test
-func (tp TestProvider) UsingFlags(test *testing.T, flags map[string]memprovider.InMemoryFlag) {
+func (tp TestProvider) UsingFlags(test interface{ Name() string }, flags map[string]memprovider.InMemoryFlag) {
```
thomaspoignant, line 31:
```suggestion
type TestFramework = interface{ Name() string }

func (tp TestProvider) UsingFlags(test TestFramework }, flags map[string]memprovider.InMemoryFlag) {
```

thomaspoignant (approved):
This looks good to me, my only small concern is that it is not obvious anymore what is the type of `test`, this is why I think we can name this type as proposed in the comment.

aepfli (approved):

PR #320, opened 2025-02-13
refactor: remove dependency on golang.org/x/exp

treuherz:
## This PR

The exp module does not have the same compatibility guarantees as the standard library. The fact that it stays on a pseudo-version instead of a proper semver tag, and the possibility for backwards-incompatible changes, makes depending on exp in a widely-imported library a bit risky. If a dependent of the SDK updated the version of exp they relied on, the MVS algorithm could select a version that broke the SDK's usage of exp's packages.

Out of an abundance of caution, I'm trying to reduce the use of exp in our private shared libraries for this reason, and the OpenFeature SDK is our only external dependency that depends on it it. The uses are pretty small. I've replaced them with existing standard library functiosn where possible, or in one place implemented the functionality myself, since `maps.Values` is only available in the standard library from go 1.23. 

While in the area, I've also updated a couple of functions that are already replaceable with functions from the standard slices package.

### How to test

Unit and E2E tests all pass localy.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ require (
 	github.com/cucumber/godog v0.15.0
 	github.com/go-logr/logr v1.4.2
 	github.com/golang/mock v1.6.0
-	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842
 	golang.org/x/text v0.22.0
 )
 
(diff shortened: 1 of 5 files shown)


beeme1mr (approved):

beeme1mr:
Greets @treuherz, could you please update this branch? Your fork must not allow me to update it myself. Please @ mention me once it's up to date so I can merge the change. Thanks again for the PR.

treuherz:
@beeme1mr Rebased!

PR #326, opened 2025-02-20
ci: updated goreleaser-action to v6 with updates to use goreleaser v2

### LINKED ISSUE #324: Upgrade GO Releaser to v6
The upgrade guide can be found here.

https://goreleaser.com/blog/goreleaser-v2/#upgrading

bbland1:
## This PR

- updates `goreleaser-action` from `v5` to `v6`
- `v6` utilizes `v2` of `goreleaser` updated the parameters of the action to specify the `distribution` and `version`
- update the `args` to use the new argument of `--clean` instead of `--rm-rf`

### Related Issues

Fixes #324

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -43,8 +43,10 @@ jobs:
           version: v1.3.0
 
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v5
+        uses: goreleaser/goreleaser-action@v6
         with:
-          args: release --rm-dist
+          distribution: goreleaser
+          version: "~> v2"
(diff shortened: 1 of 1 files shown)


beeme1mr (approved):
Thanks! It looks good from what I can tell. It may be possible to test from a fork but it's probably more trouble than it's worth. I'll just add a note to the release PR that this change was made in case publishing fails.

bbland1:
@beeme1mr sweet! Yeah I did a bit of research and it seems like it's maybe possible but it's working with like a mocking of the publishing so could have still had issues with the next release, so it seemed better to get the PR looked at!

PR #337, opened 2025-04-02
fix: add logger to client

sivchari:
## This PR

Added logger field to client to output to somewhere in the future.

### DIFF
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -42,6 +42,7 @@ type Client struct {
 	hooks             []Hook
 	evaluationContext EvaluationContext
 	domain            string
+	logger            logr.Logger
 
 	mx sync.RWMutex
 }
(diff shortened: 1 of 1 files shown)


beeme1mr (approved):

aepfli (approved):
thank you, this seems to be a valuable addition - and I am curious, why we haven't thought/discovered this earlier

openfeature/client.go
```diff
@@ -76,6 +78,7 @@ func (c *Client) State() State {
 func (c *Client) WithLogger(l logr.Logger) *Client {
 	c.mx.Lock()
 	defer c.mx.Unlock()
+	c.logger = l
```
sahidvelji, line 81: This method is deprecated according to the doc comment.

liran2000 (changes requested):
hi,
logger was on the client earlier, and was intentionally removed at #289 by the requirements as described there.

toddbaert:
I think we should close this, we don't generally want to log in "hot paths" like evaluations. I'm fine with adding logging for specific, infrequent things, but I don't think this is necessary as is.

See: https://openfeature.dev/specification/sections/flag-evaluation/#requirement-1411

PR #341, opened 2025-05-07
feat: Make Mocks Exportable

jblacker:
## This PR

- Updated the abandoned version of gomock & mockgen for the UberMock version.
- Makes generated mocks exportable to be usable by other packages or external users
- Sets a new build tag `testtools` on the generated mock files & updates the Makefile to use this build tag when running tests or linting
- Updated documentation related to using these mocks
- Added `codecov.yml` file to prevent mocks from being included in coverage reports

### Notes

I started working on migrating my multiprovider work from `go-sdk-contrib` to the main SDK here. The multiprovider makes sense to be in it's own package within the SDK considering it's part of the appendix spec similar to the `InMemoryProvider`. To fully unit test this component I would need to leverage the already generated mocks so making them exportable and usable outside the `openfeature` package. This also exposes them outside of the library so I've added the build tags and documentation about their usage. This PR is essentially a prereq task to making sure the `MultiProvider` has regression & unit tests.


aepfli (approved):
thank you, this is great

jblacker:
I wanted to put the mocks into their own package, but unfortunately I was not able to resolve the import cycles without breaking changes.

beeme1mr (approved):

thomaspoignant:
@jblacker why do you need to add a tag for the build? I am not sure to understand the benefits of having it?
Would it be better to add the mock in a separated project that we can import in the go.mod?

jblacker:
@thomaspoignant It's not _needed_, but imho it's a nice to have as it prevents the mocks from being compiled by default when building the application since they are no longer "test files".

Personally, I'm not a fan of putting mocks into a separate project; especially for a library like this.

thomaspoignant:
Ok I get it, sounds good to me.

thomaspoignant (approved):

PR #351, opened 2025-05-15
chore: require go 1.23.0

### LINKED ISSUE #342: Makefile lint target fails
The lint target of the Makefile currently fails with an error. The version of golangci-lint is outdated and the --deadline flag was renamed to --timeout.

https://github.com/golangci/golangci-lint/pull/793

### LINKED ISSUE #349: Update go directive to 1.23.0
The README states that "The OpenFeature Go SDK only supports currently maintained Go language versions". As such, I suggest that this be made explicit by updating the go directive in go.mod to be 1.23.0. This also aligns with how https://github.com/open-telemetry/opentelemetry-go manages Go versions.

https://github.com/open-feature/go-sdk/blob/640448d9e1e1c76bc9e7fe7c7b7dbce2298493cb/go.mod#L3

sahidvelji:
## This PR

Update the go directive to 1.23.0. This causes linter failures, meaning the version of golangci-lint also has to be upgraded at the same time. Doing so surfaces several linter findings, which are also fixed in this PR.

### Related Issues

Fixes #349 
fixes #342


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/open-feature/go-sdk
 
-go 1.21
```
beeme1mr, line 3: I'm reluctant to switch to Go `1.23` since  `1.22` is still supported.

FYI @toddbaert @thomaspoignant @lukas-reining

sahidvelji: It's not supported actually. Go 1.24 and 1.23 are the currently supported versions of Go. Go 1.22 stopped being supported as soon as 1.24.0 was released. This is standard.
https://go.dev/doc/devel/release

beeme1mr: I misread the policy, sorry. I thought it was the current release plus the two prior. For what's worth, the [OTel SDK only supports Go 1.23 and 1.24](https://github.com/open-telemetry/opentelemetry-go#compatibility).

I'll defer to the @open-feature/technical-steering-committee on this since changing major runtime versions can have a significant impact on consumers.

sahidvelji: No worries. Issue #349 has some additional details.

aepfli (approved):
Thank you for this, and thank you for the nice explanation of when GoLang versions are unsupported. I am in favor of moving forward. Thank you

lukas-reining (approved):
To me, changing the version to 1.23 fit's our version policy and as you said @beeme1mr, is the same with OTEL so it looks good to me :)

PR #354, opened 2025-05-15
Implement multiprovider

jblacker:
## This PR
Implements the Multiprovider specified in "Appendix A: Included Utilities" of the OpenFeature Spec. This is implementation is an improvement upon what I initially implemented in the `go-sdk-contrib` repository [here](https://github.com/open-feature/go-sdk-contrib/pull/669) & [here](https://github.com/open-feature/go-sdk-contrib/pull/678).

This includes implementations for all three strategies described in the spec along with some common functions. Unlike my original implementation this includes full support for hooks & event handling from the included handlers.

### Related Issues

https://github.com/open-feature/go-sdk-contrib/issues/599

### Notes
I reimplemented this here from the `go-sdk-contrib` repo on suggestion of other members. This includes the refactoring improvements that I was mentioning in discussions on my open PR over there.

This PR is currently in draft because it is built on top of the exported mocks in order to provide a full unit test suite.

### Follow-up Tasks

- e2e / Gherkin tests should be written
- 
### How to test

- Create an application and pass a multiprovider to the OpenFeature SDK that includes 2 or more providers. Verify that the results are as expected per the selected strategy.


jblacker:
I've actually been on vacation. I'm going to pick back up on this tomorrow.

jblacker:
This is ready for review, but codecov is blocking CI. The issue is that only the mocks in the `openfeature` package are ignored, so the newly added mocks in the `multiprovider` package is causing the % to decrease more than is accurate hitting the fail threshold.

go.mod
```diff
@@ -5,16 +5,21 @@ go 1.23.0
 require (
 	github.com/cucumber/godog v0.15.0
 	github.com/go-logr/logr v1.4.3
+	github.com/stretchr/testify v1.9.0
```
sahidvelji: This PR does introduce testify as a dependency. I don't have a strong opinion on this but I want to highlight that for other reviewers.

openfeature/multiprovider/isolation.go
```diff
@@ -0,0 +1,340 @@
...
+		evalCtx,
+	)
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.capturedContext = hookCtx
+}
+
+// deepenContext Convert a [of.FlattenedContext] to a [of.EvaluationContext]
+func deepenContext(flatCtx of.FlattenedContext) of.EvaluationContext {
+	noTargetingKey := make(map[string]interface{})
```
sahidvelji: Prefer `any` over `interface{}`. There are a couple of other instances that should be updated too.

openfeature/multi/isolation.go
```diff
@@ -0,0 +1,340 @@
...
+	"context"
+	"fmt"
+	of "github.com/open-feature/go-sdk/openfeature"
+	"sync"
+)
+
+type (
+	// HookIsolator used as a wrapper around a provider that prevents context changes from leaking between providers
+	// during evaluation
+	HookIsolator struct {
```
sahidvelji: You can embed the `openfeature.UnimplementedHook` in this struct so that you don't have to define empty `After`, `Error`, and `Finally` methods.

openfeature/multiprovider/isolation.go
jblacker: Updated throughout.

openfeature/multi/isolation.go
jblacker: Good call, this is much cleaner

openfeature/multiprovider/errors.go
```diff
@@ -0,0 +1,55 @@
...
+	_ error = (AggregateError)(nil)
+)
+
+func (e *ProviderError) Error() string {
+	return fmt.Sprintf("Provider %s: %s", e.ProviderName, e.Err.Error())
+}
+
+// NewAggregateError Creates a new AggregateError
+func NewAggregateError(providerErrors []ProviderError) AggregateError {
+	err := make(AggregateError)
```
sahidvelji:
```suggestion
	err := make(AggregateError, len(providerErrors))
```

openfeature/multiprovider/errors.go
```diff
@@ -0,0 +1,55 @@
+package multiprovider
+
+import (
+	"errors"
+	"fmt"
+)
+
+type (
+	// ProviderError is an error wrapper that species the provider name
```
sahidvelji:
```suggestion
	// ProviderError is an error wrapper that species the provider name.
```
Doc comments should be full sentences.

sahidvelji: Please update other instances too.

openfeature/multiprovider/errors.go
```diff
@@ -0,0 +1,55 @@
...
+	ProviderError struct {
+		Err          error
+		ProviderName string
+	}
+
+	// AggregateError map that contains up to one error per provider within the multi-provider
```
sahidvelji:
```suggestion
	// AggregateError is a map that contains up to one error per provider within the multi-provider.
```

openfeature/multiprovider/errors.go
```diff
@@ -0,0 +1,55 @@
...
+	default:
+		errs := make([]error, 0, size)
+		for _, err := range ae {
+			errs = append(errs, &err)
+		}
+		return errors.Join(errs...).Error()
+	}
+
+	return "" // This will never occur, switch is exhaustive
+}
```
sahidvelji:
```suggestion
	}

	errs := make([]error, 0, size)
	for _, err := range ae {
		errs = append(errs, &err)
	}
	return errors.Join(errs...).Error()
}
```
You can move the default case out of the switch so that you don't have an extra return statement.

openfeature/multiprovider/strategies.go
```diff
@@ -0,0 +1,147 @@
...
+				merged[key] = value
+			}
+		}
+		return merged
+	}
+}
+
+// BuildDefaultResult when executing a strategy that results in a failure state this can be called to return a
+// default result with an internal error
+func BuildDefaultResult[R any](strategy EvaluationStrategy, defaultValue R, err error) of.InterfaceResolutionDetail {
```
sahidvelji: Should this be exported?

openfeature/multiprovider/first_success_strategy.go
```diff
@@ -0,0 +1,134 @@
...
+			if notFoundCount == len(providers) {
+				return BuildDefaultResult[R](StrategyFirstSuccess, defaultVal, nil)
+			}
+		case <-ctx.Done():
+			var err error
+			if len(errs) > 0 {
+				err = NewAggregateError(errs)
+			} else {
+				err = ctx.Err()
+			}
```
sahidvelji:
```suggestion
			err := ctx.Err()
			if len(errs) > 0 {
				err = NewAggregateError(errs)
			}
```

openfeature/multiprovider/isolation.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+)
+
+var (
+	_ of.FeatureProvider = (*HookIsolator)(nil)
+	_ of.Hook            = (*HookIsolator)(nil)
+	_ of.EventHandler    = (*EventHandlingHookIsolator)(nil)
+)
+
+// IsolateProvider Wraps a [of.FeatureProvider] to execute its hooks along with any additional ones
```
sahidvelji:
```suggestion
// IsolateProvider wraps a [of.FeatureProvider] to execute its hooks along with any additional ones.
```

openfeature/multiprovider/strategies.go
jblacker: Yes, this is exported on purpose for those that want to implement a custom strategy.

openfeature/multiprovider/errors.go
jblacker: I don't necessarily agree that all in-code documentation needs to be in prose with full punctuation, but I've updated everything regardless. Some of them use the method name prefix as the start of the sentence, which sounds right if you include the prefix, but might sound weird otherwise.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,583 @@
...
+	Configuration struct {
+		useFallback               bool
+		fallbackProvider          of.FeatureProvider
+		customStrategy            Strategy
+		logger                    *slog.Logger
+		timeout                   time.Duration
+		hooks                     []of.Hook
+		providerHooks             map[string][]of.Hook
+		customComparator          Comparator
+		alwaysUseCustomComparator bool
```
erka: How is this used? I see that it can be set using WithCustomComparatorFunc, but I don't see any actual usage of it in the codebase.

```sh
openfeature/multiprovider/multiprovider.go
70:             alwaysUseCustomComparator bool
179:            conf.alwaysUseCustomComparator = alwaysUse
```

openfeature/multiprovider/first_success_strategy.go
```diff
@@ -0,0 +1,134 @@
...
+	metadata := make(of.FlagMetadata)
+	metadata[MetadataStrategyUsed] = StrategyFirstSuccess
+	errChan := make(chan ProviderError, len(providers))
+	notFoundChan := make(chan any)
+
+	type namedResolution struct {
+		of.InterfaceResolutionDetail
+		name string
+	}
+	finishChan := make(chan *namedResolution, len(providers))
```
erka: Multiple channels are being opened without being closed, which may result in a memory leak or resource exhaustion.

erka:
@jblacker I spent some time exploring an alternative way to implement this provider. I believe all strategies could be simplified and implemented as plain functions. I also looked into `java-sdk-contrib` and `js-sdk-contrib`,  and they follow a similar approach.

Here's how the Java multi-provider handles it: https://github.com/open-feature/java-sdk-contrib/blob/0eea5daa0115aecd66d5c2b946f48504e282afc5/providers/multiprovider/src/main/java/dev/openfeature/contrib/providers/multiprovider/Strategy.java#L10-L17. 

I pushed my research into https://github.com/erka/go-sdk/commits/rd/implement-multiprovider/. Check it out if you're curious about another way to do this. I acknowledge that it involves several trade-offs.

It would be great to double-check the implementations of `FirstMatchStrategy` and `FirstSuccessStrategy` to ensure they align with other SDKs. It looks like there are some differences in `FirstSuccessStrategy`. 

This one's a bit complex — it'll probably need some extra time before it could be merged.

openfeature/multiprovider/errors.go
sahidvelji: I should have been clearer here. What I meant to say is that all exported identifiers and packages should have doc comments that are full sentences. The reason is that those doc comments form the documentation for the SDK.
https://pkg.go.dev/github.com/open-feature/go-sdk@v1.15.0/openfeature

openfeature/multiprovider/multiprovider.go
jblacker: This is only used with the `ComparisonStrategy` and ignored otherwise. It's used for `ObjectEvaluation` for objects that are not `comparable`. If that is true it will use the custom comparator even if they are `comparable`.

openfeature/multiprovider/first_success_strategy.go
jblacker: From my understanding this should not cause a leak and will be garbage collected when falling out of scope because calling `close` is just a control function rather than a cleanup mechanism. From this understanding the only times a channel needs to be closed is if iterating on it or when waiting on the channel forever to avoid deadlock. However, I'm happy to be proven wrong if you can share some docs.

jblacker:
@erka Alright, just went through your refactoring of the strategies.

However, first I reread the spec as well as the JS reference implementation. You're right that I incorrectly implemented the `FirstSuccessStrategy` as a parallel strategy, when it should have been sequential. I think I misinterpreted this line of the spec:
>...except that errors from evaluated providers do not halt execution. Instead, it will return the first successful result from a provider. [§Appendix A#First Success Strategy](https://openfeature.dev/specification/appendix-a/#first-successful-strategy) 

I took it to mean run in parallel and return the first successful result, rather than sequential like First Match Strategy _**other than**_ not raising  the non-`FLAG_NOT_FOUND` errors from providers.

Back to the refactoring efforts: I really like the changes you've made here to simplify this and would be happy to pull them into this PR. I really how you've leveraged generics in a much bigger way than I initially put together to remove a ton of duplicated efforts & the function rather than interface eliminates a lot of boilerplate. I have a few general questions and about your changes:

1. We should definitely re-implement `FirstSuccessStrategy` to match the other implementations & the spec (as you already did), but should we also include the parallel version I accidentally implemented as a "bonus strategy" and rename it to something like `FirstSuccessParallelStrategy`? I'm not sure if worthwhile considering we'd have to maintain it nor if there's any demand for it.
2. Right now the timeout only applies to Comparison Strategy, but should we also implement it in the other strategies? If yes, we'd need to decide what to do at expiration: do we move onto the next provider or return the default value? Do we have it apply to the total process rather than per provider call?
3. One thing I wasn't sure about is what can be returned by `ObjectEvaluation`? Can it ever be a pointer-based type like a collection or actual pointer to a struct? I guess that would depend on the provider, but it was unclear per the spec. I see that yours assumes it is impossible, which makes sense for the most part.

openfeature/multiprovider/first_success_strategy.go
erka: As it should be re-implemented, this probably isn't relevant at the moment

erka:
@jblacker I've opened the PR into your repo.

> 1. We should definitely re-implement `FirstSuccessStrategy` to match the other implementations & the spec (as you already did), but should we also include the parallel version I accidentally implemented as a "bonus strategy" and rename it to something like `FirstSuccessParallelStrategy`? I'm not sure if worthwhile considering we'd have to maintain it nor if there's any demand for it.

I prefer to keep PR as small as possible, but I’m open to whatever the team decides about bonus strategy. `FirstSuccessStrategy` should be reworked. The JS SDK includes a few additional features, such as the ability to run providers either in parallel or in sequence. We could focus on that in future PR.

> 2. Right now the timeout only applies to Comparison Strategy, but should we also implement it in the other strategies? If yes, we'd need to decide what to do at expiration: do we move onto the next provider or return the default value? Do we have it apply to the total process rather than per provider call?

This is a good question.  I see that providers have their own timeouts with could be configured with options or with custom http client. I haven't checked other sdk if they have such option. Some homework to do here.

> 3. One thing I wasn't sure about is what can be returned by `ObjectEvaluation`? Can it ever be a pointer-based type like a collection or actual pointer to a struct? I guess that would depend on the provider, but it was unclear per the spec. I see that yours assumes it is impossible, which makes sense for the most part.

I might be missing something here — could you walk me through the issue in a bit more detail?

jblacker:
>One thing I wasn't sure about is what can be returned by ObjectEvaluation? Can it ever be a pointer-based type like a collection or actual pointer to a struct? I guess that would depend on the provider, but it was unclear per the spec. I see that yours assumes it is impossible, which makes sense for the most part.
>>I might be missing something here — could you walk me through the issue in a bit more detail?

What I wasn't sure about is all the possibilities of what could be returned when performing an object evaluation. If the type returned is not `Comparable` then a custom comparison method would be required.

jblacker:
@erka I may have also merged too quickly. I didn't notice that there were two commits of yours that were merged in that did not have DCO signoff.

toddbaert:
I can fix the signoff stuff with a squash. I will review this tomorrow.

toddbaert:
> I can fix the signoff stuff with a squash. I will review this tomorrow.

Sorry; I really intended to review this today, but did not have time! It's on my list, I'll try to get to it on Monday!

go.mod
toddbaert: As long as it's only used in test code it's fine with me.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,580 @@
...
+			// Update provider status to error, no event needs to be emitted yet
+			mp.updateProviderState(pErr.ProviderName, of.ErrorState)
+		} else {
+			pErr = &ProviderError{
+				Err:          err,
+				ProviderName: "unknown",
+			}
+			mp.setStatus(of.ErrorState)
+		}
+		mp.outboundEvents <- of.Event{
```
toddbaert: I don't think we need to do this anymore.

The SDK now automatically emits ready/error depending on the return values of the provider `Init`. See: https://github.com/open-feature/go-sdk/blob/5b0218b496714834b0c0cefe720b1392cff40ca7/openfeature/openfeature_api.go#L252

I don't blame you for this - this was changed a few months back; before then, these events were needed. Now, only "change events" or other events "spontaneously" occurring from the provider should be directly emitted by it; nothing about the actual init/shutdown lifecycle requires the provider to emit it's own events directly.

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,580 @@
...
+	for h := range handlers {
+		go mp.startListening(workerCtx, h.name, h.EventHandler, &mp.workerGroup)
+	}
+	mp.shutdownFunc = shutdownFunc
+
+	mp.workerGroup.Add(1)
+	go func() {
+		workerLogger := mp.logger.With(slog.String("multiprovider-worker", "event-forwarder-worker"))
+		defer mp.workerGroup.Done()
+		for e := range mp.inboundEvents {
```
toddbaert: These events we DO need.

openfeature/multiprovider/multiprovider.go
jblacker: Wonderful! Deleting code is one of my favorite pastimes.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,580 @@
...
+// OpenFeature spec. This method should only be called if the provider state mutex is locked
+func (mp *MultiProvider) evaluateState() of.State {
+	maxState := stateValues[of.ReadyState] // initialize to the lowest state value
+	for _, s := range mp.providerStatus {
+		if stateValues[s] > maxState {
+			// change in state due to higher priority
+			maxState = stateValues[s]
+		}
+	}
+	return stateTable[maxState]
```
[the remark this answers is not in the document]
jblacker: Gemini was _almost_ correct. The logic here is correct, but I had the values in the stateValues map flipped. Corrected this & added tests for regression.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,580 @@
...
+func (mp *MultiProvider) Init(evalCtx of.EvaluationContext) error {
+	var eg errgroup.Group
+	// wrapper type used only for initialization of event listener workers
+	type namedEventHandler struct {
+		of.EventHandler
+		name string
+	}
+	mp.logger.LogAttrs(context.Background(), slog.LevelDebug, "start initialization")
+	mp.inboundEvents = make(chan namedEvent, len(mp.providers))
+	handlers := make(chan namedEventHandler)
```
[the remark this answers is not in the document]
jblacker: Interesting that I haven't run into this yet, but Gemini is correct here.

openfeature/multiprovider/discard_handler.go
sahidvelji, line 1: This can be removed now that Go 1.23.0 is EOL

openfeature/multiprovider/isolation.go
```diff
@@ -0,0 +1,333 @@
...
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.capturedContext = hookCtx
+}
+
+// deepenContext Converts a [of.FlattenedContext] to a [of.EvaluationContext].
+func deepenContext(flatCtx of.FlattenedContext) of.EvaluationContext {
+	noTargetingKey := make(map[string]any)
+	for k, v := range flatCtx {
+		if k != "targetingKey" {
```
sahidvelji: Use the const `of.TargetingKey`. Update other instances too.

openfeature/multiprovider/isolation.go
```diff
@@ -0,0 +1,329 @@
...
+			noTargetingKey[k] = v
+		}
+	}
+	return of.NewEvaluationContext(flatCtx["targetingKey"].(string), noTargetingKey)
+}
+
+// flattenContext Convert a [of.EvaluationContext] to a [of.FlattenedContext]
+func flattenContext(evalCtx of.EvaluationContext) of.FlattenedContext {
+	flatCtx := evalCtx.Attributes()
+	flatCtx["targetingKey"] = evalCtx.TargetingKey()
```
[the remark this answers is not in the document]
jblacker: Unnecessary because that's what `evalCtx.Attributes()` does.

openfeature/multi/isolation.go
```diff
@@ -0,0 +1,329 @@
...
+	for i := 1; i < len(evaluationContexts); i++ {
+		if targetingKey == "" && evaluationContexts[i].TargetingKey() != "" {
+			targetingKey = evaluationContexts[i].TargetingKey()
+		}
+
+		for k, v := range evaluationContexts[i].Attributes() {
+			_, ok := attributes[k]
+			if !ok {
+				attributes[k] = v
+			}
```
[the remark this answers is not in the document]
jblacker: Gemini is incorrect here since a new map is created by calling `.Attributes()`

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,580 @@
...
+		logger:        slog.New(discardLogHandler),
+		providerHooks: make(map[string][]of.Hook),
+		timeout:       5 * time.Second, // Default timeout
+	}
+
+	for _, opt := range options {
+		opt(config)
+	}
+
+	providers := providerMap
```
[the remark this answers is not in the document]
jblacker: Not a concern here unless there's a very odd use-case for using that map outside of configuration of OpenFeature

openfeature/multiprovider/discard_handler.go
jblacker: Done

openfeature/multiprovider/isolation.go
jblacker: Done

openfeature/multiprovider/comparison_strategy.go
```diff
@@ -0,0 +1,239 @@
...
+
+	of "github.com/open-feature/go-sdk/openfeature"
+	"golang.org/x/sync/errgroup"
+)
+
+type Comparator func(values []any) bool
+
+// NewComparisonStrategy creates a new instance of ComparisonStrategy. The fallback provider specified is called when
+// there is a comparison failure -- prior to returning a default result. The Comparator parameter is optional and nil
+// can be passed as long as ObjectEvaluation is never called. Unless the `alwaysUseCustom` parameter is true the default
```
sahidvelji: I don't see an `alwaysUseCustom` parameter here. Go doc comments also don't do anything special with backticks, so you should avoid them.

openfeature/multiprovider/comparison_strategy.go
```diff
@@ -0,0 +1,239 @@
...
+// comparisons built into Go will be used. The custom Comparator will only be used for ObjectEvaluation. However, if the
+// parameter is set to true the custom Comparator will always be used regardless of evaluation type. If ObjectEvaluation
```
sahidvelji: Please use doc links to refer to any exported identifier, like [Comparator] and [of.FeatureProvider.ObjectEvaluation]. See https://tip.golang.org/doc/comment#doclinks

openfeature/multiprovider/strategies.go
```diff
@@ -0,0 +1,154 @@
...
+	of "github.com/open-feature/go-sdk/openfeature"
+)
+
+const (
+	// Strategies
+
+	// StrategyFirstMatch First provider whose response that is not FlagNotFound will be returned. This is executed
+	// sequentially, and not in parallel.
+	StrategyFirstMatch EvaluationStrategy = "strategy-first-match"
+	// StrategyFirstSuccess First provider response that is not an error will be returned. This is executed sequentially
```
sahidvelji:
```suggestion
	// StrategyFirstSuccess returns the first provider response that is not an error. This is executed sequentially.
```

Write the other doc comments in a similar fashion. They should read like sentences and the first word after the identifier being documented should not be capitalized.

openfeature/multi/strategies.go
```diff
@@ -0,0 +1,154 @@
...
+	// StrategyComparison All providers are called in parallel. If all responses agree the value will be returned.
+	// Otherwise, the value from the designated fallback provider's response will be returned. The fallback provider
+	// will be assigned to the first provider registered.
+	StrategyComparison EvaluationStrategy = "strategy-comparison"
+	// StrategyCustom allows for using a custom Strategy implementation. If this is set you MUST use the WithCustomStrategy
+	// option to set it
+	StrategyCustom EvaluationStrategy = "strategy-custom"
+
+	ReasonAggregated         of.Reason = "AGGREGATED"
+	ReasonAggregatedFallback of.Reason = "AGGREGATED_FALLBACK"
```
sahidvelji: This should be in its own const () block since it's not an [EvaluationStrategy].

openfeature/multiprovider/strategies.go
```diff
@@ -0,0 +1,154 @@
...
+	// will be assigned to the first provider registered.
+	StrategyComparison EvaluationStrategy = "strategy-comparison"
+	// StrategyCustom allows for using a custom Strategy implementation. If this is set you MUST use the WithCustomStrategy
+	// option to set it
+	StrategyCustom EvaluationStrategy = "strategy-custom"
+
+	ReasonAggregated         of.Reason = "AGGREGATED"
+	ReasonAggregatedFallback of.Reason = "AGGREGATED_FALLBACK"
+
+	ErrAggregationNotAllowedText = "object evaluation not allowed for non-comparable types without custom comparable func"
```
sahidvelji: This should be in its own `const ()` block since it's not an [EvaluationStrategy].

openfeature/multiprovider/strategies.go
```diff
@@ -0,0 +1,154 @@
...
+
+type (
+	// EvaluationStrategy Defines a strategy to use for resolving the result from multiple providers.
+	EvaluationStrategy = string
+
+	// FlagTypes defines the types that can be used for flag values.
+	FlagTypes interface {
+		int64 | float64 | string | bool | any
+	}
+	// StrategyFn is a function type that defines the signature for a strategy function.
```
sahidvelji:
```suggestion
	// StrategyFn defines the signature for a strategy function.
```

openfeature/multiprovider/strategies.go
```diff
@@ -0,0 +1,154 @@
...
+	matchCount := len(matches)
+	switch matchCount {
+	case 0, 1:
+		return msg
+	default:
+		return strings.TrimSpace(string(matches[1]))
+	}
+}
+
+// mergeFlagMeta Merges flag metadata together into a single [of.FlagMetadata] instance by performing a shallow merge.
```
sahidvelji:
```suggestion
// mergeFlagMeta merges flag metadata together into a single [of.FlagMetadata] instance by performing a shallow merge.
```

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,559 @@
...
+
+	// NamedProvider allows for a unique name to be assigned to a provider during a multi-provider set up.
+	// The name will be used when reporting errors & results to specify the provider associated.
+	NamedProvider struct {
+		Name string
+		of.FeatureProvider
+	}
+
+	// Configuration MultiProvider's internal configuration
+	Configuration struct {
```
sahidvelji: If this is internal configuration, does it need to be exported?

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,559 @@
...
+func WithProviderHooks(providerName string, hooks ...of.Hook) Option {
+	return func(conf *Configuration) {
+		conf.providerHooks[providerName] = hooks
+	}
+}
+
+// Multiprovider Implementation
+
+// AsNamedProviderSlice Converts the map into a slice of NamedProvider instances
+func toNamedProviderSlice(m ProviderMap) []*NamedProvider {
```
sahidvelji: Update the doc comment to match the function name

openfeature/multiprovider/comparison_strategy.go
jblacker: You're right. That was removed during some refactoring, but I forgot to update the docs. I'll get this cleaned up. The backticks is just a bad habit from markdown. I'll clean that up too throughout.

openfeature/multiprovider/comparison_strategy.go
jblacker: Updated throughout.

openfeature/multiprovider/strategies.go
jblacker: Ack, I'll update throughout

openfeature/multiprovider/multiprovider.go
jblacker: Good point -- it does not. Updated.

openfeature/multiprovider/multiprovider.go
jblacker: Fixed

codecov.yml
```diff
@@ -1,2 +1,3 @@
 ignore:
-  - "**/*_mock.go"
\ No newline at end of file
+  - "**/*_mock.go"
+  - "multiprovider/discard_handler.go"
```
aepfli: [nit] as we do not have any file matching this name, we should revert this.
```suggestion
```

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,567 @@
...
+	namedEvent struct {
+		of.Event
+		providerName string
+	}
+)
+
+var (
+	_                of.FeatureProvider = (*MultiProvider)(nil)
+	_                of.EventHandler    = (*MultiProvider)(nil)
+	_                of.StateHandler    = (*MultiProvider)(nil)
```
aepfli: [question] i have never seen this kind of construct - based on my current understanding of golang, I would remove those lines - but there might be some special magic I am not aware of :D so is this a deletable leftover, or does this have a purpose?

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,567 @@
...
+	stateValues      map[of.State]int
+	stateTable       [3]of.State
+	eventTypeToState map[of.EventType]of.State
+)
+
+// init Initialize "constants" used for event handling priorities and filtering.
+func init() {
+	// used for mapping provider event types & provider states to comparable values for evaluation
+	stateValues = map[of.State]int{
+		"":            -1, // Not a real state, but used for handling provider config changes
```
aepfli, line 98: [curiosity] could we name this? - most likely not, but it would be cool to have some kind of named string here, for clarity

aepfli (commented):
I started my review, but ran out of time, so here is my frist state of comments. - i will try to review in more depth later

openfeature/multiprovider/multiprovider.go
sahidvelji: These are compile time interface compliance checks. That is, these lines ensure that `*MultiProvider` implements those three interfaces.

sahidvelji: For some reason I can't seem to edit my comment above, but see this: https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance

A comment on the var block saying something like the following would be helpful I think.

```go
// Compile-time interface compliance checks.
```

OTel has several examples of this in their codebase: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go+compile-time&type=code

sahidvelji:
Sorry for the delay! I have not forgotten about this PR and I intend to take another look this week.

openfeature/multiprovider/multiprovider.go
jblacker: I'll add comments for what these are, but as @sahidvelji mentioned these guarantee that the specified interfaces are implemented at compile time.

openfeature/multi/multiprovider.go
jblacker: Unfortunately not. This is used for mapping provider states to provider events, which are almost 1:1. The emitted event means that the provider that emitted the event updated its state, so we need to update that provider's state in our state table and then recalculate the overall state of the multiprovider. The way the mapping works is we map each event to the matching state that event type represents. However, there's one event type, `ProviderConfigChange`, that we don't care about because it does not indicate a state change. This event type is mapped to the empty string, which will become important later. 

I maintain a second map that contains "scores" for each state to determine the overall state of the multiprovider. This is because there is a hierarchy where if one provider is in a "higher" state then the entire multiprovider is in that state. The nominal state score is `0`, and anything above that would represent the multiprovider's overall state not being `StateReady`. The empty string is mapped to the score `-1` to make sure it's lower than anything else and will not affect the overall state. 

The reason I can't name this is that logically I'm not sure it would make sense to do so. Although the underlying type is `string` the type used is `openfeature.State`, which is different from a `string`. This is essentially how enums are implemented in Go. This isn't a state so I'm not sure it would make sense to create a named constant of the empty string that doesn't actually represent a state. I guess I could create a non-exported constant, but I'm still not so sure it makes sense here

openfeature/multi/comparison_strategy.go
```diff
@@ -0,0 +1,239 @@
...
+			set := map[any]struct{}{}
+			for _, v := range values {
+				set[v] = struct{}{}
+			}
+
+			return len(set) == 1
+		}
+		return false
+	}
+}
```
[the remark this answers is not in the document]
jblacker: Handled the potential panic. 

I have no idea what gemini is talking about for #2, I think it's hallucinating because that doesn't seem to make any sense to me. 

For #3 it _might_ be slightly more performant, but I still need to check if these are actually comparable values first. No need to loop twice as gemini is suggesting. Instead if we encounter a nil before calling `reflect.TypeOf` we return false, or if any value in the set creation loop is nil we also short circuit. IMHO this is a simpler & clearer implementation than what it's suggesting.

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,567 @@
...
+	return []of.Hook{}
+}
+
+// BooleanEvaluation returns a boolean flag
+func (mp *MultiProvider) BooleanEvaluation(ctx context.Context, flag string, defaultValue bool, flatCtx of.FlattenedContext) of.BoolResolutionDetail {
+	res := mp.strategy(ctx, flag, defaultValue, flatCtx)
+	return of.BoolResolutionDetail{
+		Value:                    res.Value.(bool),
+		ProviderResolutionDetail: res.ProviderResolutionDetail,
+	}
```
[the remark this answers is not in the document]
jblacker: I believe that Germini is completely wrong here. It seems like it's only evaluating the these methods directly. However, if you look at L229-253 in multiprovider.go specifically within the `NewMultiProvider` function I wrap each provider that has hooks with the `HookIsolator` (or event-supported flavor). There's no need to call `Before`, `After`, `Error`, or `Finally` because that's done automatically by the isolator wrapper whenever an evaluation function is called. Since the isolator implements the same `FeatureProvider` interface as any other provider maybe that confused the AI?

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+	MetadataComparisonDisagreeingProviders = "multiprovider-comparison-disagreeing-providers"
+)
+
+type (
+	// ProviderMap is an alias for a map containing unique names for each included [of.FeatureProvider]
+	ProviderMap = map[string]of.FeatureProvider
+
+	// MultiProvider is an implementation of [of.FeatureProvider] that can execute multiple providers using various
+	// strategies.
+	MultiProvider struct {
```
sahidvelji: Because the name of the package is multiprovider, this will be referred to as `multiprovider.MultiProvider`, which is not ideal. We could also implement the multiprovider in the openfeature package, but that will "pollute" the namespace and documentation in that package with identifiers that many users won't need. Another option is to name the package `multi`, so that you could have `multi.Provider`. OpenTelemetry sort of takes this approach. See the noop package: https://pkg.go.dev/go.opentelemetry.io/otel/trace/noop

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+
+	of "github.com/open-feature/go-sdk/openfeature"
+	"golang.org/x/sync/errgroup"
+)
+
+// Metadata Keys
+const (
+	MetadataProviderName                   = "multiprovider-provider-name"
+	MetadataProviderType                   = "multiprovider-provider-type"
+	MetadataInternalError                  = "multiprovider-internal-error"
```
sahidvelji: It doesn't look like this is being used anywhere in this package.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+	case of.ReadyState:
+		if previousState != of.NotReadyState {
+			l.LogAttrs(context.Background(), slog.LevelInfo, fmt.Sprintf("provider %s has returned to ready state from %s", e.providerName, previousState))
+			return
+		}
+		l.LogAttrs(context.Background(), slog.LevelDebug, fmt.Sprintf("provider %s is ready", e.providerName))
+	case of.StaleState:
+		l.LogAttrs(context.Background(), slog.LevelWarn, fmt.Sprintf("provider %s is stale: %s", e.providerName, e.Message))
+	case of.ErrorState:
+		l.LogAttrs(context.Background(), slog.LevelError, fmt.Sprintf("provider %s is in an error state: %s", e.providerName, e.Message))
```
sahidvelji: Log messages should be static to make the most out of structured logging. Put all dynamic information into attributes. How about something like this:

```go
l.LogAttrs(context.Background(), slog.LevelDebug, "provider has transitioned state",
  slog.String("previousState", previousState),
  slog.String("newState", state),
  slog.String("providerName", e.providerName),
  slog.String("stateMessage", e.Message),
)
```

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+	}
+
+	return false
+}
+
+// updateProviderStateFromEvent updates the state of an internal provider from an event emitted from it, and then
+// re-evaluates the overall state of the multiprovider. If this method returns true the overall state changed.
+func (mp *MultiProvider) updateProviderStateFromEvent(e namedEvent) bool {
+	if e.EventType == of.ProviderConfigChange {
+		mp.logger.LogAttrs(context.Background(), slog.LevelDebug, fmt.Sprintf("ProviderConfigChange event: %s", e.Message))
```
sahidvelji: Log messages should be static.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+	mp.workerGroup.Add(1)
+	go func() {
+		workerLogger := mp.logger.With(slog.String("multiprovider-worker", "event-forwarder-worker"))
+		defer mp.workerGroup.Done()
+		for e := range mp.inboundEvents {
+			l := workerLogger.With(
+				slog.String(MetadataProviderName, e.providerName),
+				slog.String(MetadataProviderType, e.ProviderName),
+			)
+			l.LogAttrs(context.Background(), slog.LevelDebug, fmt.Sprintf("received %s event from provider", e.EventType))
```
sahidvelji: Log messages should be static.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+
+// WithLogger sets a logger to be used with slog for internal logging. By default, all logs are discarded
+func WithLogger(l *slog.Logger) Option {
+	return func(conf *configuration) {
+		conf.logger = l
+	}
+}
+
+// WithTimeout set a timeout for the total runtime for evaluation of parallel strategies
+func WithTimeout(d time.Duration) Option {
```
sahidvelji: It doesn't look like the set timeout duration is being used anywhere. I'm also not sure if we need this. All evaluation methods such as [Client.Boolean](https://pkg.go.dev/github.com/open-feature/go-sdk/openfeature#Client.Boolean) take a context parameter which can be used for timeouts.

openfeature/multiprovider/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+		collectedHooks = append(collectedHooks, wrappedProvider.Hooks()...)
+	}
+
+	multiProvider := &MultiProvider{
+		providers:      providers,
+		outboundEvents: make(chan of.Event),
+		logger:         config.logger,
+		metadata:       buildMetadata(providerMap),
+		totalStatus:    of.NotReadyState,
+		providerStatus: make(map[string]of.State),
```
sahidvelji:
```suggestion
		providerStatus: make(map[string]of.State, len(providers)),
```

openfeature/multiprovider/multiprovider.go
jblacker: ~I don't know if I agree with this. This is logging the event that was propagated up to the multiprovider but we ignore this type of event (`ProviderConfigChange`) so I figured logging it at the debug level would allow any users to debug issues that they may experience.~ Misunderstood initially. I'll get this updated.

openfeature/multiprovider/multiprovider.go
jblacker: ~Ditto here -- this is specifically on the debug level for users that may need additional verbosity for debugging issues. I understand you may disagree, but curious on your reasoning here as in my eyes this seems to only help an end-user, while hiding may be harmful to debugging efforts.~
Ditto on the misunderstanding here.

openfeature/multiprovider/multiprovider.go
jblacker: Ah yeah we can remove this. This is a remnant from prior to @erka helping me refactor the strategies to a simpler generic form. If for some reason we need this in the future it can always be added back as a new feature.

openfeature/multiprovider/multiprovider.go
jblacker: I like the idea of renaming the package `multi` and then renaming `MultiProvider` -> `Provider` to get to `multi.Provider`. I think that's a good compromise on the naming without stuttering. I'll get this updated at some point today.

openfeature/multiprovider/multiprovider.go
jblacker: Oh, you're suggesting to put the message as an attribute? I think I missed that when I replied to another comment about updating this. I can do that. I misunderstood and thought you were saying to not have **any** dynamic content, which was why I was about confused about your reasoning. I'll strike that out of my reply.

openfeature/multiprovider/multiprovider.go
jblacker: Only `MetadataInternalError`  was not used, so I've removed that one. The other ones are all used within the various strategy implementations.

openfeature/multiprovider/multiprovider.go
jblacker: Updated throughout.

jblacker:
@sahidvelji I didn't get through all of the documentation yet, but so far what I've gone over should be up to standards. I'll finish up reviewing the docs tomorrow. I did however un-export a bunch of functions and types that don't need to be exposed to end-users.

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,565 @@
...
+			l.LogAttrs(context.Background(), slog.LevelInfo, "provider has returned to ready state",
+				slog.String(MetadataProviderName, e.providerName), slog.String("previous-state", string(previousState)))
+			return
+		}
+		l.LogAttrs(context.Background(), slog.LevelDebug, "provider is ready", slog.String(MetadataProviderName, e.providerName))
+	case of.StaleState:
+		l.LogAttrs(context.Background(), slog.LevelWarn, "provider is stale",
+			slog.String(MetadataProviderName, e.providerName), slog.String("event-message", e.Message))
+	case of.ErrorState:
+		l.LogAttrs(context.Background(), slog.LevelError, "provider %s is in an error state: %s",
```
sahidvelji: I think you missed this one. Move the dynamic parts of the log message into attributes.

openfeature/multiprovider/multiprovider.go
aepfli: oh i see, and it makes total sense now. thank you for the detailed explanations

openfeature/multi/comparison_strategy.go
```diff
@@ -0,0 +1,249 @@
...
+	"reflect"
+	"slices"
+	"strings"
+
+	of "github.com/open-feature/go-sdk/openfeature"
+	"golang.org/x/sync/errgroup"
+)
+
+// ErrAggregationNotAllowed is an error returned if [of.FeatureProvider.ObjectEvaluation] is called using the [StrategyComparison]
+// strategy without a custom Comparator function configured when response objects are not comparable.
```
sahidvelji:
```suggestion
// strategy without a custom [Comparator] function configured when response objects are not comparable.
```

openfeature/multi/comparison_strategy.go
```diff
@@ -0,0 +1,249 @@
...
+var ErrAggregationNotAllowed error = errors.New(errAggregationNotAllowedText)
+
+// Comparator function used for comparing results of [of.FeatureProvider.ObjectEvaluation]. This is required if returned
+// results are not comparable.
```
sahidvelji:
```suggestion
// Comparator is used to compare the results of [of.FeatureProvider.ObjectEvaluation].
// This is required if returned results are not comparable.
```

openfeature/multi/comparison_strategy.go
```diff
@@ -0,0 +1,249 @@
...
+type Comparator func(values []any) bool
+
+// newComparisonStrategy creates a new instance of ComparisonStrategy. The fallback provider specified is called when
```
sahidvelji: This first sentence seems misleading as there is no identifier named `ComparisonStrategy`. I know this isn't exported to users, but I'd still want this to be more accurate. Maybe something like this? Let me know what you think.
> newComparisonStrategy returns a [StrategyComparison] strategy function.

openfeature/multi/isolation.go
```diff
@@ -0,0 +1,334 @@
...
+// flattenContext converts a [of.EvaluationContext] to a [of.FlattenedContext]
+func flattenContext(evalCtx of.EvaluationContext) of.FlattenedContext {
+	flatCtx := evalCtx.Attributes()
+	flatCtx[of.TargetingKey] = evalCtx.TargetingKey()
+	return flatCtx
+}
+
+// mergeContexts merges attributes from the given EvaluationContexts with the nth [of.EvaluationContext] taking precedence
+// in case of any conflicts with the (n+1)th [of.EvaluationContext].
+func mergeContexts(evaluationContexts ...of.EvaluationContext) of.EvaluationContext {
```
sahidvelji, line 312: There is a mergeContexts func in openfeature/client.go that you can use instead:
https://github.com/open-feature/go-sdk/blob/40fe27eca801dc7ffb4865e403bc2f407fddc339/openfeature/client.go#L836

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+		strategy = config.customStrategy
+	}
+	multiProvider.strategy = strategy
+	multiProvider.strategyName = evaluationStrategy
+
+	return multiProvider, nil
+}
+
+// Providers returns slice of providers wrapped in [NamedProvider] structs.
+func (mp *Provider) Providers() []*NamedProvider {
```
sahidvelji:
```suggestion
func (p *Provider) Providers() []*NamedProvider {
```
Please make the same change throughout

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+// is not a comparable type unless the [WithCustomComparator] [Option] is configured.
+func (mp *Provider) ObjectEvaluation(ctx context.Context, flag string, defaultValue any, flatCtx of.FlattenedContext) of.InterfaceResolutionDetail {
+	res := mp.strategy(ctx, flag, defaultValue, flatCtx)
+	return of.InterfaceResolutionDetail{
+		Value:                    res.Value,
+		ProviderResolutionDetail: res.ProviderResolutionDetail,
+	}
+}
+
+// Init will run the initialize method for all internal [of.FeatureProvider] instances and aggregate any errors. This
```
sahidvelji: Did you mean to write another sentence here? Looks like a typo.

erka:
If we take into account how little free time Jordan has and how perfect this PR needs to be, it will never get merged.... It’s been open since May — we’ll be throwing it a six-month birthday party soon. This is insane 😂 

Please merge it so other members can do the cleanup or close it....

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+	// ProviderMap is an alias for a map containing unique names for each included [of.FeatureProvider]
+	ProviderMap = map[string]of.FeatureProvider
+
+	// Provider is an implementation of [of.FeatureProvider] that can execute multiple providers using various
+	// strategies.
+	Provider struct {
+		providers          ProviderMap
+		metadata           of.Metadata
+		initialized        bool
+		totalStatus        of.State
```
sahidvelji: What does totalStatus refer to? Is `overallStatus` a better name?

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+		totalStatusLock    sync.RWMutex
+		providerStatus     map[string]of.State
+		providerStatusLock sync.Mutex
+		strategyName       EvaluationStrategy    // the name of the strategy used for evaluation
+		strategy           StrategyFn[FlagTypes] // used for custom strategies
```
sahidvelji:
```suggestion
		strategyFunc           StrategyFn[FlagTypes] // used for custom strategies
```

openfeature/multi/multiprovider.go
```diff
@@ -0,0 +1,569 @@
...
+		providerName string
+	}
+
+	// configuration is the internal configuration of a [multi.Provider]
+	configuration struct {
+		useFallback      bool
+		fallbackProvider of.FeatureProvider
+		customStrategy   StrategyFn[FlagTypes]
+		logger           *slog.Logger
+		timeout          time.Duration
```
sahidvelji: I think this timeout can be removed given that we also removed the WithTimeout option.

sahidvelji:
> If we take into account how little free time Jordan has and how perfect this PR needs to be, it will never get merged.... It’s been open since May — we’ll be throwing it a six-month birthday party soon. This is insane 😂
>
>Please merge it so other members can do the cleanup or close it....

You're right that it has taken quite long. To be fair, it is a very large PR and I don't get enough free time to review this in one pass. And even if I approved the PR, I would still need a second approval before being able to merge it anyways.

But I'll try to focus on the exported API only in my review so that we can get this merged and work on other, internal aspects afterwards.

beeme1mr:
Yeah, this is a big one. Let's focus the reviews on the API surface and consider marking it as experimental for now. That's what we did in the [.NET SDK](https://github.com/open-feature/dotnet-sdk#-features).

openfeature/multi/multiprovider.go
jblacker: Removed

openfeature/multi/multiprovider.go
jblacker: Updated

openfeature/multi/multiprovider.go
jblacker: Renamed to `overallStatus` it's a slightly more descriptive name (along with the lock's name too)

openfeature/multi/multiprovider.go
jblacker: Nope, but fixed the typo

openfeature/multi/multiprovider.go
jblacker: Updated throughout

openfeature/multi/comparison_strategy.go
jblacker: Updated

jblacker:
I handled all of the additional comments. IMHO most of the latest comments are nits rather than true blockers, but I've handled them nonetheless. 

I like the idea of considering this experimental until it has some more usage under its belt by someone other than me.

sahidvelji:
@jblacker Ok let's add a package doc comment stating that the package is experimental and that breaking changes are to be expected. Once added, I think we're good to merge.

jblacker:
@erka The DCO issue is because I merged in your PR to my branch too quickly because you had two unsigned commits. @toddbaert said to not worry about it because he can squash them at merge time to eliminate that error.

beeme1mr (approved):

PR #361, opened 2025-05-19
refactor: simplify and improve logging hook

sahidvelji:
## This PR

- add go doc comments
- add a `stage` key to the logs as per [the spec](https://openfeature.dev/specification/appendix-a#logging-hook)
- ensure LoggingHook method receivers are all named `h` (the `buildArgs` method had a receiver named `l`)
- use consistent parameter names for the Before, After, Error, and Finally methods
- simplify the `buildArgs` method by removing the error return parameter
- use [slog.Attr](https://pkg.go.dev/log/slog#Attr) instead of `interface{}` types to log attributes
- the above allows the use of [Logger.LogAttrs](https://pkg.go.dev/log/slog#Logger.LogAttrs), which is [the most efficient form of log output](https://pkg.go.dev/log/slog#Logger.LogAttrs:~:text=For%20the%20most%20efficient,ctx%2C%20%22hello%22%2C%20%22count%22%2C%203)
- as a consequence of the above, the logging performed by the logging hook is now context aware (the `ctx` parameter is passed to `Logger.LogAttrs`)

### DIFF
--- a/openfeature/hooks/logging_hook.go
+++ b/openfeature/hooks/logging_hook.go
@@ -17,6 +17,7 @@ const (
 	REASON_KEY             = "reason"
 	VARIANT_KEY            = "variant"
 	VALUE_KEY              = "value"
+	STAGE_KEY              = "stage"
 )
 
 // LoggingHook is a [of.Hook] that logs the flag evaluation lifecycle.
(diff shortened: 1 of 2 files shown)


aepfli (approved):
Thank you for the contributions. This looks really helpful. I am sadly not as proficient with golang, hence I will wait for another approval before merging

beeme1mr (approved):

sahidvelji:
Good call on tests. I added the `stage` key as expected output in the tests.

aepfli (approved):

beeme1mr:
Hey @sahidvelji, could you please also add a short section to the readme that mentions the hook? Here's an example from [Java](https://github.com/open-feature/java-sdk#logging-hook). It would be nice to include a short code example too (something currently missing from the Java example).

sahidvelji:
> Hey @sahidvelji, could you please also add a short section to the readme that mentions the hook? Here's an example from [Java](https://github.com/open-feature/java-sdk#logging-hook). It would be nice to include a short code example too (something currently missing from the Java example).

@beeme1mr Yes that makes sense, but I'd like to do that in a separate PR. I have created an issue. Feel free to assign me to it.

PR #371, opened 2025-05-24
feat: update telemetry package to conform to latest spec

### LINKED ISSUE #364: Update the OTel-Compatible Telemetry Util Function to latest spec
### Requirements

- [Match latest spec](https://github.com/open-feature/spec/pull/306)
- Update the description to include a disclaimer that this feature is part of the experimental OTel semcon.

> [!NOTE]
> This will be a breaking change, but the current implementation is still [unreleased](https://github.com/open-feature/go-sdk/pull/318).

sahidvelji:
## This PR

Update telemetry package to conform to latest spec. I also refactored a bit and added doc comments.

### Related Issues

Fixes #364

### DIFF
--- a/openfeature/telemetry/telemetry.go
+++ b/openfeature/telemetry/telemetry.go
@@ -32,8 +32,8 @@ const (
 	ResultVariantKey string = "feature_flag.result.variant"
 	ErrorMessageKey  string = "error.message"
 	ContextIDKey     string = "feature_flag.context.id"
-	ProviderKey      string = "feature_flag.provider.name"
-	ReasonKey        string = "feature_flag.result.reason"
+	ProviderNameKey  string = "feature_flag.provider.name"
+	ResultReasonKey  string = "feature_flag.result.reason"
 	FlagSetIDKey     string = "feature_flag.set.id"
 	VersionKey       string = "feature_flag.version"
 )
@@ -51,13 +51,13 @@ const (
 // It is intended to be used in the `Finally` stage of a [openfeature.Hook].
 func CreateEvaluationEvent(hookContext openfeature.HookContext, details openfeature.InterfaceEvaluationDetails) EvaluationEvent {
 	attributes := map[string]any{
-		FlagKey:     hookContext.FlagKey(),
-		ProviderKey: hookContext.ProviderMetadata().Name,
+		FlagKey:         hookContext.FlagKey(),
+		ProviderNameKey: hookContext.ProviderMetadata().Name,
 	}
 
-	attributes[ReasonKey] = strings.ToLower(string(openfeature.UnknownReason))
+	attributes[ResultReasonKey] = strings.ToLower(string(openfeature.UnknownReason))
 	if details.Reason != "" {
-		attributes[ReasonKey] = strings.ToLower(string(details.Reason))
+		attributes[ResultReasonKey] = strings.ToLower(string(details.Reason))
 	}
 
 	body := make(map[string]any)
--- a/openfeature/telemetry/telemetry.go
+++ b/openfeature/telemetry/telemetry.go
@@ -24,86 +24,82 @@ type EvaluationEvent struct {
 	Body map[string]any
 }
 
+// The OpenTelemetry compliant event attributes for flag evaluation.
 const (
-	// The OpenTelemetry compliant event attributes for flag evaluation.
-	// Specification: https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/
+	FlagKey          string = "feature_flag.key"
+	ErrorTypeKey     string = "error.type"
+	ResultValueKey   string = "feature_flag.result.value"
+	ResultVariantKey string = "feature_flag.result.variant"
+	ErrorMessageKey  string = "error.message"
+	ContextIDKey     string = "feature_flag.context.id"
+	ProviderKey      string = "feature_flag.provider.name"
+	ReasonKey        string = "feature_flag.result.reason"
+	FlagSetIDKey     string = "feature_flag.set.id"
+	VersionKey       string = "feature_flag.version"
+)
 
-	TelemetryKey       string = "feature_flag.key"
-	TelemetryErrorCode string = "error.type"
-	TelemetryVariant   string = "feature_flag.variant"
-	TelemetryContextID string = "feature_flag.context.id"
-	TelemetryErrorMsg  string = "feature_flag.evaluation.error.message"
-	TelemetryReason    string = "feature_flag.evaluation.reason"
-	TelemetryProvider  string = "feature_flag.provider_name"
-	TelemetryFlagSetID string = "feature_flag.set.id"
-	TelemetryVersion   string = "feature_flag.version"
+// FlagEvaluationKey is the name of the feature flag evaluation event.
+const FlagEvaluationKey string = "feature_flag.evaluation"
 
-	// Well-known flag metadata attributes for telemetry events.
-	// Specification: https://openfeature.dev/specification/appendix-d#flag-metadata
-	TelemetryFlagMetaContextId string = "contextId"
-	TelemetryFlagMetaFlagSetId string = "flagSetId"
-	TelemetryFlagMetaVersion   string = "version"
-
-	// OpenTelemetry event body.
-	// Specification: https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/
-	TelemetryBody string = "value"
-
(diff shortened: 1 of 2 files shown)


openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
+	VersionKey       string = "feature_flag.version"
 )
 
+// FlagEvaluationKey is the name of the feature flag evaluation event.
+const FlagEvaluationKey string = "feature_flag.evaluation"
+
+const (
+	flagMetaContextIDKey string = "contextId"
+	flagMetaFlagSetIDKey string = "flagSetId"
+	flagMetaVersionKey   string = "version"
```
sahidvelji, line 47: I made these unexported consts.

openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
-
-	// Well-known flag metadata attributes for telemetry events.
-	// Specification: https://openfeature.dev/specification/appendix-d#flag-metadata
-	TelemetryFlagMetaContextId string = "contextId"
-	TelemetryFlagMetaFlagSetId string = "flagSetId"
-	TelemetryFlagMetaVersion   string = "version"
-
-	// OpenTelemetry event body.
-	// Specification: https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/
-	TelemetryBody string = "value"
```
sahidvelji, line 37: I removed the word `Telemetry` from const names because the package name is `telemetry`.

openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
 type EvaluationEvent struct {
-	Name       string
+	// Name is the name of the event.
+	// It is always "feature_flag.evaluation".
+	Name string
+	// Attributes represents the event's attributes.
 	Attributes map[string]any
-	Body       map[string]any
+	// Body is the flag's value and is only present if the flag's variant is empty.
+	Body map[string]any
```
sahidvelji, line 24: I don't see the need to have a separate Body field in this struct. If you agree, I'd like for the value to simply be another attribute.

openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
 		}
+	}
 
-		if details.ErrorMessage != "" {
-			attributes[TelemetryErrorMsg] = details.ErrorMessage
-		}
+	attributes[ErrorTypeKey] = strings.ToLower(string(openfeature.GeneralCode))
+	if details.ErrorCode != "" {
+		attributes[ErrorTypeKey] = strings.ToLower(string(details.ErrorCode))
+	}
```
sahidvelji, line 95: According to the OpenTelemetry spec, these error types should be lowercase. Before this PR they were not.
https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/

erka: This should be double-checked.

Personally, I don’t see any MUST or SHOULD requirement in the OpenTelemetry spec regarding the values of `error.type`. The spec provides examples, but they are just that - examples, not requirements.

On the other hand, the OpenFeature spec defines [these values ](https://openfeature.dev/specification/types#error-code) in uppercase, and other SDKs follow that convention. If the Go SDK sends them in lowercase while others use uppercase, it could lead to inconsistency and confusion across the ecosystem.

I also find the extra step of converting them to lowercase unnecessary and a bit frustrating.

openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
+// It is intended to be used in the `Finally` stage of a [openfeature.Hook].
 func CreateEvaluationEvent(hookContext openfeature.HookContext, details openfeature.InterfaceEvaluationDetails) EvaluationEvent {
 	attributes := map[string]any{
-		TelemetryKey:      hookContext.FlagKey(),
-		TelemetryProvider: hookContext.ProviderMetadata().Name,
+		FlagKey:         hookContext.FlagKey(),
+		ProviderNameKey: hookContext.ProviderMetadata().Name,
 	}
 
+	attributes[ResultReasonKey] = strings.ToLower(string(openfeature.UnknownReason))
```
erka, line 58: The same as `error.type` but for reason. It's well-defined int the [spec](https://openfeature.dev/specification/types#resolution-reason) and should be the same for all sdks.

openfeature/telemetry/telemetry.go
sahidvelji: The footnote says 
> [1] error.type: If one of these values applies, then it MUST be used; otherwise, a custom value MAY be used.

https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/#:~:text=error.type%3A%20If%20one%20of%20these%20values%20applies%2C%20then%20it%20MUST%20be%20used%3B%20otherwise%2C%20a%20custom%20value%20MAY%20be%20used.

Similarly, for `reason`:
> feature_flag.result.reason has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.

https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/#:~:text=feature_flag.result.reason%20has%20the%20following%20list%20of%20well%2Dknown%20values.%20If%20one%20of%20them%20applies%2C%20then%20the%20respective%20value%20MUST%20be%20used%3B%20otherwise%2C%20a%20custom%20value%20MAY%20be%20used.

Additionally, the OpenFeature spec explicitly calls out that
> The error.type and feature_flag.result.reason enumerations use a lowercase "snake_case" convention (see [OpenTelemetry feature-flag log records](https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/)). OpenFeature [error codes](https://openfeature.dev/specification/types#error-code) and [resolution reasons](https://openfeature.dev/specification/types#resolution-reason) should be transformed accordingly by integrations which include this data.

https://openfeature.dev/specification/appendix-d

I agree that it's not ideal that a conversion has to be performed, but I don't see OpenTelemetry changing their spec even though its status is `Development`, because OpenTelemetry uses lowercase snake_case in all other specs too. The other option is for OpenFeature to change their spec to adopt the same lowercase snake_case convention as OpenTelemetry, which is technically possible since the [flag evaluation API spec](https://openfeature.dev/specification/sections/flag-evaluation) has a `Hardening` status.

openfeature/telemetry/telemetry.go
bbland1: I believe `Body` was there because the Otel spec at that time had the value within the body. With the spec update removing it is probably the move.

openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
 	if details.Reason != "" {
-		attributes[TelemetryReason] = strings.ToLower(string(details.Reason))
-	} else {
-		attributes[TelemetryReason] = strings.ToLower(string(openfeature.UnknownReason))
+		attributes[ResultReasonKey] = strings.ToLower(string(details.Reason))
 	}
 
-	body := map[string]any{}
+	body := make(map[string]any)
```
bbland1, line 63: believe the body here can be removed too

openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
 
 	if details.Variant != "" {
-		attributes[TelemetryVariant] = details.Variant
+		attributes[ResultVariantKey] = details.Variant
 	} else {
-		body[TelemetryBody] = details.Value
+		body[ResultValueKey] = details.Value
```
bbland1, line 68:
```suggestion
		attributes[ResultValueKey] = details.Value
```

openfeature/telemetry/telemetry.go
```diff
@@ -6,90 +7,99 @@ import (
...
 	}
 
-	contextID, exists := details.FlagMetadata[TelemetryFlagMetaContextId]
-	if !exists {
-		contextID = hookContext.EvaluationContext().TargetingKey()
+	attributes[ContextIDKey] = hookContext.EvaluationContext().TargetingKey()
```
bbland1, line 71: optional (non-blocking, possible suggestion): assigning `hookContext.EvaluationContext().TargetingKey()` to a temp variable & set `attributes[ContextIDKey]` after the if check which can set the variable if `contextID` is found.

but good as is too 👍🏽

bbland1 (approved):

aepfli (approved):

beeme1mr (approved):
Thank you!

PR #372, opened 2025-05-25
feat!: simplify and improve logging hook

### LINKED ISSUE #367: Remove `MarshaledEvaluationContext`
This struct can be deleted. The correct way to implement nesting using slog is with [slog.Group](https://pkg.go.dev/log/slog#Group). See the accompanying PR.
https://github.com/open-feature/go-sdk/blob/42e33ba0542a59677389d5542b9ee3cc7f8fdb3d/openfeature/hooks/logging_hook.go#L43

### LINKED ISSUE #368: Unexport consts in `hooks` package
Is there a good reason for exporting these constants? Either way, they should be camelCased or PascalCased, not SCREAMING_SNAKE_CASE.
https://github.com/open-feature/go-sdk/blob/42e33ba0542a59677389d5542b9ee3cc7f8fdb3d/openfeature/hooks/logging_hook.go#L11-L20

### LINKED ISSUE #369: Unnecessary error return value in `NewLoggingHook`
This function never returns an error, yet has an error return value.
https://github.com/open-feature/go-sdk/blob/42e33ba0542a59677389d5542b9ee3cc7f8fdb3d/openfeature/hooks/logging_hook.go#L31

### LINKED ISSUE #370: Remove `NewLoggingHook`
I suggest removing `NewLoggingHook` and renaming `NewCustomLoggingHook` to `NewLoggingHook`. I don't see the need to have both, especially since it is very easy for a user that wants to use the default logger to provide it like so:

```go
hooks.NewLoggingHook(true, slog.Default())
```

https://github.com/open-feature/go-sdk/blob/42e33ba0542a59677389d5542b9ee3cc7f8fdb3d/openfeature/hooks/logging_hook.go#L31

https://github.com/open-feature/go-sdk/blob/42e33ba0542a59677389d5542b9ee3cc7f8fdb3d/openfeature/hooks/logging_hook.go#L36

sahidvelji:
## This PR

Building on #361, the logging hook implementation could use additional improvement. Note that all 4 issues are breaking changes, but since hooks are considered experimental, I think it is acceptable.

- Log key consts are now camelCase and unexported.
- The NewLoggingHook function now requires the user to pass in a logger and NewCustomLoggingHook has been removed. It is very easy for a user to get the old behaviour by simply using `hooks.NewLoggingHook(true, slog.Default())` instead of `hooks.NewLoggingHook(true)`.
- The MarshaledEvaluationContext struct has been removed because it was unnecessary and created PascalCase output for the nested keys.
- The NewLoggingHook no longer returns an error, making initialization simpler for the user as they no longer have to check for an error that will never occur.

### Related Issues

fix #367 
fix #370 
fix #369 
fix #368

### DIFF
--- a/openfeature/hooks/logging_hook.go
+++ b/openfeature/hooks/logging_hook.go
@@ -8,16 +8,18 @@ import (
 )
 
 const (
-	DOMAIN_KEY             = "domain"
-	PROVIDER_NAME_KEY      = "provider_name"
-	FLAG_KEY_KEY           = "flag_key"
-	DEFAULT_VALUE_KEY      = "default_value"
-	EVALUATION_CONTEXT_KEY = "evaluation_context"
-	ERROR_MESSAGE_KEY      = "error_message"
-	REASON_KEY             = "reason"
-	VARIANT_KEY            = "variant"
-	VALUE_KEY              = "value"
-	STAGE_KEY              = "stage"
+	domainKey            = "domain"
+	providerNameKey      = "provider_name"
+	flagKeyKey           = "flag_key"
+	defaultValueKey      = "default_value"
+	evaluationContextKey = "evaluation_context"
+	targetingKeyKey      = "targeting_key"
+	attributesKey        = "attributes"
+	errorMessageKey      = "error_message"
+	reasonKey            = "reason"
+	variantKey           = "variant"
+	valueKey             = "value"
+	stageKey             = "stage"
 )
 
 // LoggingHook is a [of.Hook] that logs the flag evaluation lifecycle.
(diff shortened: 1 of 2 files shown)


aepfli (approved):

beeme1mr (approved):
Thanks! This is a good change but is breaking. Normally we would mark that in the PR title but that would trigger a new major version. We'll probably want to manually update the release notes since the breaking change only affects experimental functionality.

FYI @toddbaert @lukas-reining @thomaspoignant

openfeature/hooks/logging_hook.go
```diff
@@ -26,46 +28,35 @@ type LoggingHook struct {
 	logger                   *slog.Logger
 }
 
-// NewLoggingHook returns a new [LoggingHook] with the default logger.
-// To provide a custom logger, use [NewCustomLoggingHook].
-func NewLoggingHook(includeEvaluationContext bool) (*LoggingHook, error) {
-	return NewCustomLoggingHook(includeEvaluationContext, slog.Default())
-}
-
 // NewCustomLoggingHook returns a new [LoggingHook] with the provided logger.
```
toddbaert, line 31:
```suggestion
// NewLoggingHook returns a new [LoggingHook] with the provided logger.
```

PR #373, opened 2025-05-27
test: fix data race condition

alan-kut:
Fix data race condition in the test. Instead of waiting for 200 ms block the initialization.


beeme1mr:
Hey @alan-kut, the tests are passing but there's a lint issue. Please take a look when you have a moment. Thanks!

PR #378, opened 2025-06-03
docs: use `context.TODO()` in README examples

sahidvelji:
## This PR

By convention, it is a good idea to use context.TODO() in examples instead of context.Background() in example code to indicate to the reader that these should be updated. 

### Notes

https://pkg.go.dev/context#TODO


beeme1mr (approved):
Make sense, thanks

PR #381, opened 2025-06-04
refactor: use stdlib packages to get map values

sahidvelji:
## This PR

Use the standard library `maps` and `slices` packages to get a slice of a map's values.

### DIFF
--- a/openfeature/openfeature_api.go
+++ b/openfeature/openfeature_api.go
@@ -3,6 +3,7 @@ package openfeature
 import (
 	"errors"
 	"fmt"
+	"maps"
 	"slices"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


openfeature/openfeature_api.go
```diff
@@ -232,8 +233,10 @@ func (api *evaluationAPI) initNewAndShutdownOld(clientName string, newProvider F
 		return nil
 	}
 
+	namedProviders := slices.Collect(maps.Values(api.namedProviders))
```
aepfli, line 236: [nit] not sure how golang behaves here, but couldn't we actually skip this slices.collect, if the oldprovider is the same as the api.defaultprovider?

I know it improves readability, but now it is executed all the time, before it was more conditionally (not sure how big the impact is ;) ) - just wanted to point out

sahidvelji: That's a good question. I think it's fine. I can't imagine users having more than a handful of named providers registered, so the impact here would be negligible as it is essentially just a loop over the named providers.

PR #386, opened 2025-06-08
chore: use mockgen build_constraint instead of sed logic in Makefile

erka:
## This PR

The `mockgen` task previously relied on Bash-specific syntax in the Makefile to add build constraints to generated mocks. 
This change leverages `mockgen` built-in `-build_constraints` flag, allowing the task to work with any shell.


Makefile
```diff
@@ -3,11 +3,9 @@ GOLANGCI_LINT_VERSION:=v2.1.6
 .PHONY: mockgen
 mockgen:
 	go install go.uber.org/mock/mockgen@latest
```
beeme1mr: Unrelated to this change but I wonder if we should target a specific version (at least major version).

beeme1mr (approved):

Makefile
erka: Good call! I've pinned it to version `v0.5.2`.

Makefile
```diff
@@ -1,13 +1,12 @@
 GOLANGCI_LINT_VERSION:=v2.1.6
+MOCKGEN_VESRION:=v0.5.2
```
sahidvelji: There is a typo here.

erka: thank you!

sahidvelji (approved):

PR #388, opened 2025-06-12
refactor: use `evalCtx` and `flatCtx` consistently

### LINKED ISSUE #387: Inconsistent parameter naming for type `FlattenedContext`
There are many instances in the code where parameters of type `FlattenedContext` are named evalCtx. They should be named `flatCtx` instead since `evalCtx` is already used for parameters of type `EvaluationContext`. Since these two types are distinct, it makes sense to have distinct parameter names for readability and less confusion for users. Parameter names of type `EvaluationContext` should be named `evalCtx` and parameter names of type `FlattenedContext` should be named `flatCtx`.

This would NOT be a breaking change. Relevant documentation should also be updated.

For example, see the FeatureProvider interface:
https://github.com/open-feature/go-sdk/blob/703c3ef2b342dd27536c6f3b2bf114adcf8817b4/openfeature/provider.go#L49-L57

sahidvelji:
### Related Issues

Fixes #387

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,27 +365,27 @@ func (i MyFeatureProvider) Hooks() []openfeature.Hook {
   return []openfeature.Hook{}
 }
 // BooleanEvaluation returns a boolean flag
-func (i MyFeatureProvider) BooleanEvaluation(ctx context.Context, flag string, defaultValue bool, evalCtx openfeature.FlattenedContext) openfeature.BoolResolutionDetail {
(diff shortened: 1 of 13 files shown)


beeme1mr (approved):

erka (approved):

bbland1 (approved):
Nice! it really gives a more consistent vine and read now!

PR #398, opened 2025-06-28
test: don't defer t.Cleanup

bduffany:
The `defer` is redundant, because `t.Cleanup` already ensures that the function is not called until after the test is done.

https://pkg.go.dev/testing#T.Cleanup
> Cleanup registers a function to be called when the test (or subtest) and all its subtests complete. Cleanup functions will be called in last added, first called order.

### DIFF
--- a/openfeature/openfeature_test.go
+++ b/openfeature/openfeature_test.go
@@ -14,7 +14,7 @@ import (
 // The `API`, and any state it maintains SHOULD exist as a global singleton,
 // even in cases wherein multiple versions of the `API` are present at runtime.
 func TestRequirement_1_1_1(t *testing.T) {
-	defer t.Cleanup(initSingleton)
+	t.Cleanup(initSingleton)
 
 	ctrl := gomock.NewController(t)
 	mockProvider := NewMockFeatureProvider(ctrl)
@@ -37,7 +37,7 @@ func TestRequirement_1_1_1(t *testing.T) {
 // The `API` MUST provide a function to set the default `provider`,
 // which accepts an API-conformant `provider` implementation.
 func TestRequirement_1_1_2_1(t *testing.T) {
-	defer t.Cleanup(initSingleton)
+	t.Cleanup(initSingleton)
 	ctrl := gomock.NewController(t)
 
 	mockProvider := NewMockFeatureProvider(ctrl)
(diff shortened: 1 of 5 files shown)


openfeature/openfeature.go
```diff
@@ -89,4 +89,5 @@ func RemoveHandler(eventType EventType, callback EventCallback) {
 // Shutdown active providers
 func Shutdown() {
 	api.Shutdown()
+	initSingleton()
```
sahidvelji: Please undo this change and any changes that do more than just remove `defer`. Other unrelated changes should be part of a separate PR. This change in particular clears the entire state of the API, including all registered providers, the global evaluation context, and all registered global hooks, which is definitely not the intention of `openfeature.Shutdown()`.

sahidvelji (changes requested):

openfeature/openfeature.go
bduffany: I committed this mistakenly - this was some leftover code from when I was debugging a separate issue. Thanks for catching this! Reverted.

bduffany:
> Thanks for the PR. Could you please sign off on your commit(s)? See the [contributing guide](https://github.com/open-feature/go-sdk/blob/main/CONTRIBUTING.md).

Done

openfeature/openfeature_test.go
```diff
@@ -14,7 +14,7 @@ import (
 // The `API`, and any state it maintains SHOULD exist as a global singleton,
 // even in cases wherein multiple versions of the `API` are present at runtime.
 func TestRequirement_1_1_1(t *testing.T) {
-	defer t.Cleanup(initSingleton)
+	t.Cleanup(Shutdown)
```
sahidvelji: The tests are failing, presumably because t.Cleanup calls Shutdown instead of initSingleton.

openfeature/openfeature_test.go
sahidvelji, line 1: I think you missed a couple of `defer` statements in this file. Line 40 for example.

openfeature/openfeature_test.go
bduffany: argh, sorry about that - this was again unintentional. I started from scratch.

```
$ git reset --hard main
$ git ls-files '*_test.go' | xargs perl -pi -e 's@defer t.Cleanup@t.Cleanup@'
$ git add . && git commit -s -m "test: don't defer t.Cleanup" && git push -f
```

erka (approved):

beeme1mr (approved):

PR #400, opened 2025-06-28
fix: only shut down active providers

bduffany:
The spec says that shutdown must shut down active providers,
but the current logic shuts down inactive providers.

This change fixes that. Separately, the spec might also need to be modified
to say "only" active providers.


openfeature/openfeature.go
```diff
@@ -89,4 +89,5 @@ func RemoveHandler(eventType EventType, callback EventCallback) {
 // Shutdown active providers
 func Shutdown() {
 	api.Shutdown()
+	initSingleton()
```
sahidvelji, line 96: As mentioned in [my comment](https://github.com/open-feature/go-sdk/pull/398#discussion_r2173476237) on #398, the call to `initSingleton` clears the entire state of the API, including all registered hooks and providers, which is not what we want. This call to Shutdown should only shutdown whatever providers are active.

toddbaert:
I'll take a look at this next week.

openfeature/openfeature.go
bduffany: @sahidvelji Could you expand on why that is not desired? I would expect a function called "Shutdown" to clear all state and clean up any resources, regardless of resource type (hooks, providers, whatever). When writing tests, we want each test to be able to fully clean up after itself, including any modifications that it made to the global state. That way, subsequent test cases are starting from scratch, and any global state modifications made by previous tests will be reset, so they are guaranteed not to interfere. (Maybe there is a more preferred way of doing this that I'm not seeing?)

sahidvelji: I'd like input from others on this before deciding one way or another, but there are a couple of reasons. The doc comment says this function shuts down active providers. [Based on the spec](https://openfeature.dev/specification/sections/flag-evaluation#17-provider-lifecycle-management), I would expect that to mean that all providers that implement the StateHandler interface and are not in `FATAL` or `NOT_READY` state would be shutdown by calling the provider's Shutdown method. I would then expect each such provider to have a status of NOT_READY. We also need to consider what the other language SDKs do when Shutdown is called, to be consistent. The spec and this doc comment will need to be updated to clarify the shutdown mechanism if the desired behaviour is to reset the entire API upon shutdown.

erka: @sahidvelji The `go-sdk` doesn’t check the provider’s status before calling `provider.Shutdown`. Additionally, if the provider has already been shut down, its internal status isn’t updated. As a result, subsequent calls to `of.Shutdown` will continue attempting to shut it down.

sahidvelji: > @sahidvelji The go-sdk doesn’t check the provider’s status before calling provider.Shutdown. Additionally, if the provider has already been shut down, its internal status isn’t updated.

I realize that, which is why I think that's what this PR should update rather than just resetting everything.

erka: I reviewed a few SDK implementations, and it seems that none of them explicitly handle the provider’s state before calling `Shutdown`. The common pattern is to simply run `Shutdown` on each provider, then either create a new instance of `ProviderRegistry` or clean it up. Some SDKs also set a `NoOpProvider` as the default after shutdown. The python sdk set provider to `NOT_READY` state. 

I think this is a great topic to bring up at the community meeting.

sahidvelji: If there is no consistency among the SDKs, then that implies to me that there is ambiguity in the spec and the section(s) on shutdown need to be clarified further.

toddbaert: While there's room for divergence in implementations, maybe this is something we should specify... I'll consider a spec addition.

toddbaert: @erka @bduffany @sahidvelji What do you guys think about [this spec improvement](https://github.com/open-feature/spec/pull/323)? Is is too much? I'm open to alternatives.

toddbaert:
I hope the specification enhancement [here](https://github.com/open-feature/spec/pull/323) makes the way forward here less ambiguous.

openfeature/openfeature.go
bduffany: @toddbaert I was out for a few weeks but am revisiting this now - that spec change looks great, thanks! I updated this PR to add a new test case specifically for the new section 1.6.2, instead of updating the existing 1.6.1 test. I kept the current implementation of calling `initSingleton` since I think the spec makes it clear that this is how it should behave now, unless I'm misunderstanding. @sahidvelji could you take another look when you have a chance?

openfeature/openfeature_test.go
```diff
@@ -585,6 +585,59 @@ func TestRequirement_1_6_1(t *testing.T) {
 	}
 }
 
+// The API's `shutdown` function MUST reset all state of the API, removing all
+// hooks, event handlers, and providers.
+func TestRequirement_1_6_2(t *testing.T) {
+	defer t.Cleanup(initSingleton)
```
sahidvelji: Is this needed if Shutdown calls initSingleton anyways?

toddbaert (approved):

openfeature/openfeature_test.go
bduffany: It's technically not needed, but I think it makes sense to leave it in, just for consistency with the other tests. Also, it's conceivable that the test might call `t.Fatal` before it gets to `Shutdown()`, so this cleanup ensures that things are always cleaned up (this technically can't happen as the code is currently written, but it's easily conceivable that a minor refactor would make this no longer the case).

sahidvelji (approved):

PR #403, opened 2025-07-05
feat: expose generic resolution/evaluation details structs

sahidvelji:
## This PR

We can introduce generic versions of the ResolutionDetail and EvaluationDetails structs without a breaking change by using type aliases. These generic structs can help provider authors and application authors write more modular code.

### DIFF
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -148,30 +148,24 @@ type EvaluationDetails struct {
 	ResolutionDetail
 }
 
-type BooleanEvaluationDetails struct {
-	Value bool
+// GenericEvaluationDetails represents the result of the flag evaluation process.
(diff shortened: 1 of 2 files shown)


erka:
This change has a side‑effect for code that relies on reflect. I don’t know if this is used in practice somewhere.

```go
reflect.TypeOf(openfeature.BooleanEvaluationDetails{}).Name()
```

Before the change it returns `BooleanEvaluationDetails`.
After the change - `GenericEvaluationDetails[bool]`.

sahidvelji:
Thanks for pointing that out. I can't imagine many users using reflect in that manner. Though, even if there are such users, I would think that this PR would be a welcome change, given that it eliminates the need for such use of reflection in the first place.

aepfli (approved):

Issue #407, opened 2025-07-06
[FEATURE] Refactor telemetry package


sahidvelji:
### Requirements

I would like to improve the `telemetry` package by:

- Omitting the value from the returned attributes. The value is of type `any` and this results in the loss of type safety (use of `map[string]any`. Instead, by omitting the value, the attributes map can be of type `map[string]string`, which is much easier to work with. The value is otherwise very easy to obtain via `details.Value` anyways.
- The "Name" of the evaluation event should not be returned. Instead, since that is already a const in the package, users can just use that const wherever it is needed.
- Refactor the `CreateEvaluationEvent` function to make it work for ALL stages of a hook, rather than just `after` and `finally`. To do that, we could modify the `details openfeature.InterfaceEvaluationDetails` parameter to be a pointer so that a nil check can be performed before adding additional attributes derived from the details. This allows the package to be much more useful.

Related issue: #393

sahidvelji:
> I think the only reason map[string]any is used is because the value isn't always a string. If there's a clean way to support value without losing type safety, I'm 100% in favor of that.

What I am proposing is omitting `value` from the map entirely and have users get that value themselves. It's a minor inconvenience for users. The alternative of including it in the map is far worse than a minor inconvenience because of the loss of type safety IMO.

> @dyladan, do you know if feature_flag.evaluation can be added as a const to the semcon package?

It seems like the semconv package is mostly meant for attributes (key-value pairs) rather than consts, but I'm not certain. I'll let @dyladan confirm, but it would be great to include that given that it's part of the spec.

> Why would you like to run CreateEvaluationEvent in other hook stages? I'm not against the idea; I'm just wondering.

I figured making the package be more general would be useful for the metrics/traces/logging hooks. But on second thought, I think it's fine as is.

beeme1mr:
Okay, I must be misremembering. I'm reluctant to make a breaking change but I'd defer to you if you think it's worth it. It's a bummer that Go's type system is so inflexible.

PR #418, opened 2025-07-26
refactor: use `attribute.KeyValue` instead of map

sahidvelji:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523


openfeature/telemetry/telemetry.go
```diff
@@ -5,96 +5,64 @@ import (
...
-	ProviderNameKey  string = "feature_flag.provider.name"
-	ResultReasonKey  string = "feature_flag.result.reason"
-	FlagSetIDKey     string = "feature_flag.set.id"
-	VersionKey       string = "feature_flag.version"
-)
-
-// FlagEvaluationKey is the name of the feature flag evaluation event.
-const FlagEvaluationKey string = "feature_flag.evaluation"
+// EventName is the name of the feature flag evaluation event.
+const EventName string = "feature_flag.evaluation"
```
bbland1, line 13: this is what you & Michael were asking when it would be added to `semconv` in #407 correct?

sahidvelji: Yes, it doesn't currently exist in the latest semconv 1.34.0, which is why this const is needed for now
https://pkg.go.dev/go.opentelemetry.io/otel@v1.37.0/semconv/v1.34.0#section-readme

openfeature/telemetry/telemetry.go
```diff
@@ -5,96 +5,64 @@ import (
...
 )
 
-// CreateEvaluationEvent creates an [EvaluationEvent].
+// EventAttributes returns a slice of OpenTelemetry attributes that can be used to create an event for a feature flag evaluation.
 // It is intended to be used in the `Finally` stage of a [openfeature.Hook].
-func CreateEvaluationEvent(hookContext openfeature.HookContext, details openfeature.InterfaceEvaluationDetails) EvaluationEvent {
-	attributes := map[string]any{
-		FlagKey:         hookContext.FlagKey(),
-		ProviderNameKey: hookContext.ProviderMetadata().Name,
+func EventAttributes(hookContext openfeature.HookContext, details openfeature.InterfaceEvaluationDetails) []attribute.KeyValue {
```
beeme1mr, line 23: Since this is a new function, should we keep the old one with a deprecation warning? I don't want technical debt to pile up, but this seems like an easy way to avoid abruptly adding a breaking change.

beeme1mr (commented):
I'm in favor of the new approach but I think we should consider not removing the old functionality right away.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.23.0
 require (
 	github.com/cucumber/godog v0.15.0
 	github.com/go-logr/logr v1.4.3
+	go.opentelemetry.io/otel v1.37.0
```
toddbaert, line 8: Generally we like to keep our SDK free of non-test dependencies. Of course, we can make exceptions for some cases, and OTel could be such an exception, but I think I'd advocate for implementing this in the contrib as you alluded to in slack.

I think if we are going to implement it in the SDK directly, we should avoid this dependency. The only SDK that includes OTel bits directly is .NET, but that's a bit of a different case since recent .NET runtimes actually include OTel.

openfeature/telemetry/telemetry.go
toddbaert: We could branch our `v1` now and this could be a breaking change, but I think we should consider [this](https://github.com/open-feature/go-sdk/pull/418/files#r2245958523).

If we want to implement this in contribs instead, to avoid the OTel dep (which would be my preference) I'd recommend we just cut the `v1` maintenance branch now, and then completely remove this function here and implement it in contribs.

WDYT @beeme1mr @sahidvelji ?

go.mod
toddbaert: @beeme1mr @sahidvelji my main question is, how realistic is it for an org to reject the use of OpenFeature because of this dep? Is OTel "controversial" for some orgs? Might a big org that offers telemetry products, which would otherwise adopt OpenFeature, decide not to just because of this dep?

sahidvelji: So Go dependency "tree shaking" happens at the package level. So unless a user imports the `telemetry` package, they will not get the OTel dependency. But again, given that the OTel hooks live in contrib instead of the SDK, I think it might make sense for this package to also live in contrib instead.

beeme1mr: I think this should be fine given how it's used.

erka:
I support your proposal @sahidvelji. Since `semconv` has already provides the majority of necessary attributes, I don’t think the `telemetry` package adds significant value to `go-sdk`. Even the otel hook doesn’t rely on it.

beeme1mr:
Hey @sahidvelji, we added the telemetry method to the SDKs because I thought it would be generally useful (even outside of OTel) and would be dependency-free. If we bring in the OTel semcon dependency, that does change the equation slightly. Arguably, we don't need the dependency because the semcon shouldn't change much. I'm also one of the approvers in OTel, so hopefully changes won't go unnoticed.

Having said that, I don't really have an issue with moving the logic entirely to a hook. Nothing in the OpenFeature spec requires the telemetry function to live in the SDK. The opposite extreme is .NET, where OTel is built into the runtime, and we're considering adding OTel support natively in the SDK itself. I believe end users won't care where this logic works as long as there's an easy way to capture spec-compliant and consistent telemetry signals.

I was working with @lukas-reining on the JavaScript OTel hook last week. Pulling him in to see if he has a strong opinion. If not, I'll defer to @sahidvelji and @erka to decide what makes the most sense in the Go SDK.

sahidvelji:
Any updates here? @beeme1mr @lukas-reining

beeme1mr:
> Any updates here? @beeme1mr @lukas-reining

I still don't have a strong opinion on where this logic should live, as long as the functionality is spec-compliant. I'm hoping either today or tomorrow to expand on the observability appendix in the OpenFeature docs to capture the requirements of the telemetry hook. You can see the PR @lukas-reining wrote [here](https://github.com/open-feature/js-sdk-contrib/pull/1372). The docs I'm planning on writing will be based on the discussions we had while building that.

sahidvelji:
Closing this because we'll have this functionality live in go-sdk-contrib.

PR #421, opened 2025-08-26
refactor: Deprecate GetApiInstance & Add Direct NewDefaultClient implementations

### LINKED ISSUE #350: [DOC] inconsistent documentation for usage of openfeature.NewClient
### Change in the documentation

The documentation for [NewClient](https://pkg.go.dev/github.com/open-feature/go-sdk@v1.14.1/openfeature#NewClient) states

> NewClient returns a new Client. Name is a unique identifier for this client This helper exists for historical reasons. It is recommended to interact with IEvaluation to derive IClient instances.

But, all of the documentation uses NewClient: https://openfeature.dev/docs/reference/technologies/server/go

Should NewClient be marked deprecated? Is the following the intended usage?

```go
client := openfeature.GetApiInstance().GetClient()
```

Should the documentation be updated to replace usage of NewClient with the above?

### LINKED ISSUE #414: Remove the `GetApiInstance` function
> Remove the GetApiInstance function. Exposing the fact that the API is a singleton to users does not make sense. That is an internal implementation detail that users do not need to be aware about. It is also very clunky to have users do `openfeature.GetApiInstance().GetClient()`. Instead, have the user do `openfeature.NewClient()`.

Remove `openfeature.GetApiInstance().GetClient()` and only export `openfeature.NewClient()`

jblacker:
## This PR
completes the non-breaking changes of #414 in preparation of v2.

- Adds `GetClient` & `GetNamedClient` functions directly to the `openfeature` package using the `api` instance
- Mark `GetApiInstance` function as deprecated

### Related Issues

- Fixes #350 
- Fixes #414

### Notes
- This only handles the non-breaking changes as the actual removal of the function

### Follow-up Tasks

Follow up with breaking changes by removing the `GetApiInstance` function and no longer exporting `IEvaluation`


jblacker:
After checking that out a bit more I agree with you. I'll eliminate `GetNamedClient` as it's unnecessary as you're absolutely right that `NewClient` should be preferred. I'll update this.
Since `GetApiInstance().GetClient()` is a common pattern how about we add a `NewDefaultClient()` as the new equivalent that just passes in the empty string for simplicity?

jblacker:
With these changes we could also deprecate `GetClient` & `GetNamedClient` on `IEvaluation` since the implementations are identical and all just call the same private function under the hood. What do ya'll think?

erka:
@jblacker please update the PR title to better reflect the changes.

erka (approved):

openfeature/openfeature.go
```diff
@@ -21,13 +21,20 @@ func initSingleton() {
...
+// Deprecated: Use NewDefaultClient or NewClient directly instead
 //
 //nolint:staticcheck // Renaming this now would be a breaking change.
 func GetApiInstance() IEvaluation {
 	return api
 }
 
+// NewDefaultClient returns a client for the default domain. The default domain client is the IClient instance that
+// wraps around an unnamed FeatureProvider
+func NewDefaultClient() IClient {
```
sahidvelji: Return the concrete type instead of the interface

openfeature/openfeature.go
```diff
@@ -21,13 +21,20 @@ func initSingleton() {
 }
 
 // GetApiInstance returns the current singleton IEvaluation instance.
-// This is the preferred interface to interact with OpenFeature functionalities
+//
+// Deprecated: Use NewDefaultClient or NewClient directly instead
```
sahidvelji:
```suggestion
// Deprecated: use [NewDefaultClient] or [NewClient] directly instead.
```

Use doc links to refer to any exported identifiers.

sahidvelji (approved):

Issue #389, opened 2025-06-15
[FEATURE] Init and Shutdown functions should take `context.Context` parameter


sahidvelji:
### Requirements

The StateHandler interface defines two functions: Init and Shutdown.
https://github.com/open-feature/go-sdk/blob/dd4cf71f8226026a71d51d1edcfd1026c965be15/openfeature/provider.go#L64-L67

There is also a top-level Shutdown function:
https://github.com/open-feature/go-sdk/blob/dd4cf71f8226026a71d51d1edcfd1026c965be15/openfeature/openfeature.go#L90-L92

[Section 2.4 of the specification](https://openfeature.dev/specification/sections/providers#24-initialization) says the following (emphasis mine).
> Many feature flag frameworks or SDKs require some initialization before they can be used. They might require **the completion of an HTTP request, establishing persistent connections, or starting timers or worker threads**. The initialization function is an ideal place for such logic.

The types of initialization mentioned above normally require a `context.Context` parameter to be passed. For example, to make an HTTP request, one most definitely should be passing a `context.Context` parameter.

As for Shutdown, a context can be used to manage timeouts or cancellation.

The context is also the mechanism through which OpenTelemetry spans and other request-scoped values are propagated. If there is any chance at all that a provider author might perform I/O in the Init and Shutdown functions, then these functions need to have a context parameter.

The `context` [package documentation](https://pkg.go.dev/context) has more details on what contexts are used for and where they should be used.

I propose that the Init and Shutdown functions linked at the top take a `context.Context` parameter for the reasons outlined above. While this is a breaking change, I think it is acceptable given that both [section 2.4](https://openfeature.dev/specification/sections/providers#24-initialization) and [section 2.5](https://openfeature.dev/specification/sections/providers#25-shutdown) of the specification are currently experimental.

sahidvelji:
Somewhat related: the Shutdown function does not return an error value, but there could be cases where a shutdown is not successful, in which case the error should be propagated back to the caller.

sahidvelji:
Due to where Shutdown and Init are used in the code, `SetNamedProvider`, `SetNamedProviderAndWait`, `SetProvider`, and `SetProviderAndWait` would all also need to take a context parameter to make this work.

thomaspoignant:
I am not against moving to `2.0`, but as you said we will have to keep supporting the `1.0` in the repo since GO is working like this for the go modules.

thomaspoignant:
[...]

PR #442, opened 2025-10-23
feat: Adds ability to pass a context.Context to provider with new public methods.

leoromanovsky:
## This PR
  

Providers are currently unable to respect timeouts during initialization, which can cause applications to hang indefinitely when providers fail to connect to external services. 

This PR adds optional context-aware initialization methods that enable timeout support while maintaining 100% backward compatibility.

**Key additions:**
* `ContextAwareStateHandler` interface extending `StateHandler` with `InitWithContext(ctx, evalCtx) error`
* New API methods: `SetProviderWithContext`, `SetProviderWithContextAndWait`, `SetNamedProviderWithContext`, `SetNamedProviderWithContextAndWait`
* Smart detection automatically uses context-aware initialization when available, falls back to regular `Init()` for existing providers

👉 This PR is structured to be additive to the v1 branch and be usable immediately.

### Related Issues
  

* https://github.com/open-feature/go-sdk/issues/415
* https://github.com/open-feature/go-sdk/issues/389

### Notes
  

* Zero breaking changes - all existing providers continue to work unchanged
* Thread-safe implementation using existing mutex patterns
* Comprehensive unit test coverage included
* Updated mocks to support new interface methods

### Follow-up Tasks
  
  

* Documentation updates for provider implementation guide
* Example provider implementations showing timeout usage

### How to test
  

```go
// Test timeout functionality
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()

err := openfeature.SetProviderWithContextAndWait(ctx, myProvider)
if err == context.DeadlineExceeded {
    log.Println("Provider initialization timed out")
}
```

Testing passing locally:

```
make test
go test --short -tags testtools -cover -timeout 1m ./...
ok  	github.com/open-feature/go-sdk/e2e	0.280s	coverage: [no statements]
ok  	github.com/open-feature/go-sdk/openfeature	(cached)	coverage: 55.7% of statements
ok  	github.com/open-feature/go-sdk/openfeature/hooks	(cached)	coverage: 100.0% of statements
?   	github.com/open-feature/go-sdk/openfeature/internal	[no test files]
ok  	github.com/open-feature/go-sdk/openfeature/memprovider	(cached)	coverage: 85.2% of statements
ok  	github.com/open-feature/go-sdk/openfeature/multi	(cached)	coverage: 81.6% of statements
ok  	github.com/open-feature/go-sdk/openfeature/telemetry	(cached)	coverage: 100.0% of statements
ok  	github.com/open-feature/go-sdk/openfeature/testing	(cached)	coverage: 91.9% of statements
	github.com/open-feature/go-sdk/pkg/openfeature		coverage: 0.0% of statements
	github.com/open-feature/go-sdk/pkg/openfeature/memprovider		coverage: 0.0% of statements
```

### DIFF
--- a/openfeature/resolution_error.go
+++ b/openfeature/resolution_error.go
@@ -103,6 +103,16 @@ func NewInvalidContextResolutionError(msg string) ResolutionError {
 	}
 }
 
+// NewProviderFatalResolutionError constructs a resolution error with code PROVIDER_FATAL
+//
+// Explanation - The provider is in an irrecoverable error state.
+func NewProviderFatalResolutionError(msg string) ResolutionError {
+	return ResolutionError{
+		code:    ProviderFatalCode,
+		message: msg,
+	}
+}
+
 // NewGeneralResolutionError constructs a resolution error with code GENERAL
 //
 // Explanation - The error was for a reason not enumerated above.
@@ -142,7 +152,7 @@ func (e *ProviderInitError) Error() string {
 //nolint:staticcheck // Renaming these would be a breaking change
 var (
 	// ProviderNotReadyError signifies that an operation failed because the provider is in a NOT_READY state.
-	ProviderNotReadyError = errors.New("provider not yet initialized")
+	ProviderNotReadyError = NewProviderNotReadyResolutionError("provider not yet initialized")
 	// ProviderFatalError signifies that an operation failed because the provider is in a FATAL state.
-	ProviderFatalError = errors.New("provider is in an irrecoverable error state")
+	ProviderFatalError = NewProviderFatalResolutionError("provider is in an irrecoverable error state")
 )


leoromanovsky:
> Hey @leoromanovsky , thank you for starting this.
> 
> I believe the stakeholders agreed that it’s time for v2 and to place context handling in the right places. Breaking changes are expected and accepted by the team - trying to avoid them would only make things more complex in the long run, and the Go SDK would end up with awkward method names or need breaking changes later anyway.
> 
> If you could introduce context into the existing API, that would serve as a solid starting point for v2.

@erka 

Thanks for the context about targeting v2.

I completely understand wanting to work towards larger API shifts there: especially given the set of planned breaking changes like generics and renamed lifecycle methods.

That said, my team at Datadog is preparing to launch on the Go SDK soon, and we’d like to build on a stable release (v1.x) rather than a pre-v2 snapshot. The context-aware lifecycle is a foundational need for us: it lets us manage initialization and shutdown deadlines properly in production.

To balance stability with forward progress, here’s what I’d propose:
* Land this functionality in v1.x as an opt-in bridge (e.g. via a ProviderWithInitContext interface and/or WithInitContext(ctx) option).
* Non-breaking, SemVer-safe, and fully compatible with v2’s upcoming shape. Mark it as provisional in docs (“will be standardized in v2”).
* That gives early adopters a path now, while keeping the mainline clean.

In exchange, I’d love to get a clear picture of what v2 looks like specifically: which of these are considered “in scope” for the breaking release? Is the openfeature specification fully developed? My team is available to help here.

openfeature/provider.go
```diff
@@ -66,6 +66,46 @@ type StateHandler interface {
...
+//		return p.InitWithContext(context.Background(), evalCtx)
+//	}
+//
+// Best practices:
+// - Always check ctx.Done() in long-running initialization operations
+// - Use reasonable timeout values (typically 5-30 seconds)
+// - Return ctx.Err() when the context is cancelled
+// - Maintain backward compatibility by implementing both interfaces
+type ContextAwareStateHandler interface {
+	StateHandler // Embed existing interface for backward compatibility
```
toddbaert, line 84: Good idea.

openfeature/openfeature_api.go
```diff
@@ -88,6 +89,108 @@ func (api *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
...
+	err = api.eventExecutor.registerDefaultProvider(provider)
+	if err != nil {
+		return fmt.Errorf("failed to register default provider %q: %w", provider.Metadata().Name, err)
+	}
+
+	return nil
+}
+
+// SetNamedProviderWithContext sets a provider with client name using context-aware initialization.
+func (api *evaluationAPI) SetNamedProviderWithContext(ctx context.Context, clientName string, provider FeatureProvider, async bool) error {
```
toddbaert, line 130: I dislike that we mix `...AndWait` and `aysnc bool` in our public API - it's inconsistent - but this isn't something you introduced, it already existed elsewhere. I think in v2 we should go for one or the other (the spec doesn't dictate function naming so we have flexibility here). 

I've made a v2 issue: https://github.com/open-feature/go-sdk/issues/443

openfeature/openfeature_api.go
```diff
@@ -88,6 +89,108 @@ func (api *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
...
+		}(api.eventExecutor, api.evalCtx)
+	} else {
+		event, err := initializerWithContext(ctx, newProvider, api.evalCtx)
+		api.eventExecutor.states.Store(clientName, stateFromEventOrError(event, err))
+		api.eventExecutor.triggerEvent(event, newProvider)
+		if err != nil {
+			return err
+		}
+	}
+
```
toddbaert: This is very similar to the existing `initNewAndShutdownOld` - since this is a private func, is it reasonable to try to centralize this and check for nil context?

Same with `initializer`.

openfeature/openfeature_api.go
```diff
@@ -88,6 +89,108 @@ func (api *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
...
+	api.mu.Lock()
+	defer api.mu.Unlock()
+
+	if provider == nil {
+		return errors.New("provider cannot be set to nil")
+	}
+
+	// Initialize new named provider and Shutdown the old one
+	oldProvider := api.namedProviders[clientName]
+	api.namedProviders[clientName] = provider
```
toddbaert, line 140: I think this is now repeated a couple times, perhaps a function can be extracted.

openfeature/openfeature.go
```diff
@@ -1,6 +1,9 @@
 package openfeature
 
-import "github.com/go-logr/logr"
+import (
+	"context"
```
toddbaert, line 4: `make fix` will resolve this lint error.

openfeature/openfeature.go
```diff
@@ -64,6 +128,61 @@ func SetNamedProviderAndWait(domain string, provider FeatureProvider) error {
...
+//	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+//	defer cancel()
+//
+//	// Wait for critical providers to be ready
+//	if err := SetNamedProviderWithContextAndWait(ctx, "critical-service", criticalProvider); err != nil {
+//		return fmt.Errorf("critical provider failed to initialize: %w", err)
+//	}
+//
+//	// Now safe to use the client
+//	client := NewClient("critical-service")
```
toddbaert: This lint error is a bug, I'm 99% sure.

openfeature/openfeature.go
```diff
@@ -47,6 +50,67 @@ func SetProviderAndWait(provider FeatureProvider) error {
...
+//
+//	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+//	defer cancel()
+//
+//	err := SetProviderWithContext(ctx, myProvider)
+//	if err != nil {
+//		log.Printf("Failed to start provider setup: %v", err)
+//		return err
+//	}
+//	// Provider continues initializing in background
```
sahidvelji: Examples should be implemented as Example functions. This is preferred over including examples as Go doc comments because:

1. If the example doesn't compile, tests will fail. This ensures that examples are always up-to-date.
2. Users can run the example on the Go doc site, which can be helpful in some cases.
3. Example functions are used extensively by the Go standard library, and so that approach provides a consistent user experience.

See this blog: https://go.dev/blog/examples
And see this example, for instance: https://pkg.go.dev/github.com/open-feature/go-sdk/openfeature#example-Client.Boolean

openfeature/provider.go
```diff
@@ -66,6 +66,46 @@ type StateHandler interface {
...
+	InitWithContext(ctx context.Context, evaluationContext EvaluationContext) error
```
sahidvelji, line 85: Should we also have ShutdownWithContext here?

lukas-reining:
> If you'd rather just stick to non-breaking enhancements to 1.x for now, I also fully support that, though I have to admit it's not our preference.

I think so too! @toddbaert It would be great to build the enhancement you built @leoromanovsky on the V2 if possible. 
If you need this now, I think it would be acceptable to just include it in a non breaking `1.x` release and rework it in the V2.

> Myself and the rest of the technical committee will fully support you guys in helping move the v2 along if that's the route you'd like to go.

Absolutely!

openfeature/interfaces.go
```diff
@@ -67,6 +67,11 @@ type evaluationImpl interface {
 	SetLogger(l logr.Logger)
 
 	ForEvaluation(clientName string) (FeatureProvider, []Hook, EvaluationContext)
+
+	// Context-aware provider setup methods
+	SetProviderWithContext(ctx context.Context, provider FeatureProvider) error
+	SetProviderWithContextAndWait(ctx context.Context, provider FeatureProvider) error
+	SetNamedProviderWithContext(ctx context.Context, clientName string, provider FeatureProvider, async bool) error
```
sahidvelji, line 75: I think you missed `SetNamedProviderWithContextAndWait` here

openfeature/openfeature_api.go
toddbaert: Should we also add an optional context-aware global shutdown, and insude test if the providers are  `ContextAwareStateHandler` instances? This meant to be called when your application is terminating.

sahidvelji: You mean a context-aware version of this Shutdown function, right? Then yes, I agree that would be good to add.
https://github.com/open-feature/go-sdk/blob/v1.16.0/openfeature/openfeature.go#L101

openfeature/openfeature.go
leoromanovsky: Thanks; adding `make fix` to my workflow before pushing.

openfeature/openfeature_api.go
```diff
@@ -88,6 +90,139 @@ func (api *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
...
+}
+
+// initNewAndShutdownOldWithContext is a context-aware helper to initialise new FeatureProvider and Shutdown the old FeatureProvider.
+func (api *evaluationAPI) initNewAndShutdownOldWithContext(ctx context.Context, clientName string, newProvider FeatureProvider, oldProvider FeatureProvider, async bool) error {
+	return api.initNewAndShutdownOldInternal(&ctx, clientName, newProvider, oldProvider, async)
+}
+
+// initNewAndShutdownOldInternal is the unified helper to initialise new FeatureProvider and Shutdown the old FeatureProvider.
+// If ctx is nil, uses the non-context-aware initializer; otherwise uses the context-aware initializer.
+func (api *evaluationAPI) initNewAndShutdownOldInternal(ctx *context.Context, clientName string, newProvider FeatureProvider, oldProvider FeatureProvider, async bool) error {
```
sahidvelji:
```suggestion
func (api *evaluationAPI) initNewAndShutdownOldInternal(ctx context.Context, clientName string, newProvider FeatureProvider, oldProvider FeatureProvider, async bool) error {
```

openfeature/openfeature_api.go
```diff
@@ -210,46 +379,19 @@ func (api *evaluationAPI) setProvider(provider FeatureProvider, async bool) erro
...
-	if oldProvider == api.defaultProvider || slices.Contains(namedProviders, oldProvider) {
-		return nil
-	}
-
-	go func(forShutdown StateHandler) {
-		forShutdown.Shutdown()
-	}(v)
-
-	return nil
+	return api.initNewAndShutdownOldInternal(nil, clientName, newProvider, oldProvider, async)
```
sahidvelji:
```suggestion
	return api.initNewAndShutdownOldInternal(context.Background(), clientName, newProvider, oldProvider, async)
```
Generally a nil context should never be used. A Background context can be used in its place where a context is not available.

openfeature/openfeature_api.go
sahidvelji: The context parameter should not be a pointer

openfeature/openfeature_api.go
```diff
@@ -88,6 +90,139 @@ func (api *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
...
+func (api *evaluationAPI) initNewAndShutdownOldInternal(ctx *context.Context, clientName string, newProvider FeatureProvider, oldProvider FeatureProvider, async bool) error {
+	if async {
+		go func(executor *eventExecutor, evalCtx EvaluationContext) {
+			// for async initialization, error is conveyed as an event
+			var event Event
+			if ctx != nil {
+				event, _ = initializerWithContext(*ctx, newProvider, evalCtx)
+			} else {
+				event, _ = initializer(newProvider, evalCtx)
+			}
```
sahidvelji: This can be simplified to always call initializerWithContext. I see that initializer calls initializerWithContext with `context.Background()` anyways.

openfeature/example_context_test.go
```diff
@@ -0,0 +1,165 @@
+package openfeature
```
sahidvelji:
```suggestion
package openfeature_test
```
I missed this earlier but Example functions should live in package openfeature_test to ensure that the examples are shown in full.

openfeature/openfeature_api.go
```diff
@@ -1,15 +1,23 @@
...
+	"time"
 
 	"github.com/go-logr/logr"
 )
 
+// DefaultShutdownTimeout is the maximum time to wait for provider shutdown.
+// This prevents hanging if a provider becomes unresponsive during shutdown.
+// 10 seconds allows sufficient time for network cleanup while preventing
+// indefinite hangs during application termination.
+const DefaultShutdownTimeout = 10 * time.Second
```
sahidvelji: I don't think this should be exported. Instead, use Go doc comments on the relevant functions to mention the default.

openfeature/example_context_test.go
sahidvelji: This will require you to update the identifiers in the file and add an import statement. For example,
```go
	err := openfeature.SetProviderWithContext(ctx, provider)
```
instead of
```go
	err := SetProviderWithContext(ctx, provider)
```

openfeature/openfeature_api.go
```diff
@@ -208,48 +366,15 @@ func (api *evaluationAPI) setProvider(provider FeatureProvider, async bool) erro
...
-		return nil
-	}
-
-	go func(forShutdown StateHandler) {
-		forShutdown.Shutdown()
-	}(v)
-
-	return nil
+// initNewAndShutdownOldLegacy is a helper to initialise new FeatureProvider and Shutdown the old FeatureProvider using background context.
+func (api *evaluationAPI) initNewAndShutdownOldLegacy(clientName string, newProvider FeatureProvider, oldProvider FeatureProvider, async bool) error {
```
erka: Please remove this function. Could you also update `evaluationAPI.SetNamedProvider`, `evaluationAPI.Shutdown`, and `evaluationAPI.setProvider` to use the new API methods with `context.Background()`? That should help improve coverage and remove duplication.

openfeature/openfeature_api.go
```diff
@@ -158,6 +282,40 @@ func (api *evaluationAPI) Shutdown() {
...
+		if contextHandler, ok := provider.(ContextAwareStateHandler); ok {
+			if err := contextHandler.ShutdownWithContext(ctx); err != nil && firstError == nil {
+				firstError = fmt.Errorf("named provider %q shutdown failed: %w", name, err)
+			}
+		} else if stateHandler, ok := provider.(StateHandler); ok {
+			stateHandler.Shutdown()
+		}
+	}
+
+	return firstError
```
sahidvelji: I think it would be more useful to join the errors using `errors.Join` so that I can see all errors that occurred rather than just the first one.

sahidvelji: This should be my last comment!

leoromanovsky: Thank you for the careful review!

erka (approved):

openfeature/openfeature_api.go
```diff
@@ -88,6 +93,125 @@ func (api *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
...
+
+	// check for multiple bindings
+	if oldProvider == api.defaultProvider || slices.Contains(namedProviders, oldProvider) {
+		return nil
+	}
+
+	go func(forShutdown StateHandler) {
+		// Check if the provider supports context-aware shutdown
+		if contextHandler, ok := forShutdown.(ContextAwareStateHandler); ok {
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), defaultShutdownTimeout)
```
tangenti: Why creating a new context here instead of passing `ctx`?

openfeature/openfeature_api.go
```diff
@@ -258,6 +367,45 @@ func initializer(provider FeatureProvider, evalCtx EvaluationContext) (Event, er
...
+		event.EventType = ProviderError
+		if errors.Is(ctx.Err(), context.Canceled) {
+			event.Message = "Provider initialization cancelled before start"
+		} else {
+			event.Message = "Provider initialization timed out before start"
+		}
+		return event, ctx.Err()
+	default:
+		// Context is still valid, proceed with initialization
+	}
```
tangenti: There's no IO before this check. Why it's needed?

openfeature/openfeature_api.go
```diff
@@ -258,6 +367,45 @@ func initializer(provider FeatureProvider, evalCtx EvaluationContext) (Event, er
...
+	// Validate that the context is not already done before starting initialization
+	select {
+	case <-ctx.Done():
+		event.EventType = ProviderError
+		if errors.Is(ctx.Err(), context.Canceled) {
+			event.Message = "Provider initialization cancelled before start"
+		} else {
+			event.Message = "Provider initialization timed out before start"
+		}
+		return event, ctx.Err()
```
tangenti: I'd suggest not using both of `Done` and `ctx.Err()`. 

`ctx.Err()` is always nil if `Done` is not closed.

openfeature/openfeature_api.go
```diff
@@ -267,14 +415,13 @@ func initializer(provider FeatureProvider, evalCtx EvaluationContext) (Event, er
 	err := handler.Init(evalCtx)
 	if err != nil {
 		event.EventType = ProviderError
-		event.Message = fmt.Sprintf("Provider initialization error, %v", err)
+		event.Message = fmt.Sprintf("Provider initialization failed: %v", err)
```
tangenti, line 397: It was not a good practice of Golang error message, but it's probably better to keep it in case any automation or error checks exam the error message.

openfeature/openfeature_api.go
sahidvelji: Good catch. Not passing ctx here results in ShutdownWithContext always using the default timeout without being able to override it.

openfeature/openfeature_api.go
beeme1mr: Hey @tangenti, just for my own understanding, what's the recommended way to log a message like this?

openfeature/openfeature_api.go
```diff
@@ -197,17 +197,23 @@ func (api *evaluationAPI) initNewAndShutdownOld(ctx context.Context, clientName
...
-			shutdownCtx, cancel := context.WithTimeout(context.Background(), defaultShutdownTimeout)
-			defer cancel()
+			// Use the passed context but ensure we have a reasonable timeout for shutdown
+			shutdownCtx := parentCtx
+			if deadline, ok := parentCtx.Deadline(); !ok || time.Until(deadline) < defaultShutdownTimeout {
+				// If parent context has no deadline or insufficient time, create a timeout context
+				var cancel context.CancelFunc
+				shutdownCtx, cancel = context.WithTimeout(parentCtx, defaultShutdownTimeout)
+				defer cancel()
+			}
```
sahidvelji: Perhaps to keep things simple, if we avoid the defaultShutdownTimeout and just let the user manage their own timeouts via the provided context, we wouldn't need this logic of checking the deadline and creating our own timeout. We could always add in a default shutdown timeout later if that's desired. Removing it would mean being able to get rid of this code. What do you think?

openfeature/openfeature_api.go
tangenti: If the error is from calling another function, the convention is to use `: %v`

PR #444, opened 2025-10-24
refactor(multi): simplify API with functional options pattern

erka:
## This PR
- Add WithProvider option for registering named providers with optional hooks
- Remove ProviderMap type in favor of variadic functional options
- Change NamedProvider from struct to interface
- Update NewProvider signature to take strategy first, then options
- Add Track method support to MultiProvider for tracking events
- Update tests, documentation and examples to reflect new API


toddbaert (approved):
I have no issues with the changes but I think the title should be `feat(multi): add tracking, functional options` or something like that, since we added that new functionality.

Also it's probably worth noting there's breaking changes here with the multi-provider, but I think that's fine since we explicitly said it's experimental, and it's an auxiliary feature (so let's not add a `!` to the title).

toddbaert:
@erka I'm sorry, I during review I accidentally removed your changes, and GH auto-closes PRs when that happens.

I opened a new one [here](https://github.com/open-feature/go-sdk/pull/446) with all your commits. We can either continue there, or you can try to open this PR again if you move the changes back to this branch (I think that works).

PR #445, opened 2025-10-29
feat: add fatal error constructor

toddbaert:
We have been missing this constructor for a while.

### DIFF
--- a/openfeature/resolution_error.go
+++ b/openfeature/resolution_error.go
@@ -103,6 +103,16 @@ func NewInvalidContextResolutionError(msg string) ResolutionError {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


openfeature/resolution_error.go
```diff
@@ -103,6 +103,16 @@ func NewInvalidContextResolutionError(msg string) ResolutionError {
 	}
 }
 
+// NewProviderFatalResolutionError constructs a resolution error with code INVALID_CONTEXT
```
erka: `INVALID_CONTEXT` typo?

beeme1mr: Good catch... I missed that

toddbaert: I fixed this locally, and never pushed. Thanks!

PR #446, opened 2025-10-29
feat(multi): add tracking, functional options

toddbaert:
Replay of https://github.com/open-feature/go-sdk/pull/444 (these contributions are not mine, but @erka 's)

## This PR
- Add WithProvider option for registering named providers with optional hooks
- Remove ProviderMap type in favor of variadic functional options
- Change NamedProvider from struct to interface
- Update NewProvider signature to take strategy first, then options
- Add Track method support to MultiProvider for tracking events
- Update tests, documentation and examples to reflect new API

### DIFF
--- a/openfeature/multi/multiprovider.go
+++ b/openfeature/multi/multiprovider.go
@@ -54,8 +54,6 @@ type (
 		of.FeatureProvider
 		// Name returns the unique name assigned to the provider.
 		Name() string
-		// unwrap returns the underlying [of.FeatureProvider] instance.
-		unwrap() of.FeatureProvider
 	}
 
 	// namedProvider allows for a unique name to be assigned to a provider during a multi-provider set up.
@@ -93,10 +91,12 @@ type (
 	}
 )
 
+// Name returns the unique name assigned to the provider.
 func (n *namedProvider) Name() string {
 	return n.name
 }
 
+// unwrap returns the underlying [of.FeatureProvider] instance wrapped by this [namedProvider].
 func (n *namedProvider) unwrap() of.FeatureProvider {
 	return n.FeatureProvider
 }
@@ -374,7 +374,7 @@ func (p *Provider) InitWithContext(ctx context.Context, evalCtx of.EvaluationCon
 		prov := provider
 		eg.Go(func() error {
 			l.LogAttrs(ctx, slog.LevelDebug, "starting initialization")
-			if stateHandle, ok := prov.unwrap().(of.StateHandler); ok {
+			if stateHandle, ok := tryAs[of.StateHandler](prov); ok {
 				var err error
 				if contextAwareHandle, ok := stateHandle.(of.ContextAwareStateHandler); ok {
 					err = contextAwareHandle.InitWithContext(ctx, evalCtx)
@@ -394,7 +394,7 @@ func (p *Provider) InitWithContext(ctx context.Context, evalCtx of.EvaluationCon
 				l.LogAttrs(ctx, slog.LevelDebug, "StateHandle not implemented, skipping initialization")
 			}
 			l.LogAttrs(ctx, slog.LevelDebug, "initialization successful")
-			if eventer, ok := prov.unwrap().(of.EventHandler); ok {
+			if eventer, ok := tryAs[of.EventHandler](prov); ok {
 				l.LogAttrs(ctx, slog.LevelDebug, "detected EventHandler implementation")
 				handlers <- namedEventHandler{eventer, name}
 			}
@@ -575,7 +575,7 @@ func (p *Provider) ShutdownWithContext(ctx context.Context) error {
 
 	for _, provider := range p.providers {
 		name := provider.Name()
-		if stateHandle, ok := provider.unwrap().(of.StateHandler); ok {
+		if stateHandle, ok := tryAs[of.StateHandler](provider); ok {
 			meg.Go(func() error {
 				if contextAwareHandle, ok := stateHandle.(of.ContextAwareStateHandler); ok {
 					if err := contextAwareHandle.ShutdownWithContext(ctx); err != nil {
@@ -640,8 +640,28 @@ func (p *Provider) Track(ctx context.Context, trackingEventName string, evaluati
 		}
 	}
 	for _, provider := range providers {
-		if tracker, ok := provider.unwrap().(of.Tracker); ok {
+		if tracker, ok := tryAs[of.Tracker](provider); ok {
 			tracker.Track(ctx, trackingEventName, evaluationContext, details)
 		}
 	}
+}
+
+// tryAs attempts to extract and type-assert the underlying [of.FeatureProvider] from a [NamedProvider].
+// It first checks if the provider implements an unwrap() method to access the wrapped provider,
+// then attempts to cast that provider to type T. Returns the casted value and true if successful,
+// or the zero value of T and false if the provider doesn't support unwrapping or doesn't implement type T.
+// This is used internally to check if wrapped providers implement optional interfaces like
+// [of.StateHandler], [of.EventHandler], or [of.Tracker].
+func tryAs[T any](p NamedProvider) (T, bool) {
+	var v T
+
+	unwrapped, ok := p.(interface {
+		unwrap() of.FeatureProvider
+	})
+	if !ok {
+		return v, false
+	}
+
+	v, ok = unwrapped.unwrap().(T)
+	return v, ok
 }


openfeature/multi/multiprovider.go
```diff
@@ -512,40 +551,56 @@ func logProviderState(l *slog.Logger, e namedEvent, previousState of.State) {
...
+				if contextAwareHandle, ok := p.(of.ContextAwareStateHandler); ok {
+					if err := contextAwareHandle.ShutdownWithContext(ctx); err != nil {
+						errs = append(errs, fmt.Errorf("provider %s shutdown failed: %w", name, err))
+					}
+				} else {
+					p.Shutdown()
+				}
+			}(stateHandle)
+		}
 	}
```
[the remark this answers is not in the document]
erka: /gemini review

openfeature/multi/errors.go
```diff
@@ -44,3 +54,30 @@ func (ae AggregateError) Error() string {
...
+		if err := fn(); err != nil {
+			g.mu.Lock()
+			g.errors = append(g.errors, err)
+			g.mu.Unlock()
+		}
+	}()
+}
+
+// Wait waits for all goroutines to complete.
+// Returns a combined error or nil if none).
```
sahidvelji:
```suggestion
// Returns a combined error or nil if none.
```

openfeature/multi/multiprovider.go
```diff
@@ -46,17 +43,25 @@ type (
...
-		inboundEvents      chan namedEvent
 		workerGroup        sync.WaitGroup
 		shutdownFunc       context.CancelFunc
 		globalHooks        []of.Hook
 	}
 
-	// NamedProvider allows for a unique name to be assigned to a provider during a multi-provider set up.
+	// NamedProvider extends [of.FeatureProvider] by adding a unique provider name. The Name method returns
+	// the assigned provider name, while provider returns the underlying [of.FeatureProvider] instance.
+	NamedProvider interface {
```
sahidvelji, line 53: Does this interface need to be exported?

openfeature/multi/multiprovider.go
```diff
@@ -46,17 +43,25 @@ type (
...
+		of.FeatureProvider
+		Name() string
+		unwrap() of.FeatureProvider
+	}
```
sahidvelji: I suggest adding doc comments on the methods of the interface directly rather than describing the methods in the doc comment of the interface. See for example https://pkg.go.dev/log/slog#Handler

openfeature/multi/multiprovider.go
```diff
@@ -158,39 +179,28 @@ func WithGlobalHooks(hooks ...of.Hook) Option {
 	}
 }
 
-// WithProviderHooks sets [of.Hook] instances that execute only for a specific [of.FeatureProvider]. The providerName
-// must match the unique provider name set during [multi.Provider] creation. This should only be used if you need hooks
-// that execute around a specific provider, but that provider does not currently accept a way to set hooks. This [Option]
-// can be used multiple times using unique provider names. Using a provider name that is not known will cause an error.
-func WithProviderHooks(providerName string, hooks ...of.Hook) Option {
+// WithProvider registers a specific [of.FeatureProvider] instance under the given providerName. The providerName
+// must be unique and correspond to the name used when creating the [multi.Provider]. Optional [of.Hook] instances
```
sahidvelji:
```suggestion
// must be unique and correspond to the name used when creating the [Provider]. Optional [of.Hook] instances
```

openfeature/multi/multiprovider.go
```diff
@@ -158,39 +179,28 @@ func WithGlobalHooks(hooks ...of.Hook) Option {
 	}
 }
 
-// WithProviderHooks sets [of.Hook] instances that execute only for a specific [of.FeatureProvider]. The providerName
-// must match the unique provider name set during [multi.Provider] creation. This should only be used if you need hooks
-// that execute around a specific provider, but that provider does not currently accept a way to set hooks. This [Option]
-// can be used multiple times using unique provider names. Using a provider name that is not known will cause an error.
-func WithProviderHooks(providerName string, hooks ...of.Hook) Option {
```
sahidvelji, line 165: Since we're removing WithProviderHooks here, please update the doc comment of WithGlobalHooks to remove the reference to it.

openfeature/multi/multiprovider.go
```diff
@@ -203,53 +213,53 @@ func buildMetadata(m ProviderMap) of.Metadata {
...
+		if _, ok := provider.FeatureProvider.(of.EventHandler); ok {
+			wrappedProvider = isolateProviderWithEvents(provider, provider.extraHooks)
 		} else {
-			wrappedProvider = isolateProvider(provider, config.providerHooks[name])
+			wrappedProvider = isolateProvider(provider, provider.extraHooks)
 		}
 
-		providers[name] = wrappedProvider
+		providers = append(providers, wrappedProvider)
 		collectedHooks = append(collectedHooks, wrappedProvider.Hooks()...)
```
sahidvelji:
```suggestion
		collectedHooks = slices.Concat(collectedHooks, wrappedProvider.Hooks())
```
I know this isn't part of your PR, but just something I noticed could use improvement.

openfeature/multi/multiprovider.go
```diff
@@ -350,106 +355,137 @@ func (p *Provider) ObjectEvaluation(ctx context.Context, flag string, defaultVal
...
+
+	go func() {
+		wg.Wait()
+		close(pipe)
+	}()
+
+	for e := range pipe {
+		l := workerLogger.With(
+			slog.String(MetadataProviderName, e.providerName),
+			slog.String(MetadataProviderType, e.ProviderName),
```
sahidvelji, line 483: Are these log attrs supposed to have the same values? As in, should MetadataProviderType have value e.ProviderName?

openfeature/multi/multiprovider_test.go
```diff
@@ -18,99 +18,67 @@ func TestMultiProvider_ProvidersMethod(t *testing.T) {
...
 
 	p := mp.Providers()
 	assert.Len(t, p, 2)
-	assert.Regexp(t, regexp.MustCompile("provider[1-2]"), p[0].Name)
-	assert.NotNil(t, p[0].FeatureProvider)
-	assert.Regexp(t, regexp.MustCompile("provider[1-2]"), p[1].Name)
-	assert.NotNil(t, p[1].FeatureProvider)
+	assert.NotNil(t, p[0])
+	assert.Implements(t, (*of.FeatureProvider)(nil), p[0])
+	assert.Regexp(t, regexp.MustCompile("provider1"), p[0].Name())
```
sahidvelji: No need for regexp in this case since the regexp is just a string literal here. Use Assert.Equal.

openfeature/multi/multiprovider_test.go
```diff
@@ -18,99 +18,67 @@ func TestMultiProvider_ProvidersMethod(t *testing.T) {
...
+	assert.NotNil(t, p[1])
+	assert.Implements(t, (*of.FeatureProvider)(nil), p[1])
+	assert.Regexp(t, regexp.MustCompile("provider2"), p[1].Name())
```
sahidvelji: No need for regexp in this case since the regexp is just a string literal here. Use Assert.Equal.

openfeature/multi/multiprovider.go
erka: I think there’s a naming issue here - I’ve struggled with it too. `providerName` is a user-defined name in the `WithProvider` option in `namedEvent`, while `ProviderName` is the defined name in `of.Event`.

openfeature/multi/multiprovider.go
erka: I guess so... There's [User Defined Custom Strategy](https://openfeature.dev/specification/appendix-a/#user-defined-custom-strategy) section which mentions `providerName: string;`.

openfeature/multi/multiprovider.go
```diff
@@ -48,11 +49,12 @@ type (
...
 
-	// NamedProvider extends [of.FeatureProvider] by adding a unique provider name. The Name method returns
-	// the assigned provider name, while provider returns the underlying [of.FeatureProvider] instance.
+	// NamedProvider extends [of.FeatureProvider] by adding a unique provider name.
 	NamedProvider interface {
 		of.FeatureProvider
+		// Name returns the unique name assigned to the provider.
 		Name() string
+		// unwrap returns the underlying [of.FeatureProvider] instance.
 		unwrap() of.FeatureProvider
```
sahidvelji: If this interface is exported, then `unwrap` should also be exported

erka: I don’t want to export it. I’ve reworked it, but it might look a bit messy. wdyt?

sahidvelji: I guess my question is: would it make more sense to unexport the NamedProvider interface and export instead the namedProvider struct, so that the interface is used internally, while the struct is exposed? What we've got now is also fine, but I prefer to minimize exporting interfaces.

erka: I don’t have a strong opinion on this. Let’s see what the other members think.

sahidvelji: I'm ok with merging this as is, especially since this package is marked experimental. We can perform a full review of the exported API before removing its experimental status.

PR #458, opened 2026-01-11
fix: ⚠️ correct execution order for the same level hooks 

### LINKED ISSUE #456: [BUG] Hooks registered at the same level execute out of order
### Observed behavior

Let's say there is hookA and hookB which are registered on the API level

go-sdk v1.17.0

```go
hookA := //....
hookB := //....
openfeature.AddHooks(hookA, hookB)
```
The execution will be
- hookA.before
- hookB.before
- hookA.after
- hookB.after
- hookA.finally
- hookB.finally

### Expected Behavior

The execution should be
- hookA.before
- hookB.before
- hookB.after
- hookA.after
- hookB.finally
- hookA.finally

### Steps to reproduce

https://github.com/open-feature/go-sdk/blob/4b3d6d953076f3056d78c512b31c9fc2ec29ab89/openfeature/hooks_test.go#L790-L822

`mockHook2.After` is missing the call in this test. The correct test should expect call on `mockHook2.After` and return an error and `mockHook1.After` should not be called.

erka:
TODO: mention this PR in CHANGELOG.

---

## This PR

- fix the issue with hook execution order

### Related Issues

closes #456


sahidvelji (approved):

openfeature/client.go
```diff
@@ -807,27 +808,36 @@ func (c *Client) beforeHooks(
...
 	ctx context.Context, hookCtx HookContext, hooks []Hook, evalDetails InterfaceEvaluationDetails, options EvaluationOptions,
 ) error {
-	for _, hook := range hooks {
-		if err := hook.After(ctx, hookCtx, evalDetails, options.hookHints); err != nil {
+	// reverse order
+	for i := len(hooks) - 1; i >= 0; i-- {
+		if err := hooks[i].After(ctx, hookCtx, evalDetails, options.hookHints); err != nil {
 			return err
 		}
 	}
```
sahidvelji:
```suggestion
	for _, hook := range slices.Backward(hooks) {
		if err := hook.After(ctx, hookCtx, evalDetails, options.hookHints); err != nil {
			return err
		}
	}
```

By using the `slices` package, we get slightly cleaner syntax. The same applies to errorHooks and finallyHooks

toddbaert:
@sahidvelji agreed. It's obviously not "breaking" in a compilation sense, but it is in a behavioral sense. I would learn toward calling out in our release notes but not doing a breaking change version.

Reviewing now.

PR #460, opened 2026-01-13
feat!: go sdk v2

erka:
## This PR

- OpenFeature Go SDK v2

### Notes

- The starting point for discussions.
- Play with it
```
go get go.openfeature.dev/openfeature/v2@next
```
- Please read MIGRATION_GUIDE.md


providers/multi/errors.go
```diff
@@ -64,15 +64,13 @@ type multiErrGroup struct {
...
-	go func() {
-		defer g.wg.Done()
+	g.wg.Go(func() {
 		if err := fn(); err != nil {
 			g.mu.Lock()
 			g.errors = append(g.errors, err)
 			g.mu.Unlock()
 		}
-	}()
+	})
```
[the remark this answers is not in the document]
erka: go 1.25 has it

AGENTS.md
```diff
@@ -0,0 +1,250 @@
...
+Excluded rules:
+
+- G101 (hardcoded credentials detection - too noisy)
+- G404 (weak random number generation - used in tests)
+
+### Naming Conventions
+
+- **Variables**: camelCase (e.g., `clientMetadata`, `evalCtx`)
+- **Exported Types/Functions**: PascalCase (e.g., `Client`, `NewClient`)
+- **Interfaces**: PascalCase, typically I-prefixed (e.g., `IClient`, `IFeatureProvider`) or role-based (e.g., `FeatureProvider`)
```
sahidvelji: It's not actually recommended to name interfaces with an `I` prefix. I suggest we omit this part. Also, only exported interfaces can be PascalCase, while unexported interfaces will be camelCase. In general I don't think we need to specify case for types here, as AI coding assistants don't get that wrong in my experience.

providers/inmemory/inmemory_provider.go
sahidvelji, line 1: Should we change the name of the package here to match the name of the directory?

AGENTS.md
erka: I've removed that section

providers/inmemory/inmemory_provider.go
erka: done

interfaces.go
```diff
@@ -2,53 +2,33 @@ package openfeature
...
+// DetailEvaluator defines OpenFeature details evaluator contract.
+type DetailEvaluator interface {
+	BooleanValueDetails(ctx context.Context, flag string, defaultValue bool, evalCtx EvaluationContext, options ...Option) (BooleanEvaluationDetails, error)
+	StringValueDetails(ctx context.Context, flag string, defaultValue string, evalCtx EvaluationContext, options ...Option) (StringEvaluationDetails, error)
+	FloatValueDetails(ctx context.Context, flag string, defaultValue float64, evalCtx EvaluationContext, options ...Option) (FloatEvaluationDetails, error)
+	IntValueDetails(ctx context.Context, flag string, defaultValue int64, evalCtx EvaluationContext, options ...Option) (IntEvaluationDetails, error)
+	ObjectValueDetails(ctx context.Context, flag string, defaultValue any, evalCtx EvaluationContext, options ...Option) (ObjectEvaluationDetails, error)
 }
 
 // IEventing defines the OpenFeature eventing contract
```
sahidvelji: I suggest we rename this to something without the `I` prefix. Maybe `Eventer` or even just `Eventing`?

PR #461, opened 2026-01-13
fix: Handle closed EventChannel from multiprovider

saleemrashid:
## This PR

This fixes a bug where the multiprovider is briefly `READY`, then becomes `NOT_READY`.

The multiprovider closes the event channel under certain circumstances, e.g.:

```go
	if len(handlers) > 0 {
		p.workerGroup.Add(1)
		go p.forwardProviderEvents(workerCtx, handlers)
	} else {
		// we don't emit any events so we can just close the channel
		close(p.outboundEvents)
	}
```

But the `eventExecutor` goroutine copying from the provider's event channel was not checking if it was closed, so it was constantly reading zero-value events. This resulted in `triggerEvent` being called constantly with zero-value events, for which `stateFromEvent` returns `NotReady`.

### DIFF
--- a/openfeature/event_executor_test.go
+++ b/openfeature/event_executor_test.go
@@ -1564,9 +1564,9 @@ func TestEventHandler_ChannelClosure(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, executor.activeSubscriptions, 1)
 
-	var eventCount int64
+	var eventCount atomic.Int64
 	callBack := func(e EventDetails) {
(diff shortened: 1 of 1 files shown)


dangra:
> Could you please merge [this PR](https://github.com/superfly/open-feature-go-sdk/pull/1) into your branch?

@erka : Saleem is away until later today. I merged it in case we can save another sun cycle :)  thanks!

beeme1mr (approved):

PR #464, opened 2026-01-23
fix(memprovider): use int64 type in IntEvaluation to match API

leoromanovsky:
## Motivation

When using the InMemoryProvider for testing with integer flags, users encounter unexpected TYPE_MISMATCH errors even when storing values as int64:

```
  provider := memprovider.NewInMemoryProvider(map[string]memprovider.InMemoryFlag{
      "max-retries": {
          State:          memprovider.Enabled,
          DefaultVariant: "high",
          Variants:       map[string]any{"high": int64(10)},
      },
  })

  result := provider.IntEvaluation(ctx, "max-retries", 1, nil)
  // Expected: result.Value == 10
  // Actual: result.Value == 1, result.ResolutionError == TYPE_MISMATCH
```

This makes the InMemoryProvider unusable for integer flag testing.

## Root Cause

The IntEvaluation method calls genericResolve[int] (line 98), but the FeatureProvider interface specifies int64:

```
  // provider.go:55 - The official interface
  IntEvaluation(ctx context.Context, flag string, defaultValue int64, flatCtx FlattenedContext) IntResolutionDetail
```

The genericResolve helper performs a type assertion value.(T). When T is int but the stored value is int64, the assertion fails because int64 != int in Go's type system, even on 64-bit platforms.

Compare with FloatEvaluation which correctly uses genericResolve[float64].

##  Changes

  1. in_memory_provider.go: Changed genericResolve[int] to genericResolve[int64] to match the API contract
  2. in_memory_provider_test.go: Updated test to use explicit int64() types
  3. testprovider_test.go: Updated test to use explicit int64() types

## Design Decision

  This fix requires users to explicitly type integer variants as int64:

```
  // Correct usage after this fix
  Variants: map[string]any{"high": int64(10)}

  // Will not work (untyped constant becomes int)
  Variants: map[string]any{"high": 10}
```

  This is consistent with:
  - The FeatureProvider interface contract which specifies int64
  - The FloatEvaluation behavior which requires float64
  - The documented return type IntEvaluationDetails = GenericEvaluationDetails[int64]

An alternative would be to handle both int and int64 in genericResolve, but this would add complexity and deviate from the pattern used by other evaluation methods.

### DIFF
--- a/openfeature/memprovider/in_memory_provider.go
+++ b/openfeature/memprovider/in_memory_provider.go
@@ -159,6 +159,9 @@ func (i InMemoryProvider) find(flag string) (*InMemoryFlag, *openfeature.Provide
 // genericResolve is a helper to extract type verified evaluation and fill openfeature.ProviderResolutionDetail.
 // It coerces smaller numeric types to their canonical forms (int* -> int64, float32 -> float64)
 // to provide a more forgiving API for test flag configuration.
+//
+// Note: Only signed integer types are supported for conversion. Unsigned integer types
+// (uint, uint8, uint16, uint32, uint64) will result in type mismatch errors.
 func genericResolve[T comparable](value any, defaultValue T, detail *openfeature.ProviderResolutionDetail) T {
 	// Try direct type assertion first
 	if v, ok := value.(T); ok {


openfeature/memprovider/in_memory_provider.go
```diff
@@ -156,14 +156,46 @@ func (i InMemoryProvider) find(flag string) (*InMemoryFlag, *openfeature.Provide
...
+		case float32:
+			return any(float64(v)).(T)
+		case int8:
+			return any(float64(v)).(T)
+		case int16:
+			return any(float64(v)).(T)
+		case int32:
+			return any(float64(v)).(T)
+		case int:
+			return any(float64(v)).(T)
```
erka:
```suggestion
```
I think I might’ve overdone it.

openfeature/memprovider/in_memory_provider.go
```diff
@@ -156,14 +156,38 @@ func (i InMemoryProvider) find(flag string) (*InMemoryFlag, *openfeature.Provide
 
 // helpers
 
-// genericResolve is a helper to extract type verified evaluation and fill openfeature.ProviderResolutionDetail
+// genericResolve is a helper to extract type verified evaluation and fill openfeature.ProviderResolutionDetail.
+// It coerces smaller numeric types to their canonical forms (int* -> int64, float32 -> float64)
+// to provide a more forgiving API for test flag configuration.
```
erka, line 161:
```suggestion
// to provide a more forgiving API for test flag configuration.
//
// Note: Only signed integer types are supported for conversion. Unsigned integer types
// (uint, uint8, uint16, uint32, uint64) will result in type mismatch errors.
```

PR #470, opened 2026-02-04
fix(testing): make test provider context-aware for goroutine

### LINKED ISSUE #436: [BUG] testing.TestProvider can't reliably be used when tests invoke separate goroutines
### Observed behavior

Currently the [`testing.TestProvider`](https://pkg.go.dev/github.com/open-feature/go-sdk@v1.16.0/openfeature/testing#TestProvider) makes use of goroutine-local variables to store the current test's name, so that the provider can be scoped to a given test.

This has the unfortunate side-effect that tests that need to launch goroutines may panic with the `unable to detect test name; be sure to call `UsingFlags`	in the scope of a test (in T.run)!")` [message](https://github.com/open-feature/go-sdk/blob/v1.16.0/openfeature/testing/testprovider.go#L78C10-L78C103).

### Expected Behavior

Ideally, the TestProvider should be able to operate reliably when goroutines are involved

### Steps to reproduce

_No response_

erka:
:warning: The TestFramework interface now requires Context() context.Context, which may break custom implementations.

## This PR
  - Added context-based value resolution for TestProvider.
  - Preserved existing goroutine-local behavior as a fallback.

## Related issues
closes #436


openfeature/testing/testprovider.go
```diff
@@ -28,12 +31,38 @@ type TestProvider struct {
...
 }
 
-type TestFramework = interface{ Name() string }
+type TestFramework = interface {
+	Name() string
+}
+
+// testFrameworkWithContext is an optional interface for tests that can provide
+// a context, enabling context-aware evaluation when flags are used across goroutines.
+type testFrameworkWithContext interface {
```
hairyhenderson: what's the purpose of this secondary interface? why not just update `TestFramework` to contain `Context() context.Context`? it'd be much simpler that way...

for that matter, why not use [`testing.TB`](https://pkg.go.dev/testing#TB) instead? the `TestFramework` interface seems like an unnecessary abstraction.

README.md
```diff
@@ -514,14 +513,28 @@ for name, tt := range tests {
...
+```go
+// In your test, 't' is a *testing.T
+ctx := testProvider.UsingFlags(t, tt.flags)
+
+go func() {
+    // Make sure to use the context returned by UsingFlags in the new goroutine.
+    // The context carries the necessary information for the TestProvider.
+    _ = openfeature.NewDefaultClient().Boolean(ctx, "my_flag", false, openfeature.EvaluationContext{})
+}()
+
```
hairyhenderson, line 531: the closing code-fence is missing here (to match the opening one on ln519)

erka: thank you!

openfeature/testing/testprovider.go
erka: I would love to use `testing.TB`, but there will likely be someone who doesn’t use it in a specific case, and their tests or code might fail. Since go-sdk is already at v1, that could make people unhappy. So I work around...

openfeature/testing/testprovider.go
```diff
@@ -71,11 +100,18 @@ func (tp TestProvider) Metadata() openfeature.Metadata {
...
 }
 
-func (tp TestProvider) getProvider() openfeature.FeatureProvider {
+func (tp TestProvider) getProvider(ctx context.Context) openfeature.FeatureProvider {
 	// Retrieve the test name from the goroutine-local storage.
-	testName, ok := getGoroutineLocal().(string)
+	//
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
sahidvelji: I think we can remove this. If ctx is nil here, then it's a programmer error that would be masked by silent substitution with context.Background().

erka: thanks for the suggestion

PR #474, opened 2026-02-16
fix: consistent SetProviderAndWait init flow

dd-oleksii:
:warning: Providers that lazily initialize their `EventChannel()` in `Init()` rather than at construction time may be impacted by this change; these should be verified to return a valid `EventChannel()` before `Init()` is called.

## Context

OpenFeature specification defines SetProviderAndWait as a waiting version of SetProvider, or as a shortcut for waiting on provider ready event. However, currently SetProvider and SetProviderAndWait exhibit non-trivial behavior differences besides waiting.

SetProvider runs initialization asynchronously and potentially concurrently with shutdown of the old provider. The API is not blocked and the application author may initialize other providers concurrently, run evaluations, etc.

🐛: in this mode, the error from initializer is ignored when updating the provider state, so fatal/error states may be not set properly.

SetProviderAndWait runs initialization synchronously while holding exclusive api.mu lock. This almost completely locks OpenFeature SDK: the application author cannot initialize other providers (for different domains), configure context or hooks, evaluate feature flags, or shutdown SDK. If a provider initialization blocks forever'ish, the SDK remains unusable and is unrecoverable.

Another difference is that old provider is shutdown only after new provider has successfully initialized.

🐛: if the new provider fails to initialize, the old provider is already unset in API but will never be shutdown, and the new provider is not registered with api.eventExecutor (so if it comes back online after some time, nobody listens to its events, and the state will go out of sync if old provider continues emitting events).

🐛: in both modes, given that shutdown is run concurrently with updating subscriptions in eventExecutor, it is possible for the old provider to override the state of the new provider:
1. init finishes, emits provider ready event (directly from goroutine), updates state
2. old provider emits some event during shutdown (e.g., PROVIDER_ERROR or PROVIDER_STALE), eventExecutor receives the event and updates the state to error/stale
3. new provider is registered with eventExecutor but the state is already wrong.

## This PR

This PR introduces a couple of changes:

Make initialization flow consistent across both modes: always initialize async but make "AndWait" methods wait for initialization outside of critical section. Make init respect returned error.

Always call shutdown on old provider (if it is no longer used).

Always register new provider with event executor. Do this before we start init/shutdown, so the old provider cannot influence state of the new provider. Make event executor registration non-erroring by making shutdown channel buffered (there were no good way to recover from registration error).

Refactored `SetProviderXxx` methods to remove unnecessary duplication.

### Related Issues

Can't find any.

### DIFF
--- a/openfeature/multi/multiprovider.go
+++ b/openfeature/multi/multiprovider.go
@@ -118,7 +118,6 @@ var (
 func init() {
 	// used for mapping provider event types & provider states to comparable values for evaluation
 	stateValues = map[of.State]int{
-		"":            -1, // Not a real state, but used for handling provider config changes
 		of.ReadyState: 0,
 		of.StaleState: 1,
 		of.ErrorState: 2,
@@ -130,10 +129,9 @@ func init() {
 		of.ErrorState, // 2
 	}
 	eventTypeToState = map[of.EventType]of.State{
-		of.ProviderConfigChange: "",
-		of.ProviderReady:        of.ReadyState,
-		of.ProviderStale:        of.StaleState,
-		of.ProviderError:        of.ErrorState,
+		of.ProviderReady: of.ReadyState,
+		of.ProviderStale: of.StaleState,
+		of.ProviderError: of.ErrorState,
 	}
 }
 
@@ -483,6 +481,16 @@ func (p *Provider) forwardProviderEvents(workerCtx context.Context, handlers cha
 			slog.String(MetadataProviderType, e.ProviderName),
 		)
 		l.LogAttrs(workerCtx, slog.LevelDebug, "received event from provider", slog.String("event-type", string(e.EventType)))
+
+		// ConfigurationChanged events are always forwarded directly without affecting provider state tracking.
+		// This matches the JS SDK reference behavior where ConfigurationChanged is re-emitted as a direct
+		// pass-through, independent of status change logic.
+		if e.EventType == of.ProviderConfigChange {
+			p.outboundEvents <- e.Event
+			l.LogAttrs(workerCtx, slog.LevelDebug, "forwarded configuration changed event")
+			continue
+		}
+
 		if p.updateProviderStateFromEvent(e) {
 			p.outboundEvents <- e.Event
 			l.LogAttrs(workerCtx, slog.LevelDebug, "forwarded state update event")
@@ -509,10 +517,8 @@ func (p *Provider) updateProviderState(name string, state of.State) bool {
 
 // updateProviderStateFromEvent updates the state of an internal provider from an event emitted from it, and then
 // re-evaluates the overall state of the multiprovider. If this method returns true the overall state changed.
+// Note: ProviderConfigChange events are handled separately in forwardProviderEvents and never reach this method.
 func (p *Provider) updateProviderStateFromEvent(e namedEvent) bool {
-	if e.EventType == of.ProviderConfigChange {
-		p.logger.LogAttrs(context.Background(), slog.LevelDebug, "ProviderConfigChange event", slog.String("event-message", e.Message))
-	}
 	p.providerStatusLock.Lock()
 	previousState := p.providerStatus[e.providerName]
 	p.providerStatusLock.Unlock()
--- a/openfeature/multi/multiprovider_test.go
+++ b/openfeature/multi/multiprovider_test.go
@@ -371,6 +371,93 @@ func TestMultiProvider_StateUpdateWithSameTypeProviders(t *testing.T) {
 	}
 }
 
+func TestMultiProvider_ConfigurationChangedEventForwarding(t *testing.T) {
+	// awaitEvent drains the outbound channel until an event of the given type is found.
+	awaitEvent := func(t *testing.T, ch <-chan of.Event, eventType of.EventType) of.Event {
+		t.Helper()
+		var found of.Event
+		require.Eventually(t, func() bool {
+			for {
+				select {
+				case e, ok := <-ch:
+					if !ok {
+						return false
+					}
+					if e.EventType == eventType {
+						found = e
+						return true
+					}
+				default:
+					return false
+				}
+			}
+		}, time.Second, 10*time.Millisecond, "expected %s event was not received", eventType)
+		return found
+	}
+
+	// setup creates a two-provider multi-provider, initializes it, and waits for READY.
+	setup := func(t *testing.T) (*Provider, *mockProviderWithEvents, *mockProviderWithEvents) {
+		t.Helper()
+		ctrl := gomock.NewController(t)
+		t.Cleanup(ctrl.Finish)
+
+		p1 := newMockProviderWithEvents(ctrl, "provider1")
+		p2 := newMockProviderWithEvents(ctrl, "provider2")
+
+		mp, err := NewProvider(
+			StrategyFirstMatch,
+			WithProvider("provider1", p1),
+			WithProvider("provider2", p2),
+		)
+		require.NoError(t, err)
+		t.Cleanup(mp.Shutdown)
+
+		require.NoError(t, mp.Init(of.NewEvaluationContext("test", nil)))
+		require.Eventually(t, func() bool {
+			return mp.Status() == of.ReadyState
+		}, time.Second, 10*time.Millisecond)
+
+		return mp, p1, p2
+	}
+
+	t.Run("event is forwarded with correct payload and does not corrupt provider state", func(t *testing.T) {
+		mp, provider1, _ := setup(t)
+
+		provider1.EmitEvent(of.ProviderConfigChange, "flags updated")
+		e := awaitEvent(t, mp.outboundEvents, of.ProviderConfigChange)
+
+		assert.Equal(t, "flags updated", e.Message)
+		assert.Equal(t, "provider1", e.EventMetadata[MetadataProviderName])
+
+		// Provider state should remain READY (not corrupted to empty string)
+		mp.providerStatusLock.Lock()
+		assert.Equal(t, of.ReadyState, mp.providerStatus["provider1"])
+		assert.Equal(t, of.ReadyState, mp.providerStatus["provider2"])
+		mp.providerStatusLock.Unlock()
+
+		assert.Equal(t, of.ReadyState, mp.Status())
+	})
+
+	t.Run("does not affect aggregate state when another provider is degraded", func(t *testing.T) {
+		mp, provider1, provider2 := setup(t)
+
+		// Put provider2 into STALE state
+		provider2.EmitEvent(of.ProviderStale, "stale")
+		awaitEvent(t, mp.outboundEvents, of.ProviderStale)
+		require.Equal(t, of.StaleState, mp.Status())
+
+		// ConfigurationChanged from provider1 should not affect aggregate
+		provider1.EmitEvent(of.ProviderConfigChange, "flags updated")
+		awaitEvent(t, mp.outboundEvents, of.ProviderConfigChange)
+
+		assert.Equal(t, of.StaleState, mp.Status())
+
+		mp.providerStatusLock.Lock()
+		assert.Equal(t, of.ReadyState, mp.providerStatus["provider1"])
+		mp.providerStatusLock.Unlock()
+	})
+}
+
 func TestMultiProvider_Track(t *testing.T) {
 	t.Run("forwards tracking to all ready providers that implement Tracker", func(t *testing.T) {
 		ctrl := gomock.NewController(t)


openfeature/openfeature_api.go
```diff
@@ -203,8 +159,27 @@ func (api *evaluationAPI) initNewAndShutdownOld(ctx context.Context, clientName
...
+func (api *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
+	api.mu.RLock()
+	defer api.mu.RUnlock()
+
+	provider, ok := api.namedProviders[name]
+	if !ok {
+		return ProviderMetadata()
+	}
+
+	return provider.Metadata()
```
dd-oleksii, line 183: These were not changed—just moved around to make _all_ provider setters close together.

openfeature/event_executor.go
```diff
@@ -192,32 +190,32 @@ func (e *eventExecutor) State(domain string) State {
...
 	oldProvider := e.defaultProviderReference
 	e.defaultProviderReference = newProvider
 
-	return e.startListeningAndShutdownOld(newProvider, oldProvider)
+	e.startListeningAndShutdownOld(newProvider, oldProvider)
 }
 
 // registerNamedEventingProvider registers a named FeatureProvider and remove event listener for old named provider
-func (e *eventExecutor) registerNamedEventingProvider(associatedClient string, provider FeatureProvider) error {
+func (e *eventExecutor) registerNamedEventingProvider(associatedClient string, provider FeatureProvider) {
```
erka, line 205: Hey @dd-oleksii 

Could you please keep it as it is? This change doesn’t help fix the issue and makes it harder to understand and merge into the v2 branch.

dd-oleksii: What exactly do you want to keep as is? Changes in this file fix a potential error to shutdown an event listener, and remove an error condition from which openfeatureAPI cannot recover cleanly (except ignoring it)

sahidvelji:
The tests seem to be failing with the `-race` flag. Though I see that the Makefile `test` target does not include the `-race` flag, so we'll need to update that. For now, please run the tests locally with
```
go test -count=1 -race --short -tags testtools -cover -timeout 1m ./...
```

openfeature/event_executor.go
erka: I don’t see how `return nil` is different from `return`, or why slog in this file is better than in another one. If the caller isn’t interested in the error, they can simply ignore it, right?

dd-oleksii:
@sahidvelji thanks for the hint! I have fixed two tests that relied on events emitted before initialization to be received _after_ initialization is complete. I fixed two tests to emit error/stale events _after_ provider initialization is complete.

openfeature/event_executor.go
dd-oleksii: The caller _is_ interested in handling errors. The `error` return type signals that errors are possible and _should_ be handled.

If registering new provider fails/errors, the proper handling would be to stop setting new provider and keep the old one. But then eventExecutor is not handling that properly either, leaving SDK in a limbo state between the old and the new provider. Returning that error to the user is not actionable either.

The change in this file is not simply reshuffling where the error is ignored:
1. It removes the wait.
2. It _forbids_ registration to error.
3. It makes error impossible (not ignored).

I can help merging this change into `next` if that's the only thing holding it back.

dd-oleksii: Merge into `next`: https://github.com/open-feature/go-sdk/pull/475

dd-oleksii:
oh yeah, it's the same flawed pattern in this test as well:
1. Provider emits some event (`ProviderStale` in this case)
2. Provider is set in the API
3. API runs initialization, emits provider ready event, and sets the state to ready
4. Test expects the state to be "stale" (while the spec expects it to be "ready")

In the old "and wait" implementation, there was a bug that the SDK would only start listening for provider events _after_ initialization has succeeded. So it would receive ProviderStale event _after_ ProviderReady, despite the fact that it was issued _before_ the initialization was attempted.

In the non-waiting implementation, the subscription and initialization happen concurrently, so it's undefined what event arrives first in this situation and what the end state would be. That's why the test is flaky. This PR moves subscription start earlier in the chain (before we attempt initialization), so it's more likely that the pre-existing events arrive first — that's why it started flaking more often.

There are two main issues here:

First, the tests are just making a wrong assumption about the ordering of events. This one is easy to fix — we should emit events only after initialization (or make initialization return error).

The harder issue is ensuring consistent ordering of pre-existing events and the initialization. I guess the simplest way to achieve that is by draining the provider's event channel _before_ starting initialization. Another option is to send the provider ready event through the same channel, so the ordering is preserved. (In either case, it's probably best to handle this in a separate PR — this one is trying to fix too much things simultaneously.

openfeature/openfeature_api.go
```diff
@@ -34,163 +34,117 @@ func newEvaluationAPI(eventExecutor *eventExecutor) *evaluationAPI {
...
-			executor.states.Store(clientName, stateFromEventOrError(event, nil))
-			executor.triggerEvent(event, provider)
-		}(api.eventExecutor, api.evalCtx, ctx, newProvider, clientName)
-	} else {
-		event, err := initializerWithContext(ctx, newProvider, api.evalCtx)
-		api.eventExecutor.states.Store(clientName, stateFromEventOrError(event, err))
-		api.eventExecutor.triggerEvent(event, newProvider)
 		if err != nil {
-			return err
+			err = fmt.Errorf("failed to initialize named provider %q for domain %q: %w", provider.Metadata().Name, clientName, err)
```
toddbaert: I think should check the domain here and use a different message if one is set or not (to be less confusing and also consistent with the previous logging).

`err = fmt.Errorf("failed to initialize default provider %q: %w", provider.Metadata().Name, err)` vs `err = fmt.Errorf("failed to initialize named provider %q for domain %q: %w", provider.Metadata().Name, clientName, err)`

openfeature/openfeature_api.go
```diff
@@ -34,163 +34,117 @@ func newEvaluationAPI(eventExecutor *eventExecutor) *evaluationAPI {
...
-func (api *evaluationAPI) SetNamedProviderWithContextAndWait(ctx context.Context, clientName string, provider FeatureProvider) error {
-	return api.SetNamedProviderWithContext(ctx, clientName, provider, false)
-}
+func (api *evaluationAPI) initNew(ctx context.Context, clientName string, newProvider FeatureProvider) <-chan error {
+	errCh := make(chan error, 1)
+
+	// Initialize new provider async. The caller may wait on the channel.
+	go func(executor *eventExecutor, evalCtx EvaluationContext, ctx context.Context, provider FeatureProvider, clientName string) {
+		event, err := initializerWithContext(ctx, provider, evalCtx)
+		executor.states.Store(clientName, stateFromEventOrError(event, err))
```
toddbaert:
```suggestion
```

`initNew`  is writing provider state via outside the `e.mu` lock in the executor, then `executor.triggerEvent` is writing it again inside the lock... I think that unprotected first write could race with a simultaneous provider event... It also seems redundant unless I'm missing something.

I think we can remove this unprotected write and let triggerEvent be the single place this state is updated. I removed this line locally and all tests pass... so if this IS important, we are missing a test for it.

dd-oleksii: Good catch! I moved this piece from old `initNewAndShutdownOld` without checking, so this race also exists in main

dd-oleksii: +, verified that`triggerEvent` properly updates the state

openfeature/openfeature_api.go
dd-oleksii: 👍 fixed

openfeature/openfeature_api.go
dd-oleksii: although `states` is a `sync.Map` and has its own locking, it's probably a bad idea still to bypass executor

toddbaert:
Pushed one additional change to remove some now-dead code. Will merge Monday unless I hear objections cc @erka @sahidvelji

sahidvelji (approved):

erka:
@toddbaert This PR introduces a breaking change relative to the OpenFeature spec. According to the spec, if a provider emits an event and there is a subscriber for that event, the subscriber event handlers must be run.

The test changes clearly demonstrate this breakage - they were modified to hide the underlying problem rather than address it. The `eventExecutor` is buggy in how it tracks states with events from a non-current provider, but dropping those non-current provider events is not an appropriate fix, in my opinion.

erka:
@toddbaert Honestly, the simplest way to address all of this complexity might be to prohibit provider replacement at the OpenFeature spec level. Do we know if providers are actually replaced in production scenarios?

Let me bring up another elephant: tracking. From what I can see in the go sdk, tracking relies on the current (default/domain) provider. A client may perform an evaluation using one provider, then that provider has been replaced, and only afterward client receives a Track call. In that situation, which provider is expected to handle tracking?

openfeature/openfeature_api.go
```diff
@@ -34,163 +34,120 @@ func newEvaluationAPI(eventExecutor *eventExecutor) *evaluationAPI {
...
+func (api *evaluationAPI) setProviderWithContext(ctx context.Context, provider FeatureProvider) <-chan error {
 	api.mu.Lock()
 	defer api.mu.Unlock()
 
-	if provider == nil {
-		return errors.New("default provider cannot be set to nil")
-	}
-
 	oldProvider := api.defaultProvider
 	api.defaultProvider = provider
```
erka: This could be dangerous now with `SetProviderAndWait`. Mutex is released and `ForEvaluation` could return not initialized provider for client evaluations.

erka:
There is also a change in how `SetProviderAndWait` works: the `StateHandler` provider may receive evaluation requests before it is initialized. Is this behavior expected?

This is an example
```go
func main() {
	client := openfeature.NewDefaultClient()
	go func() {
		for {
			<-time.After(time.Microsecond)
			result, err := client.BooleanValueDetails(context.TODO(), "flag", false, openfeature.NewTargetlessEvaluationContext(nil))
			fmt.Println(result, err)
		}
	}()
	time.Sleep(100 * time.Microsecond)
	go func() {
		provider, _ := flagd.NewProvider()
		_ = openfeature.SetProviderAndWait(provider)
		fmt.Println("flagd set and wait done")
	}()
	time.Sleep(time.Second)
}

```

Before output:
```shell
//...
{false {flag bool {default-variant DEFAULT   map[]}}} <nil>
flagd set and wait done
{false {flag bool { ERROR PROVIDER_NOT_READY connection not made map[]}}} error code: PROVIDER_NOT_READY: connection not made
//...
```

After output

```shell
//...
{false {flag bool {default-variant DEFAULT   map[]}}} <nil>
{false {flag bool {    map[]}}} PROVIDER_NOT_READY: provider not yet initialized
//...
{false {flag bool {    map[]}}} PROVIDER_NOT_READY: provider not yet initialized
flagd set and wait done
{false {flag bool { ERROR PROVIDER_NOT_READY connection not made map[]}}} error code: PROVIDER_NOT_READY: connection not made
//...
```

erka:
OpenFeature Requirement 2.4.1 says `Many feature flag frameworks or SDKs require some initialization before they can be used.`. I read it as non-initialized provider should not receive evaluation request calls. 

The thing about `EventChannel()` call. There is no guidance when this channel should be initialized. If some provider creates that channel in `Init`, this PR breaks their implementation. 

This PR aims to address certain problems, but it also introduces new ones. I see it more as a breaking change not as a bug fix.

dd-oleksii:
> Since the Go SDK is v1, some users might be surprised by the new behavior of `SetProviderAndWait`. Here’s a small app to observe some of the changes.

> The client reports Ready, but evaluations say the provider is not ready / not initialized

oh, this one looks like another bug in main that we can fix: setting new provider does not reset state to not ready as it should. You can observe this behavior with `SetProvider` today.

Otherwise, your example demonstrates two spec violations that got fixed:
1. Evaluations calls were blocked which goes against [1.4.12](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1412) which suggests that evaluation methods should be non-blocking. The current implementation may block evaluation calls _indefinitely_ long which is a serious issue and is the reason why I started the PR.
2. The SDK didn't call the client event handler after initialization. Clearly violates [5.3.2](https://openfeature.dev/specification/sections/events#requirement-532)

Now the questions is whether fixing these two is a breaking change or not. My argument for "not" is that these were _undocumented_ deviations from the spec. So if anyone relied on these, they were clearly relying on undocumented behavior that was not part of the public API. But I can see the point that it might be considered a breaking change if we want to play safe ([obligatory xkcd reference](https://xkcd.com/1172/)).

@erka so how do you want to proceed? do you want me to bump the major version? do you want to bundle these changes with your branch? do you want to fix the bugs in a different way?

PR #483, opened 2026-03-26
feat: Add targeting match split reason

AleksandrasMaliuginas-vinted:
## This PR

- adds TARGETING_MATCH_SPLIT reason

### Notes

This reason is used in go-feature-flag lib: https://github.com/thomaspoignant/go-feature-flag/blob/916ae7747b17a62dc904ee6a5f02a57c7147da25/modules/core/flag/resolution_reason.go#L15


toddbaert:
Hey @AleksandrasMaliuginas-vinted . Reasons are free-form. Reasons can be added custom for whatever; but I wouldn't add a default reason in an SDK without first adding it to the [spec](https://github.com/open-feature/spec). Please start there if you think this should be done. Closing for now.

PR #492, opened 2026-04-22
feat: implement isolated API instances

### LINKED ISSUE #486: Support isolated API instances
The spec now supports creating independent, non-singleton instances of the OpenFeature API. Each instance has its own providers, context, hooks, and events; totally separate from the global singleton and from each other. This is useful for things like DI frameworks and testing.

### What needs to happen

- Add a factory function (or equivalent) that creates new, independent API instances
- Each instance should conform to the same API contract as the existing singleton
- Ideally the factory lives in its own package for discoverability (recommended, not required)
- A single provider instance shouldn't be bound to more than one API instance at a time

### Links

- Parent issue: https://github.com/open-feature/spec/issues/374
- Spec section: https://openfeature.dev/specification/sections/flag-evaluation#18-isolated-api-instances

marcozabel:
## This PR
This PR introduces support for creating isolated OpenFeature API instances, each with their own providers, hooks, context, and event handling - enabling multi-tenant or side-by-side usage without shared global state.

### Related Issues
https://github.com/open-feature/go-sdk/issues/486


openfeature/isolated_api_test.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// Singleton should still have the default NoopProvider
+	if api.GetProviderMetadata().Name == "instance-provider" {
+		t.Error("provider set on isolated instance leaked into the singleton")
+	}
+}
+
+// Requirement 1.8.1 (independence): State set on the singleton MUST NOT affect isolated instances.
+func TestIsolatedAPI_SingetonDoesNotAffectInstance(t *testing.T) {
```
sahidvelji:
```suggestion
func TestIsolatedAPI_SingletonDoesNotAffectInstance(t *testing.T) {
```

openfeature/openfeature.go
```diff
@@ -18,12 +18,33 @@ func init() {
...
+//
+// Each instance conforms to the same [IEvaluation] contract as the global singleton (spec 1.8.2).
+// Per spec 1.8.4, a provider instance SHOULD NOT be bound to more than one API instance at a time;
+// attempting to do so will return an error from [IEvaluation.SetProvider] or [IEvaluation.SetNamedProvider].
+//
+// Callers MUST invoke [IEvaluation.Shutdown] when the instance is no longer needed to release
+// provider resources and free the provider bindings held by the global registry.
+//
+// Use [IEvaluation.GetClient] or [IEvaluation.GetNamedClient] to create clients bound to this instance.
+func NewAPI() IEvaluation {
```
sahidvelji: I want to avoid returning an interface here. I suggest we export [evaluationAPI] and return this concrete type instead of the interface.

openfeature/openfeature_api.go
```diff
@@ -84,25 +165,61 @@ func (api *evaluationAPI) SetNamedProviderWithContextAndWait(ctx context.Context
 	return api.SetNamedProviderWithContext(ctx, clientName, provider, false)
 }
 
+// maybeUnbindOldProvider removes oldProvider from the global binding registry if it is no longer
+// referenced by this API instance. Must be called with api.mu held.
+// All comparisons use pointer identity (via providerBindingKey) to avoid panics from unhashable
+// FeatureProvider implementations that contain maps or slices.
+func (api *evaluationAPI) maybeUnbindOldProvider(oldProvider FeatureProvider) {
```
sahidvelji: The word "maybe" seems odd in this method name. Perhaps `unbindIfUnreferenced`?

openfeature/openfeature.go
```diff
@@ -18,12 +18,33 @@ func init() {
...
+// providers, evaluation context, hooks, and events. This satisfies spec requirement 1.8.1.
+//
+// Each instance conforms to the same [IEvaluation] contract as the global singleton (spec 1.8.2).
+// Per spec 1.8.4, a provider instance SHOULD NOT be bound to more than one API instance at a time;
+// attempting to do so will return an error from [IEvaluation.SetProvider] or [IEvaluation.SetNamedProvider].
+//
+// Callers MUST invoke [IEvaluation.Shutdown] when the instance is no longer needed to release
+// provider resources and free the provider bindings held by the global registry.
+//
+// Use [IEvaluation.GetClient] or [IEvaluation.GetNamedClient] to create clients bound to this instance.
```
toddbaert: A couple of things: section 1.8 is experimental in the spec, so worth noting that in the doc comment (e.g. `// Experimental: ...` at the top). Also, the spec recommends (1.8.3) that the factory live in a distinct module/package for discoverability; in Go this would typically be a sub-package like `openfeature/isolated`. WDYT?

openfeature/isolated_api_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	instance := NewAPI()
+	if instance.GetProviderMetadata().Name == "singleton-provider" {
+		t.Error("singleton provider leaked into newly created isolated instance")
+	}
+}
+
+// Requirement 1.8.4: A provider instance SHOULD NOT be bound to more than one API instance at a time.
+func TestRequirement_1_8_4_CrossInstanceBinding(t *testing.T) {
+	t.Cleanup(func() {
```
toddbaert, line 74: The provider isolation tests are solid and I like that both directions are covered. Worth adding similar tests for hooks, evaluation context, and events between isolated instances to match the coverage in other SDKs.

erka:
`IEvaluation` interface and `openfeature` package functions are out of sync like `SetProviderWithContext` is missing, `NewDefaultClient` is `GetClient`(?), etc... `func GetApiInstance() IEvaluation {` is deprecated. 

I don't see that we implement `Requirement 1.8.2: Isolated instances MUST conform to the same API contract as the singleton` here. Should there be the table driven test which runs against `GetApiInstance` and `NewAPI`? 

The developer's experience isn't great after all in my opinion. If someone tries a new isolated API they will have to rewrite the code. It's not just plug and play @toddbaert

openfeature/openfeature.go
toddbaert: Thanks for adding the experimental note! Just bumping the second half of this; any thoughts on moving `NewAPI()` to a sub-package like `openfeature/isolated`? Per 1.8.3 the factory should be in a distinct module/package so it's intentionally less discoverable than the singleton. WDYT?

toddbaert:
I pushed a rebase, specifically because @erka recently deleted a bunch of code which this PR benefits from.

toddbaert:
> `IEvaluation` interface and `openfeature` package functions are out of sync like `SetProviderWithContext` is missing, `NewDefaultClient` is `GetClient`(?), etc... `func GetApiInstance() IEvaluation {` is deprecated.
> 
> I don't see that we implement `Requirement 1.8.2: Isolated instances MUST conform to the same API contract as the singleton` here. Should there be the table driven test which runs against `GetApiInstance` and `NewAPI`?
> 
> The developer's experience isn't great after all in my opinion. If someone tries a new isolated API they will have to rewrite the code. It's not just plug and play @toddbaert

@erka Ya, fair point about the DX. We had similar discoveries in other languages. Because we need to allocate the new instance, as far as I can tell, it's not feasible to make this completely transparent. We can't have a simple import allocate a new isolated singleton; a factory function needs to be exported somewhere and its returned reference kept by the caller. I don't see a way around that.

toddbaert:
Added an `openfeature/isolated` sub-package exposing `NewAPI()` (per spec 1.8.3 "factory in a distinct module/package"); consistent with the Kotlin and PHP SDKs. The original `openfeature.NewAPI()` is unexported now; the sub-package reaches the constructor via an `openfeature/internal/factory` bridge so the construction path stays hidden from external callers.

toddbaert (approved):

toddbaert:
Rebased to fix conflicts - will merge Fri unless somebody gives me a reason not to.

@sahidvelji I agree that returning a struct is better [as you mentioned here](https://github.com/open-feature/go-sdk/pull/492#discussion_r3140984518), but I just want to call out that it's more surface area we are adding here, and I don't think it's fair to mark that one experimental.

If you want to keep it, I'm not opposed to it.

toddbaert (changes requested):
@erka noticed that this doesn't include the recent fixes for leak goroutines fix [here](https://github.com/open-feature/go-sdk/pull/508). There also might be some races related to this. Blocking for now.

toddbaert (approved):
> @erka noticed that this doesn't include the recent fixes for leak goroutines fix [here](https://github.com/open-feature/go-sdk/pull/508). There also might be some races related to this. Blocking for now.

@erka I pushed an update

- cleans up event-related goroutines in isolated API instances as you mentioned
- tests for the above are moved/centralized, there's no point in duplicating these for the package singleton if we do them for instances
- trimmed some AI-autocompleted docs which were too verbose
- removed useless empty `init()`

I inspected the global state changes here and can't find any obvious races. Let me know if you can.

erka (approved):

toddbaert:
@erka after merging your branch I made some small tweaks (`api()` was shadowing our receiver, so I renamed it) and I made a couple small tweaks and removed your TODO.

I think this looks good, but two things:

- ~do we want to remove `eventingImpl` as you suggested? TBH I think we have too many internal interfaces, but I don't feel too strongly about it, and it could be done in a different PR if we want to at all~
  - I've removed this for now
- we've gone back to returning `IEvaluation`, which @sahidvelji wasn't a fan of, but again, but I think that it's safer and less breaking; I wish we'd not returned this interface historically, consistent with Go idioms, but I think "the cat is out of the bag" now.

erka:
Thanks for cleanups @toddbaert!

It's a great time to think about `EvaluationAPI` struct and what it should look like. I don't want to rush into that in this PR.

PR #494, opened 2026-04-23
fix(event_executor): update provider status before invoking API-level handlers

### LINKED ISSUE #493: BUG: update provider status before invoking API-level event handlers
In `event_executor.go`, `triggerEvent()` dispatches API-level (global) handlers _before_ calling `states.Store()`. This means a global handler that reads client state will see the stale value.

Client-scoped handlers are fine; `states.Store()` runs before those.

The fix should be straightforward: update state for all relevant domains before dispatching any handlers.

## Spec reference

[Requirement 5.3.5](https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md#requirement-535):

> If the provider emits an event, the value of the client's provider status MUST be updated to the status associated with that event **before** the SDK invokes any event handlers for that event, so that handlers observe a consistent status.

Spec PR clarifying this: https://github.com/open-feature/spec/pull/367

SAY-5:
Fixes open-feature/go-sdk#493.

Per OpenFeature spec [requirement 5.3.5](https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md#requirement-535):

> If the provider emits an event, the value of the client's provider status MUST be updated to the status associated with that event **before** the SDK invokes any event handlers for that event, so that handlers observe a consistent status.

`triggerEvent` previously dispatched API-level (global) handlers before calling `states.Store`, so a global handler that queried client state observed the stale status. Client-scoped handlers were fine — `states.Store` ran just before those.

## Fix

Collect the domain writes first (every named domain matching this provider, plus `defaultDomain` if the handler is the default provider), call `states.Store` for each, then dispatch API-level → named-domain → default-domain handlers in the same order as before. No change to handler ordering, just the state-write ordering.

## Test

`go test -tags testtools ./openfeature/...` clean.

### DIFF
--- a/openfeature/event_executor.go
+++ b/openfeature/event_executor.go
@@ -297,29 +297,40 @@ func (e *eventExecutor) triggerEvent(event Event, handler FeatureProvider) {
 	e.mu.Lock()
 	defer e.mu.Unlock()
 
-	// first run API handlers
+	// Per OpenFeature spec requirement 5.3.5, the client's provider
+	// status MUST be updated BEFORE any event handlers (API-level or
+	// client-scoped) fire, so every handler — including API-level
+	// handlers that query client state — observes the new status.
+	// Write the state for every domain affected by this handler's
+	// provider reference first, then dispatch handlers.
+	isDefault := e.defaultProviderReference.equals(newProviderRef(handler))
+	namedDomains := make([]string, 0)
+	for domain, reference := range e.namedProviderReference {
+		if reference.equals(newProviderRef(handler)) {
+			namedDomains = append(namedDomains, domain)
+			e.states.Store(domain, stateFromEvent(event))
+		}
+	}
+	if isDefault {
+		e.states.Store(defaultDomain, stateFromEvent(event))
+	}
+
+	// first run API handlers (after state has been written above)
 	for _, c := range e.apiRegistry[event.EventType] {
 		e.executeHandler(*c, event)
 	}
 
-	// then run client handlers
-	for domain, reference := range e.namedProviderReference {
-		if !reference.equals(newProviderRef(handler)) {
-			continue
-		}
-
-		e.states.Store(domain, stateFromEvent(event))
+	// then run client handlers bound to a named domain matching this provider
+	for _, domain := range namedDomains {
 		for _, c := range e.scopedRegistry[domain].callbacks[event.EventType] {
 			e.executeHandler(*c, event)
 		}
 	}
 
-	if !e.defaultProviderReference.equals(newProviderRef(handler)) {
+	if !isDefault {
 		return
 	}
 
-	// handling the default provider
-	e.states.Store(defaultDomain, stateFromEvent(event))
 	// invoke default provider bound (no provider associated) handlers by filtering
 	for domain, registry := range e.scopedRegistry {
 		if _, ok := e.namedProviderReference[domain]; ok {
--- a/openfeature/event_executor.go
+++ b/openfeature/event_executor.go
@@ -297,22 +297,20 @@ func (e *eventExecutor) triggerEvent(event Event, handler FeatureProvider) {
 	e.mu.Lock()
 	defer e.mu.Unlock()
 
-	// Per OpenFeature spec requirement 5.3.5, the client's provider
-	// status MUST be updated BEFORE any event handlers (API-level or
-	// client-scoped) fire, so every handler — including API-level
-	// handlers that query client state — observes the new status.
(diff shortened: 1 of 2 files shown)


openfeature/event_executor.go
```diff
@@ -297,29 +297,40 @@ func (e *eventExecutor) triggerEvent(event Event, handler FeatureProvider) {
...
+	// Per OpenFeature spec requirement 5.3.5, the client's provider
+	// status MUST be updated BEFORE any event handlers (API-level or
+	// client-scoped) fire, so every handler — including API-level
+	// handlers that query client state — observes the new status.
+	// Write the state for every domain affected by this handler's
+	// provider reference first, then dispatch handlers.
+	isDefault := e.defaultProviderReference.equals(newProviderRef(handler))
+	namedDomains := make([]string, 0)
+	for domain, reference := range e.namedProviderReference {
+		if reference.equals(newProviderRef(handler)) {
```
toddbaert: `newProviderRef(handler)` is loop invariant here; I think we can hoist it out of the loop entirely as well.

toddbaert: Oh ya, Gemini found the same thing: https://github.com/open-feature/go-sdk/pull/494#discussion_r3133558739

openfeature/event_executor.go
```diff
@@ -297,29 +297,40 @@ func (e *eventExecutor) triggerEvent(event Event, handler FeatureProvider) {
...
 
-	// first run API handlers
+	// Per OpenFeature spec requirement 5.3.5, the client's provider
+	// status MUST be updated BEFORE any event handlers (API-level or
+	// client-scoped) fire, so every handler — including API-level
+	// handlers that query client state — observes the new status.
+	// Write the state for every domain affected by this handler's
+	// provider reference first, then dispatch handlers.
+	isDefault := e.defaultProviderReference.equals(newProviderRef(handler))
+	namedDomains := make([]string, 0)
```
sahidvelji:
```suggestion
	namedDomains := make([]string, 0, len(e.namedProviderReference))
```

SAY-5:
@toddbaert added `TestEventHandler_StateUpdatedBeforeHandlersRun` with three subtests: API handler sees ReadyState on default ProviderReady, client handler sees ErrorState on named ProviderError, and stale→ready handler sees ReadyState (not the prior StaleState). I confirmed all three fail against the prior ordering (state written after handlers fired) and pass with this PR's fix.

openfeature/event_executor_test.go
```diff
@@ -1557,3 +1557,123 @@ func TestEventHandler_ChannelClosure(t *testing.T) {
...
+	t.Run("API handler observes new state for default provider", func(t *testing.T) {
+		executor := newEventExecutor()
+
+		provider := struct {
+			FeatureProvider
+			EventHandler
+		}{
+			NoopProvider{},
+			&ProviderEventing{c: make(chan Event, 1)},
+		}
```
erka:
```suggestion
		provider := NewMockFeatureProvider(ctrl)
```
Could we use mock providers in these tests? The current providers work, but their setup adds extra logic noise for no reason.

openfeature/event_executor.go
erka: do we really want to preallocate here? I think 99% of people will have only default provider. `var namedDomains []string` could be good here too as we hope each provider will be set only once.

erka (commented):
thank you @SAY-5 for working on this. it looks good.

openfeature/event_executor.go
sahidvelji: Agreed, `var namedDomains []string` also works. Just thought the initialization with 0 capacity was a bit odd.

openfeature/event_executor.go
```diff
@@ -297,29 +297,40 @@ func (e *eventExecutor) triggerEvent(event Event, handler FeatureProvider) {
 	e.mu.Lock()
 	defer e.mu.Unlock()
 
-	// first run API handlers
+	// Per OpenFeature spec requirement 5.3.5, the client's provider
+	// status MUST be updated BEFORE any event handlers (API-level or
+	// client-scoped) fire, so every handler — including API-level
+	// handlers that query client state — observes the new status.
+	// Write the state for every domain affected by this handler's
+	// provider reference first, then dispatch handlers.
```
toddbaert: This comment is a bit too much IMO, let's keep it to 1-2 lines.

toddbaert (approved):
LGTM - will wait for a review from @erka or @sahidvelji as well.

erka (approved):
Thanks for contributing @SAY-5

PR #472, opened 2026-02-11
fix: goroutine leaks in event executor and improve shutdown handling

beeme1mr:
## This PR

- properly shuts down the event executor

### Related Issues

Fixes #471

### Notes

Added go.uber.org/goleak to try and detect other leaks.

### How to test

Run test suite

PR #508, opened 2026-06-08
fix: prevent goroutine leaks in event executor and improve shutdown handling

### LINKED ISSUE #471: [BUG]  Unexpected Goroutine Leak in OpenFeature Go SDK  in Test Environment.
### Brief Description

The event executor initiates a [background goroutine](https://github.com/open-feature/go-sdk/blob/main/openfeature/event_executor.go#L275-L283) during package initialization to manage FeatureProvider events, which remains active as no mechanism exists to signal its completion during testing..

### Observed Behavior
A goroutine remains active after test completion:
- Goroutine state: chan receive
- Test suites using goleak.VerifyTestMain(m) fail during goroutine leak verification.

### Stack trace

``` 
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 21 in state chan receive, with github.com/open-feature/go-sdk/openfeature.(*eventExecutor).startEventListener.func1.1 on top of the stack:
github.com/open-feature/go-sdk/openfeature.(*eventExecutor).startEventListener.func1.1()
	/Users/denis/go/pkg/mod/github.com/open-feature/go-sdk@v1.17.0/openfeature/event_executor.go:278 +0xb0
created by github.com/open-feature/go-sdk/openfeature.(*eventExecutor).startEventListener.func1 in goroutine 1
	/Users/denis/go/pkg/mod/github.com/open-feature/go-sdk@v1.17.0/openfeature/event_executor.go:277 +0x60
```

### Expected Behavior

All goroutines should be properly cleaned up after test execution, leaving no lingering background processes.

### Steps to reproduce

You can find the complete reproducible example in this GitHub Gist: [[goleak-openfeature-poc](https://gist.github.com/grambbledook/b8f0f7083a1ad6a28214380f315c01a8)]

JamieSinn:
## This PR
- properly shuts down the event executor
- Fixes internal test goroutines leaking + consumer threads leaking too in downstream tests. (#471 )

### Related Issues
- Fixes #471 
### Notes
This is an updated version of #472 because i lack perms to push to that branch :P 

### How to test
CI/Unit testing handles this.

### DIFF
--- a/openfeature/context_aware_test.go
+++ b/openfeature/context_aware_test.go
@@ -302,7 +302,7 @@ func TestGlobalContextAwareShutdown(t *testing.T) {
 		defer exec.shutdown()
 		// ShutdownWithContext below reinitializes the global event executor via
 		// initSingleton; shut that replacement down too so it doesn't leak.
-		defer func() { eventing.(*eventExecutor).shutdown() }()
+		defer func() { eventing.shutdown() }()
 		testAPI := newEvaluationAPI(exec)
 		api = testAPI
 		eventing = exec
@@ -342,7 +342,7 @@ func TestGlobalContextAwareShutdown(t *testing.T) {
 		defer exec.shutdown()
 		// ShutdownWithContext below reinitializes the global event executor via
 		// initSingleton; shut that replacement down too so it doesn't leak.
-		defer func() { eventing.(*eventExecutor).shutdown() }()
+		defer func() { eventing.shutdown() }()
 		testAPI := newEvaluationAPI(exec)
 		api = testAPI
 		eventing = exec
@@ -388,7 +388,7 @@ func TestGlobalContextAwareShutdown(t *testing.T) {
 		defer exec.shutdown()
 		// ShutdownWithContext below reinitializes the global event executor via
 		// initSingleton; shut that replacement down too so it doesn't leak.
-		defer func() { eventing.(*eventExecutor).shutdown() }()
+		defer func() { eventing.shutdown() }()
 		testAPI := newEvaluationAPI(exec)
 		api = testAPI
 		eventing = exec
--- a/openfeature/context_aware_test.go
+++ b/openfeature/context_aware_test.go
@@ -97,6 +97,7 @@ func TestContextAwareInitialization(t *testing.T) {
 
 	// Create fresh API for isolated testing
 	exec := newEventExecutor()
+	defer exec.shutdown()
 	testAPI := newEvaluationAPI(exec)
 	api = testAPI
 	eventing = exec
@@ -235,6 +236,7 @@ func TestContextAwareShutdown(t *testing.T) {
 
 	// Create fresh API for isolated testing
 	exec := newEventExecutor()
+	defer exec.shutdown()
 	testAPI := newEvaluationAPI(exec)
 	api = testAPI
 	eventing = exec
@@ -297,6 +299,10 @@ func TestGlobalContextAwareShutdown(t *testing.T) {
 	t.Run("shutdown with context affects all providers", func(t *testing.T) {
 		// Create fresh API for isolated testing
 		exec := newEventExecutor()
+		defer exec.shutdown()
+		// ShutdownWithContext below reinitializes the global event executor via
+		// initSingleton; shut that replacement down too so it doesn't leak.
+		defer func() { eventing.(*eventExecutor).shutdown() }()
 		testAPI := newEvaluationAPI(exec)
 		api = testAPI
 		eventing = exec
@@ -333,6 +339,10 @@ func TestGlobalContextAwareShutdown(t *testing.T) {
 	t.Run("shutdown timeout handling", func(t *testing.T) {
 		// Create fresh API for isolated testing
 		exec := newEventExecutor()
+		defer exec.shutdown()
+		// ShutdownWithContext below reinitializes the global event executor via
+		// initSingleton; shut that replacement down too so it doesn't leak.
+		defer func() { eventing.(*eventExecutor).shutdown() }()
 		testAPI := newEvaluationAPI(exec)
 		api = testAPI
 		eventing = exec
@@ -375,6 +385,10 @@ func TestGlobalContextAwareShutdown(t *testing.T) {
 	t.Run("backward compatibility with regular providers", func(t *testing.T) {
 		// Create fresh API for isolated testing
 		exec := newEventExecutor()
+		defer exec.shutdown()
+		// ShutdownWithContext below reinitializes the global event executor via
+		// initSingleton; shut that replacement down too so it doesn't leak.
+		defer func() { eventing.(*eventExecutor).shutdown() }()
 		testAPI := newEvaluationAPI(exec)
 		api = testAPI
 		eventing = exec
@@ -494,6 +508,7 @@ func TestContextPropagationFixes(t *testing.T) {
 
 	// Create fresh API for isolated testing
 	exec := newEventExecutor()
+	defer exec.shutdown()
 	testAPI := newEvaluationAPI(exec)
 	api = testAPI
 	eventing = exec
@@ -542,6 +557,7 @@ func TestContextPropagationFixes(t *testing.T) {
 	t.Run("shutdown respects context cancellation", func(t *testing.T) {
 		// Reset API
 		exec = newEventExecutor()
+		defer exec.shutdown()
 		testAPI = newEvaluationAPI(exec)
 		api = testAPI
 		eventing = exec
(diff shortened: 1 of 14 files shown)


openfeature/event_executor.go
```diff
@@ -221,23 +225,31 @@ func (e *eventExecutor) startListeningAndShutdownOld(newProvider providerReferen
...
+						}:
+						case <-newProvider.shutdownSemaphore:
+							return
 						}
 					case <-newProvider.shutdownSemaphore:
 						return
 					}
 				}
-			}()
+			})
```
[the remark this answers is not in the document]
JamieSinn: https://pkg.go.dev/sync#WaitGroup.Go

openfeature/event_executor.go
```diff
@@ -281,14 +293,22 @@ func (e *eventExecutor) startListeningAndShutdownOld(newProvider providerReferen
...
+					if !ok {
+						return
+					}
+					e.triggerEvent(payload.event, payload.handler)
+				case <-e.done:
+					return
+				}
 			}
-		}()
+		})
```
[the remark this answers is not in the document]
JamieSinn: https://pkg.go.dev/sync#WaitGroup.Go

erka:
I believe there is a misunderstanding of what needs to be fixed. The person is reporting that their tests have a goroutine leak because of openfeature/go-sdk. People could not call `shutdownEventing`, and it’s hard to know that it needs to be called.

JamieSinn:
> I believe there is a misunderstanding of what needs to be fixed. The person is reporting that their tests have a goroutine leak because of openfeature/go-sdk. People could not call `shutdownEventing`, and it’s hard to know that it needs to be called.

The internal test fixes are actually a side effect of me playing whackamole trying to chase down the issue as we had no coverage for `-race` in tests - so trying to nail it down was annoying to test.

The fix for the issue is actually here: https://github.com/open-feature/go-sdk/pull/508/changes#diff-0e992ea209b1c8b10d26d7f7177238e743f4d2cf88fbc265c35b58151d497c11R230 

Previously the event executor was started on the creation of the object; and never properly cleaned up (hence the test error from the issue). 

This is at first glance definitely a bit confusing to see where the actual fix is; but its actually the lazy loading based on provider creation that fixes this issue.

openfeature/openfeature.go
erka, line 12: should we change eventing type from `eventingImpl` to `*eventExecutor` (we probably have to use `atomic.Pointer[eventExecutor]` but not in this PR)?

we could keep that interface with `var _ eventingImpl = (*eventExecutor)(nil)` for now.

openfeature/openfeature.go
```diff
@@ -149,6 +154,7 @@ func RemoveHandler(eventType EventType, callback EventCallback) {
 // hooks, event handlers, and providers.
 func Shutdown() {
 	api.Shutdown()
+	eventing.(*eventExecutor).shutdown()
```
erka:
```suggestion
```

`initSingleton` does it too. Any reason do we need this?

openfeature/openfeature.go
```diff
@@ -159,6 +165,7 @@ func Shutdown() {
 // Returns an error if any provider shutdown fails or if context is cancelled during shutdown.
 func ShutdownWithContext(ctx context.Context) error {
 	err := api.ShutdownWithContext(ctx)
+	eventing.(*eventExecutor).shutdown()
```
erka:
```suggestion
```

`initSingleton` does it too. Any reason do we need this?

erka (approved):
Thank you @JamieSinn. I've missed that openfeature.Shutdown has got the new call.
Some nits from me.

openfeature/openfeature.go
JamieSinn: Not a bad idea, I don't know the downstream impact and would have to investigate but it would make the hard casting a lot less painful/fugly.

openfeature/openfeature.go
JamieSinn: Not a need, just more an explicit step to call shutdown and not assume that the initialization of `initSingleton` will shut down things given the name of initialization.

openfeature/openfeature.go
JamieSinn: same reply as before - Not a need, just more an explicit step to call shutdown and not assume that the initialization of initSingleton will shut down things given the name of initialization.

openfeature/openfeature.go
erka: They are all internal and should not be available to others. What do you think @sahidvelji ?

JamieSinn: I think a cleaner option here would be to require the implementation of a shutdown function in some format in the interface itself; allowing for more go-esque design here. 

Definitely not this PR - but happy to look into that

JamieSinn: IMO Much cleaner - https://github.com/JamieSinn/go-sdk/compare/fix-event-goroutine-leak...JamieSinn:go-sdk:eventing-shutdown?expand=1

You could argue that it could fit in here pretty well too.

erka: It looks better. Let's apply it to this branch.

JamieSinn: Done

openfeature/event_executor_test.go
```diff
@@ -1657,3 +1688,81 @@ func TestEventHandler_StateUpdatedBeforeHandlersRun(t *testing.T) {
 		}
 	})
 }
+
+// TestBasicShutdown verifies that the shutdown method stops goroutines
+func TestBasicShutdown(t *testing.T) {
+	// Create a new event executor
+	exec := newEventExecutor()
+
```
toddbaert, line 1695:
```suggestion
	exec := newEventExecutor()
	exec.startEventListener()

```

I think you need this to make this test meaningful - without it no routine starts, I think?

openfeature/openfeature.go
```diff
@@ -18,6 +18,11 @@ func init() {
 }
 
 func initSingleton() {
```
toddbaert:
```suggestion
func resetSingleton() {
```

We aren't initing, we are also clearing. I think this name helps with that.

openfeature/openfeature.go
```diff
@@ -159,6 +165,7 @@ func Shutdown() {
 // Returns an error if any provider shutdown fails or if context is cancelled during shutdown.
 func ShutdownWithContext(ctx context.Context) error {
 	err := api.ShutdownWithContext(ctx)
+	eventing.shutdown()
```
toddbaert:
```suggestion
```

This is done in `initSingleton` already, but see: https://github.com/open-feature/go-sdk/pull/508/changes#r3389633438

openfeature/context_aware_test.go
```diff
@@ -297,6 +299,10 @@ func TestGlobalContextAwareShutdown(t *testing.T) {
 	t.Run("shutdown with context affects all providers", func(t *testing.T) {
 		// Create fresh API for isolated testing
 		exec := newEventExecutor()
+		defer exec.shutdown()
+		// ShutdownWithContext below reinitializes the global event executor via
+		// initSingleton; shut that replacement down too so it doesn't leak.
+		defer func() { eventing.shutdown() }()
 		testAPI := newEvaluationAPI(exec)
```
toddbaert: Nit: There's a lot of duplication here - not a blocker but you could extract this setup/teardown into a function, maybe that uses t.Cleaup.

openfeature/event_executor.go
```diff
@@ -221,23 +225,31 @@ func (e *eventExecutor) startListeningAndShutdownOld(newProvider providerReferen
 		e.activeSubscriptions = append(e.activeSubscriptions, newProvider)
 
 		if v, ok := newProvider.featureProvider.(EventHandler); ok {
-			go func() {
+			// Ensure the event listener is running now that an event-emitting
+			// provider produces events on eventChan (started once, lazily).
+			e.startEventListener()
```
toddbaert, line 230: This move makes sense. I think the old naming of this method was misleading until this change - it wasn't actually setting up all activity required for "listening".

toddbaert (approved):
Approved, but consider my nits please! Thanks!

openfeature/event_executor_test.go
JamieSinn: This was intended to basically prove that this fixed the issue in the first place; where it did not auto-start, so this is intended to not start it.

openfeature/openfeature.go
JamieSinn: yeah i like this a lot better

openfeature/context_aware_test.go
JamieSinn: done - good callout.

openfeature/openfeature.go
JamieSinn: cleaned up the above comments where it was called out by @erka too since the rename makes more sense now.

openfeature/event_executor_test.go
toddbaert: Understand, thanks.

toddbaert:
I moved `TestNoGoroutineLeakWithMultipleProviders` because it was in events_executor.go, but really was a higher level test that only touched `openfeature` - so I moved it to `openfeature_test.go`; it's basically an integration test.

Will merge.

PR #503, opened 2026-06-08
chore: delete deprecated pkg package

erka:
## This PR

- The pkg package has been deprecated long enough, and it's time to drop it.

PR #509, opened 2026-06-09
chore: delete deprecated pkg package

toddbaert:
Replay of https://github.com/open-feature/go-sdk/pull/503


erka (approved):

beeme1mr (approved):
How are we going to communicate this? Shouldn't this be something we release in v2?

Issue #443, opened 2025-10-24
Be consistent in blocking provider API function (async param vs `...AndWait` name)


toddbaert:
We should improve the consistency of the function for setting a provider and blocking. In some cases we have an `async` param, and in others we use an `...AndWait` function name. 

## Background:

> I dislike that we mix `...AndWait` and `aysnc bool` in our public API - it's inconsistent - but this isn't something you introduced, it already existed elsewhere. I think in v2 we should go for one or the other (the spec doesn't dictate function naming so we have flexibility here).  

 _Originally posted by @toddbaert in [#442](https://github.com/open-feature/go-sdk/pull/442/files#r2461023401)_

Issue #409, opened 2025-07-17
Consider breaking up primary interfaces, or not exporting them


toddbaert:
> Remove (or unexport) certain identifiers (types, functions, methods, interfaces) that are unnecessary. For example, [openfeature.IEvaluation](https://pkg.go.dev/github.com/open-feature/go-sdk/openfeature#IEvaluation) or [openfeature.IClient](https://pkg.go.dev/github.com/open-feature/go-sdk/openfeature#IClient) should most definitely not make it into v2. We will want to be especially careful about which interfaces we are exporting. Those that are exported must be named appropriately (not like IClient or ifaceClient or ClientIface).

We need to discuss this further, but we should rethink how we expose these interfaces.

jblacker:
I think that `IEvaluation` can be removed completely or at least unexposed. I think instead a client "builder" should be the primary way to set up getting an implementation of `IClient`. If we don't go the `builder` route it might be more idiomatic to have an `initialization` function that implements the option pattern.

On the topic of `IClient` I also don't think needs to be exported, but instead parts of it should be put into smaller interfaces that can compose an internal `IClient`, with parts of the smaller interfaces exposed. Won't necessarily be able to remove many of the methods until v2, but it could be a good exercise in moving in the right direction.

```go
iClient interface { // internal (need a better name)
  Metadata() ClientMetadata
  SetEvaluationContext(evalCtx EvaluationContext) // May not be needed
  EvaluationContext() EvaluationContext    // May not be needed
  Evaluator
  detailEvaluator
}

FlagType interface { // exported
  int64 | float64 | string | bool | comparable
}

Client interface { //exported
  Evaluator
  DetailEvaluator
}

Evaluator interface { // exported
  Boolean(ctx context.Context, flag string, defaultValue bool, evalCtx EvaluationContext, options ...Option) bool
  String(ctx context.Context, flag string, defaultValue string, evalCtx EvaluationContext, options ...Option) string
  Float(ctx context.Context, flag string, defaultValue float64, evalCtx EvaluationContext, options ...Option) float64
  Int(ctx context.Context, flag string, defaultValue int64, evalCtx EvaluationContext, options ...Option) int64
  Object[T FlagType](ctx context.Context, flag string, defaultValue any, evalCtx EvaluationContext, options ...Option) T
}

DetailEvaluator interface { // exported
[...]

Issue #401, opened 2025-07-04
Tracking issue for V2


sahidvelji:
In issues #389 and #158, the need to create a V2 of the go-sdk was identified. This topic was also discussed in the biweekly community meeting on Thursday July 3, 2025.

There are several changes that I would like to see as part of a V2 (not exhaustive):

- [ ] We should also consider #350. In v2, I would like to change the signature of `openfeature.NewClient` to take no parameters, and create a new function called `openfeature.NewDomainClient` that takes a string parameter called `domain`. I want to avoid the names `openfeature.GetClient` and `openfeature.GetDomainClient` as such `Get` functions are not as common in Go and `New` is more idiomatic.

Out of the list above, only number 8 and potentially number 1 (depending on how/where we introduce generics) will cause a breaking change to providers. For number 8, taking a context parameter and returning an error is small enough of a change that I think that is reasonable, especially considering that the StateHandler interface is opt-in and not required to implement.

To implement the above, I think we can actually make a lot of progress before we even create a v2 module. For example, for those identifiers that are to be renamed, we can create the target type/function and deprecate the old one. Then once v2 is created, all we need to do is delete the deprecated identifiers, making that transition a bit easier on our end.

Lastly, before we officially release V2 and deprecate V1, we will want to gather as much feedback as possible about the SDK to ensure that no further breaking changes are anticipated after V2 is released. Once it is released, we will also want to be careful going forward about adding new exported identifiers to ensure that they align with language best practices.

dmathieu:
This API concern had breaking changes but could be considered for a v2.
https://github.com/open-feature/go-sdk/issues/259

erka:
> This API concern had breaking changes but could be considered for a v2. [#259](https://github.com/open-feature/go-sdk/issues/259)

@dmathieu if you don't want to handle errors there is [client.Boolean](https://pkg.go.dev/github.com/open-feature/go-sdk@v1.15.1/openfeature#Client.Boolean).

toddbaert:
My notes, addressing each of @sahidvelji 's points:

[...]

PR #511, opened 2026-06-13
feat: export EvaluationAPI and simplify provider/client its surface

erka:
## This PR

- Rename unexported evaluationAPI into exported EvaluationAPI
- Replace evaluationImpl interface with providerBindingFn function type
- Deprecated IEvaluation interface
- Remove deprecated `GetApiInstance() and SetLogger()
- Consolidate provider methods: SetProvider(ctx, provider, opts...) with
  WithDomain() replacing SetNamedProvider*and SetProviderWithContext*
- Consolidate client creation: NewClient(opts...) replacing GetClient +
  GetNamedClient
- Add tests for new API surface and isolated instance example


sahidvelji:
Looks good overall. Found a few instances where doc comments could be improved.

## Real exported → unexported leak

### `APIOption` signature exposes unexported `apiOptions`

`openfeature/openfeature_api.go:89-90`

```go
// APIOption configures API-level operations such as domain selection.
type APIOption func(*apiOptions)
```

godoc renders the type signature verbatim,
so consumers see an exported type whose underlying signature names a type
(`apiOptions`) they cannot reference.

**Suggested fix** — use the interface-with-unexported-method pattern
(`grpc.DialOption`, `slog.HandlerOption`, etc.):

```go
type APIOption interface {
    apply(*apiOptions)
}

type apiOptionFunc func(*apiOptions)
func (f apiOptionFunc) apply(o *apiOptions) { f(o) }

func WithDomain(domain string) APIOption {
    return apiOptionFunc(func(o *apiOptions) { o.domain = domain })
}
```

Alternative: export the options struct as `APIOptions` with unexported fields.

## Stale doc text from the rename

Inverse of the rule (unexported function with a doc comment using the *old
exported* name). Not godoc-visible to consumers,
but clearly leftover from the PR's own renames.

- `openfeature/openfeature_api.go:296` —
  `// GetProviderMetadata returns the default FeatureProvider's metadata`
  sits above `func (a *EvaluationAPI) getProviderMetadata()`.
- `openfeature/openfeature_api.go:304` —
  `// getDomainProviderMetadata returns the default FeatureProvider's metadata`
  is wrong: the function returns the *domain-scoped* provider's metadata,
  falling back to the default only when the domain isn't registered.
- `openfeature/openfeature_api.go:317` —
  `// getDomainProviders returns named providers map.`
  uses the old "named" terminology after the rename to "domain".
- `openfeature/openfeature.go:10` —
  `// api is the global evaluationImpl implementation.`
  references `evaluationImpl`, which this PR deleted.

## Missing or inaccurate doc on exported items

### `(*EvaluationAPI).AddHooks` has no doc comment

`openfeature/openfeature_api.go:350`

```go
func (a *EvaluationAPI) AddHooks(hooks ...Hook) {
```

Trivial to add: `// AddHooks appends to the API-level hook collection.`

### `(*EvaluationAPI).NewClient` doc claims it returns `IClient`

`openfeature/openfeature_api.go:325-326`

```go
// NewClient returns a IClient bound to the default provider, or to a named
// provider if WithDomain is provided.
func (a *EvaluationAPI) NewClient(opts ...APIOption) *Client {
```

The function returns `*Client`, not `IClient`.
Also "a IClient" should be "a [Client]".

**Suggested rewrite:**

```go
// NewClient returns a [Client] bound to the default provider,
// or to a domain-scoped provider when [WithDomain] is supplied.
```

### Package-level `NewClient` doc references a nonexistent `Name` parameter

`openfeature/client.go:52`

```go
// NewClient returns a new Client. Name is a unique identifier for this client.
func NewClient(domain string) *Client {
```

The parameter is `domain`, not `Name`.
Pre-existing wording, but the PR touched the function,
so it's natural to fix here.

**Suggested rewrite:**

```go
// NewClient returns a new [Client] for the given domain.
```

### `factory.CurrentAPI` references a nonexistent `isolated.CurrentAPI`

`openfeature/internal/factory/factory.go:11-13`

```go
// CurrentAPI is set by openfeature.init and read by openfeature/isolated.CurrentAPI.
// Returns the current singleton evaluation API as any to avoid an import cycle;
// callers must type-assert to [*openfeature.EvaluationAPI].
var CurrentAPI func() any
```

No `CurrentAPI` is exported from `openfeature/isolated` —
only `openfeature/isolated_test` uses `factory.CurrentAPI`.
The sibling `NewAPI` doc on line 6 is accurate
because `isolated.NewAPI` does exist.

**Suggested rewrite:**

```go
// CurrentAPI is set by openfeature.init and read by openfeature/isolated tests.
```

## Summary

Only one item is a true exported→unexported leak
(`APIOption func(*apiOptions)` — godoc-visible).
The rest are stale-rename artifacts and minor inaccuracies
introduced or surfaced by the refactor.
All are mechanical fixes that fit naturally in a single follow-up commit
on this PR.

openfeature/openfeature_api.go
```diff
@@ -212,7 +217,11 @@ func (a *evaluationAPI) setProviderWithContext(ctx context.Context, provider Fea
 	return a.initNew(ctx, "", provider), nil
 }
 
-func (a *evaluationAPI) setNamedProviderWithContext(ctx context.Context, clientName string, provider FeatureProvider) (<-chan error, error) {
+func (a *EvaluationAPI) setDomainProvider(ctx context.Context, clientName string, provider FeatureProvider) (<-chan error, error) {
```
sahidvelji:
```suggestion
func (a *EvaluationAPI) setDomainProvider(ctx context.Context, domain string, provider FeatureProvider) (<-chan error, error) {
```

openfeature/openfeature_api.go
```diff
@@ -285,89 +294,87 @@ func (a *evaluationAPI) shutdownOld(ctx context.Context, oldProvider FeatureProv
...
+func (a *EvaluationAPI) NewClient(opts ...APIOption) *Client {
+	o := apiOptions{}
+	for _, opt := range opts {
+		opt(&o)
+	}
+	return &Client{
+		domain:            o.domain,
+		providerBinding:   a.resolveBinding,
+		clientEventing:    a.eventExecutor,
+		metadata:          ClientMetadata{domain: o.domain}, //nolint:staticcheck
```
sahidvelji:
```suggestion
		metadata:          NewClientMetadata(o.domain)
```

openfeature/openfeature_api.go
```diff
@@ -285,89 +294,87 @@ func (a *evaluationAPI) shutdownOld(ctx context.Context, oldProvider FeatureProv
...
-func (a *evaluationAPI) GetNamedProviderMetadata(name string) Metadata {
+// getDomainProviderMetadata returns the default FeatureProvider's metadata
+func (a *EvaluationAPI) getDomainProviderMetadata(name string) Metadata {
 	a.mu.RLock()
 	defer a.mu.RUnlock()
 
-	provider, ok := a.namedProviders[name]
+	provider, ok := a.domainProviders[name]
 	if !ok {
 		return ProviderMetadata()
```
sahidvelji:
```suggestion
		return a.defaultProvider.Metadata()
```
Not part of the diff of this PR, but worth fixing here

openfeature/isolated_api_test.go
```diff
@@ -2,34 +2,32 @@ package openfeature
...
+		t.Fatal("newAPI() returned nil")
 	}
 }
 
-// Requirement 1.8.2: Isolated instances MUST conform to the same API contract as the singleton (IEvaluation).
+// Requirement 1.8.2: Isolated instances MUST conform to the same API contract as the singleton.
 func TestRequirement_1_8_2(t *testing.T) {
-	// compile-time check: *EvaluationAPI must satisfy IEvaluation
-	var _ IEvaluation = newAPIForTest()
+	a := newAPI()
```
sahidvelji:
```suggestion
  if _, ok := factory.NewAPI().(*EvaluationAPI); !ok {
      t.Error("factory.NewAPI() did not return *EvaluationAPI")
  }
```
I suggest either applying the modernize fix or rewrite like above.

erka:
@coderabbitai review

erka:
@coderabbitai full review

openfeature/openfeature_api.go
```diff
@@ -86,90 +86,99 @@ func unbindProvider(provider FeatureProvider, apiInst *evaluationAPI) {
...
-	return a.SetProviderWithContextAndWait(context.Background(), provider)
-}
-
-// SetProviderWithContext sets the default FeatureProvider with context-aware initialization.
-func (a *evaluationAPI) SetProviderWithContext(ctx context.Context, provider FeatureProvider) error {
-	if provider == nil {
-		return errors.New("default provider cannot be set to nil")
+// SetProvider sets a FeatureProvider with context-aware initialization.
+// If WithDomain is provided, the provider is bound to the given domain.
+func (a *EvaluationAPI) SetProvider(ctx context.Context, provider FeatureProvider, opts ...APIOption) error {
```
sahidvelji, line 134: api.SetProvider(ctx, p, WithDomain(cfg.Domain)) where cfg.Domain is empty silently registers p as the default provider.

  The godoc on WithDomain doesn't say. Pick one and document:
  - "An empty domain is treated as no domain (default provider scope)" — explicit foot-gun warning.
  - Or reject empty in the option (return an error sentinel via the apply func).

erka: The option 1 is better for `NewClient`

sahidvelji (approved):

openfeature/openfeature_api.go
```diff
@@ -86,90 +86,99 @@ func unbindProvider(provider FeatureProvider, apiInst *evaluationAPI) {
...
+}
+
+type (
+	apiOptionFunc func(*apiOptions)
+	apiOptions    struct {
+		domain string
+	}
+)
+
+func (f apiOptionFunc) apply(o *apiOptions) { f(o) }
```
toddbaert, line 101: Initially I thought this was overbuilt, but I can see this way gives us control over `apiOptions` and will allow us to hide internal changes from customers. Nice.

openfeature/openfeature.go
```diff
@@ -30,44 +31,40 @@ func resetSingleton() {
...
-	apiRef := api()
-	return newClient("", apiRef, apiRef.eventExecutor)
+	return api().NewClient()
 }
 
 // SetProvider sets the default [FeatureProvider]. Provider initialization is asynchronous and status can be checked from
 // provider status
 func SetProvider(provider FeatureProvider) error {
-	return api().SetProvider(provider)
+	return api().SetProvider(context.Background(), provider)
```
toddbaert, line 61: The v2 tracking issue #401 and #414 / #443 / #409 cover the bigger picture, but I don't see an open issue about the package-level setters (`SetProvider`, `SetNamedProvider`, etc.) getting the same options-shape change, which would let singleton users reach `WithDomain` (and any future option) without going to `isolated.NewAPI()`. Is that intentional (keep wrappers as legacy forever) or worth a v2 sub-issue for removal?

I'm guessing that's the intent, right?

toddbaert (approved):
I really like the shape. This feels better.

Please see: https://github.com/open-feature/go-sdk/pull/511/changes#r3466966882

Also, should we add README additions about the new isolated stuff and the associated new API here or in a separate PR? WDYT @erka @sahidvelji ?

openfeature/interfaces.go
```diff
@@ -2,11 +2,11 @@ package openfeature
 
 import (
 	"context"
-
-	"github.com/go-logr/logr"
 )
 
 // IEvaluation defines the OpenFeature API contract
+//
+// Deprecated: IEvaluation will be removed in v1.20. Use [EvaluationAPI] instead.
```
toddbaert:
```suggestion
// Deprecated: IEvaluation will be removed in later versions. Use [EvaluationAPI] instead.
```

If you have some specific idea of what will bring us to v1.20 in the near future, feel free to keep this, but if not I would just keep it a bit vague.

openfeature/openfeature.go
erka: I would like to align these methods with the other SDKs (such as JavaScript and .NET), which do not have SetDomain.... 

JS as an example
```js
// Registering the default provider
OpenFeature.setProvider(new TypedInMemoryProvider(myFlags));
// Registering a provider to a domain
OpenFeature.setProvider('my-domain', new TypedInMemoryProvider(someOtherFlags));
```

I also don't like `...WithContext` methods, so it's a trade-off in many ways with golang.

The v2 issues are outdated in some ways, and they would likely need to be revisited and rethought if v2 remains on the roadmap. I don’t see any community interest in v2 right now.

erka:
`openfeature/isolated/isolated_example_test.go` is an example of usage and will be included in the godoc.

image

I believe none of the SDKs mention the isolated API in their README.

PR #522, opened 2026-07-07
fix: guard against nil ContextEvaluator func in memprovider Resolve

### LINKED ISSUE #521: [BUG]  memprovider.ContextEvaluator nil-check is insufficient — non-nil pointer to nil function panics
### Observed behavior

`ContextEvaluator` is typed as `*func(...)` (pointer-to-function) at in_memory_provider.go. The `Resolve` method checks if `flag.ContextEvaluator != nil` before calling the function, but this only verifies the pointer is non-nil, not the underlying function value.
A user can create an `InMemoryFlag` with a non-nil pointer to a nil function, which passes the guard and causes a nil-pointer dereference panic at the call.

### Expected Behavior

_No response_

### Steps to reproduce

```go
package main

import (
    "github.com/open-feature/go-sdk/openfeature/memprovider"
)

func main() {
    var evalFn func(this memprovider.InMemoryFlag, flatCtx openfeature.FlattenedContext) (any, openfeature.ProviderResolutionDetail)

    flag := memprovider.InMemoryFlag{
        Key:              "flag2",
        DefaultVariant:   "on",
        Variants:         map[string]any{"on": true},
        ContextEvaluator: &evalFn, // non-nil pointer, nil function
    }

    flag.Resolve(true, openfeature.FlattenedContext{})
    // panic: nil pointer 
}
```

somaz94:
`memprovider.InMemoryFlag.ContextEvaluator` is typed `*func(...)`. `Resolve`
only checked the pointer for nil, so a non-nil pointer to a nil func passed the
guard and panicked on the dereference-and-call. This adds the missing inner
nil-func check so such a flag falls back to the default variant, plus a
regression test.

Fixes #521

Validation:
- `make test` (go test -tags testtools -race): all packages pass, memprovider at 86.9% coverage
- `make lint` (golangci-lint v2.9.0): 0 issues
- Confirmed the new test panics against the old guard and passes with the fix

### DIFF
--- a/openfeature/memprovider/in_memory_provider_test.go
+++ b/openfeature/memprovider/in_memory_provider_test.go
@@ -246,19 +246,17 @@ func TestInMemoryProvider_NilContextEvaluatorFunc(t *testing.T) {
 
 	ctx := t.Context()
 
-	t.Run("nil evaluator func falls back to default variant without panic", func(t *testing.T) {
-		evaluation := memoryProvider.BooleanEvaluation(ctx, "nilFnFlag", false, nil)
+	evaluation := memoryProvider.BooleanEvaluation(ctx, "nilFnFlag", false, nil)
 
-		if evaluation.Value != true {
-			t.Errorf("incorrect evaluation, expected %v, got %v", true, evaluation.Value)
-		}
-		if evaluation.Variant != "true" {
-			t.Errorf("incorrect variant, expected %q, got %q", "true", evaluation.Variant)
-		}
-		if evaluation.Reason != openfeature.StaticReason {
-			t.Errorf("incorrect reason, expected %q, got %q", openfeature.StaticReason, evaluation.Reason)
-		}
-	})
+	if evaluation.Value != true {
+		t.Errorf("incorrect evaluation, expected %v, got %v", true, evaluation.Value)
+	}
+	if evaluation.Variant != "true" {
+		t.Errorf("incorrect variant, expected %q, got %q", "true", evaluation.Variant)
+	}
+	if evaluation.Reason != openfeature.StaticReason {
+		t.Errorf("incorrect reason, expected %q, got %q", openfeature.StaticReason, evaluation.Reason)
+	}
 }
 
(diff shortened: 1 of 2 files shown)


openfeature/memprovider/in_memory_provider_test.go
```diff
@@ -225,6 +225,42 @@ func TestInMemoryProvider_WithContext(t *testing.T) {
...
+		if evaluation.Value != true {
+			t.Errorf("incorrect evaluation, expected %v, got %v", true, evaluation.Value)
+		}
+		if evaluation.Variant != "true" {
+			t.Errorf("incorrect variant, expected %q, got %q", "true", evaluation.Variant)
+		}
+		if evaluation.Reason != openfeature.StaticReason {
+			t.Errorf("incorrect reason, expected %q, got %q", openfeature.StaticReason, evaluation.Reason)
+		}
+	})
```
erka:
```suggestion
	evaluation := memoryProvider.BooleanEvaluation(ctx, "nilFnFlag", false, nil)

	if evaluation.Value != true {
		t.Errorf("incorrect evaluation, expected %v, got %v", true, evaluation.Value)
	}
	if evaluation.Variant != "true" {
		t.Errorf("incorrect variant, expected %q, got %q", "true", evaluation.Variant)
	}
	if evaluation.Reason != openfeature.StaticReason {
		t.Errorf("incorrect reason, expected %q, got %q", openfeature.StaticReason, evaluation.Reason)
	}
```

erka (approved):
Thank you @somaz94 

One small nit: I don't think a subtest is necessary if there's only one test case.

openfeature/memprovider/in_memory_provider_test.go
somaz94: Good call — applied in f9fcb72. Flattened the single-case subtest into direct assertions since the extra `t.Run` wrapper added no value here. Thanks for the review!

sahidvelji (approved):

PR #523, opened 2026-07-28
fix: disallow provider registration and initialization after API shutdown

### LINKED ISSUE #495: [BUG] Provider lifecycle calls may occur in invalid order
### Observed behavior

In some situations, provider Init may be called after openfeature.Shutdown() call. Additionally, a provider that was not received `Init` call may got a `Shutdown` call.

### Expected Behavior

_No response_

### Steps to reproduce

```go
func TestProviderInitShutdown(t *testing.T) {
	p := &ttprovider{t: t}
	err := openfeature.SetProvider(p)
	if err != nil {
		t.Fatalf("expected no error, got %v", err)
	}
	openfeature.Shutdown()
	time.Sleep(time.Second)
}

var _ openfeature.StateHandler = (*ttprovider)(nil)

type ttprovider struct {
	openfeature.NoopProvider
	halted atomic.Bool
	t      testing.TB
}

func (a *ttprovider) Init(openfeature.EvaluationContext) error {
	if a.halted.Load() {
		a.t.Fatal("unexpected init call after Shutdown")
	}
	return nil
}

func (a *ttprovider) Shutdown() {
	if a.halted.CompareAndSwap(false, true) {
		a.t.Log("Shutdown was called")
	}
}
```

erka:
## This PR

- track evaluationAPIState (active/shutdown)
- guard SetProvider and SetDomainProvider to reject calls after shutdown
- wait for in-flight provider initializations before shutting down process

closes #495

### DIFF
--- a/openfeature/openfeature_api.go
+++ b/openfeature/openfeature_api.go
@@ -127,7 +127,7 @@ type EvaluationAPI struct {
 
 // newEvaluationAPI is a helper to generate an API. Used internally
 func newEvaluationAPI(eventExecutor *eventExecutor) *EvaluationAPI {
-	a := &EvaluationAPI{
+	return &EvaluationAPI{
 		defaultProvider: NoopProvider{},
 		domainProviders: map[string]FeatureProvider{},
 		hks:             []Hook{},
@@ -136,7 +136,6 @@ func newEvaluationAPI(eventExecutor *eventExecutor) *EvaluationAPI {
 		eventExecutor:   eventExecutor,
 		state:           evaluationAPIStateActive,
 	}
-	return a
 }
 
 // SetProvider sets a FeatureProvider with context-aware initialization.
--- a/openfeature/openfeature_api.go
+++ b/openfeature/openfeature_api.go
@@ -8,7 +8,6 @@ import (
 	"reflect"
 	"slices"
 	"sync"
-	"sync/atomic"
 )
 
 // providerBindingEntry holds a strong reference to the provider and the API instance it is bound to.
@@ -122,7 +121,7 @@ type EvaluationAPI struct {
 	evalCtx         EvaluationContext
 	eventExecutor   *eventExecutor
 	mu              sync.RWMutex
-	state           atomic.Uint32
+	state           uint32
 	initWg          sync.WaitGroup
 }
 
@@ -135,9 +134,8 @@ func newEvaluationAPI(eventExecutor *eventExecutor) *EvaluationAPI {
 		evalCtx:         EvaluationContext{},
 		mu:              sync.RWMutex{},
 		eventExecutor:   eventExecutor,
-		state:           atomic.Uint32{},
+		state:           evaluationAPIStateActive,
 	}
-	a.state.Store(evaluationAPIStateActive)
 	return a
 }
 
@@ -217,7 +215,7 @@ func (a *EvaluationAPI) setProvider(ctx context.Context, provider FeatureProvide
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
-	if a.state.Load() == evaluationAPIStateShutdown {
+	if a.state == evaluationAPIStateShutdown {
 		return nil, errAPIShutdown
 	}
 
@@ -248,7 +246,7 @@ func (a *EvaluationAPI) setDomainProvider(ctx context.Context, domain string, pr
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
-	if a.state.Load() == evaluationAPIStateShutdown {
+	if a.state == evaluationAPIStateShutdown {
 		return nil, errAPIShutdown
 	}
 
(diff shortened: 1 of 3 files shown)


openfeature/openfeature_api.go
```diff
@@ -385,6 +404,28 @@ func (a *EvaluationAPI) RemoveHandler(eventType EventType, callback EventCallbac
...
+	if a.state == evaluationAPIStateShutdown {
+		return nil
+	}
+	a.state = evaluationAPIStateShutdown
+
+	// Wait for in-flight provider initializations to finish.
+	// sync.WaitGroup requires that if reused, all new Add calls happen
+	// after all previous Wait calls have returned. This ensures we can
+	// safely register new providers if shutdown is cancelled (ctx.Done).
+	a.initWg.Wait()
```
sahidvelji, line 416: `initWg.Wait()` runs while `Shutdown` still holds the `a.mu` write lock, and the init goroutine runs the provider's `Init`/`InitWithContext`, which can be arbitrarily slow (network, retries). Two consequences:

1. **Evaluation starvation.** For the entire duration of a slow in-flight init, every reader that takes `a.mu.RLock()` blocks — including `resolveBinding` (called on every flag evaluation), `getProvider`, `getDomainProviders`, and the metadata/state lookups. A single slow `SetProvider` concurrent with `Shutdown` freezes all evaluations until init completes.
2. **Reentrant deadlock.** If a provider's `Init` calls back into `SetProvider`/`Shutdown` on this same API instance, that call blocks on `a.mu.Lock()` (held by `Shutdown`), which can only be released after `Wait()` returns — but `Wait()` can't return until that `Init` (and its deferred `initWg.Done()`) completes. Exotic, but newly deadlock-prone.

Consider snapshotting the shutdown decision under the lock, then releasing `a.mu` before `initWg.Wait()` (or making the wait context-aware).

openfeature/openfeature_api.go
```diff
@@ -385,6 +404,28 @@ func (a *EvaluationAPI) RemoveHandler(eventType EventType, callback EventCallbac
...
+
+	select {
+	case <-ctx.Done():
```
sahidvelji, line 419: `initWg.Wait()` above is unconditional and runs before this `select`, so a cancelled or deadline-bound context is only honored *after* the (potentially unbounded) init wait, not during it. A caller that passes a tight deadline hoping to bound how long `Shutdown` blocks will still wait for the full in-flight init before this branch returns `ctx.Err()`.

Also, the comment on this branch ("Wait for in-flight initializations to finish before restoring state") is inaccurate — the wait already completed above the `select`; nothing is waited on here.

sahidvelji (commented):
Automated review pass. The lifecycle gating and the `Shutdown()` teardown ordering fix look correct, and the added tests are race-clean. Two inline comments, plus one outside-diff note.

These are distinct from CodeRabbit's earlier concurrency comments: those were filed against the earlier implementation that used atomic `state.Load()`/`state.Store()` and a detached Wait goroutine. In the current code, `state` and all `initWg` operations are accessed under the pre-existing `a.mu` write lock (state checks in `setProvider`/`setDomainProvider` and the transitions + `initWg.Wait()` in `Shutdown` all run inside `a.mu.Lock()` sections; `initWg.Add(1)` in `initNew` is only reached from those locked paths). That serialization closes the observe-then-transition and Add-during-Wait races CodeRabbit described, so those comments no longer apply.

**Outside-diff note — the provider-replacement path still has issue #495's ordering bug (`openfeature_api.go:310-319`, `shutdownOld`).** The fix covers the API `Shutdown()` teardown path via `initWg.Wait()`, but replacing a provider before its `Init` finishes is unaffected: `SetProvider(A)` then `SetProvider(B)` triggers `shutdownOld(ctx, A)`, which spawns a goroutine calling `A.Shutdown()`/`ShutdownWithContext` immediately, with no coordination with A's still-in-flight `Init` goroutine. So A can receive `Shutdown` before or during its `Init` — the exact class of invalid ordering #495 describes. Worth confirming whether #495 should be considered fully closed, or whether the same await-in-flight-init coordination should extend to `shutdownOld`.

openfeature/openfeature_api.go
erka: The godocs for sync.WaitGroup says

> If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned.

So I don't see a way to prevent unbounded init wait time at the moment. I would like to have suggestions how it could be improved.

openfeature/openfeature_api.go
sahidvelji: Following up on my own comment — after tracing the package-level shutdown path this is much less of a concern than I made it sound, so this can be resolved.

`ShutdownWithContext` swaps in a fresh `EvaluationAPI` *before* shutting the old one down (`resetSingletonWithContext`), so on the global path the instance parked in `initWg.Wait()` is the discarded one — live evaluations already resolve against the fresh instance, so there's no starvation. It could still block readers of a *directly-held isolated instance* that's being torn down, but that's a narrow case and the instance is terminal afterward anyway. The re-entrant-`Init` deadlock likewise needs a provider that calls back into the same API instance during its own initialization, which is an unusual pattern.

Not worth acting on — resolving.

openfeature/openfeature_api.go
sahidvelji: Not blocking — take it or leave it (or a follow-up). Given the additional complexity for such a narrow, unusual case, I've reconsidered and don't think this is worth addressing. But since you asked for suggestions:

I think the unbounded wait is really a sign that `sync.WaitGroup` is the wrong primitive here. The reuse rule you quoted only bites because a cancelled `Shutdown` restores `active`, which forces a full drain before any new `Add`.

If you swap the WaitGroup for a self-managed in-flight counter, that constraint goes away — you own the counter, so there's no Add-after-Wait rule. You can then bound the wait *and* still accept new registrations after a cancellation, and you can release `a.mu` before waiting (so a slow init doesn't block readers or deadlock a re-entrant `Init`):

```go
type EvaluationAPI struct {
    // ...
    initMu    sync.Mutex
    initCount int
    initIdle  chan struct{} // set by a waiter; closed when initCount hits 0
}

// init start (under a.mu in initNew), replaces initWg.Add(1):
a.initMu.Lock()
a.initCount++
a.initMu.Unlock()

// init done, replaces defer initWg.Done():
a.initMu.Lock()
a.initCount--
if a.initCount == 0 && a.initIdle != nil {
    close(a.initIdle)
    a.initIdle = nil
}
a.initMu.Unlock()

// Shutdown:
a.mu.Lock()
if a.state == evaluationAPIStateShutdown {
    a.mu.Unlock()
    return nil
}
a.state = evaluationAPIStateShutdown // admission closed: no new initCount++
a.mu.Unlock()

a.initMu.Lock()
if a.initCount > 0 {
    a.initIdle = make(chan struct{})
}
idle := a.initIdle
a.initMu.Unlock()

if idle != nil {
    select {
    case <-idle: // all in-flight inits finished
    case <-ctx.Done():
        a.initMu.Lock()
        a.initIdle = nil
        a.initMu.Unlock()
        a.mu.Lock()
        a.state = evaluationAPIStateActive
        a.mu.Unlock()
        return ctx.Err()
    }
}

a.mu.Lock()
defer a.mu.Unlock()
// existing teardown...
```

Because `state == shutdown` gates `initNew`, `initCount` only ever decreases once shutdown starts, so nothing races the wait; and after a cancelled shutdown restores `active`, a later `SetProvider` just does `initCount++`, which is always legal with a plain counter.

One caveat: dropping `a.mu` before the wait opens a window where a second concurrent `Shutdown` sees `state == shutdown` and returns `nil` before teardown finishes. If that matters, a distinct "shutting down" sub-state (so a second call joins rather than returns early) closes it.

Separately, the branch comment here ("Wait for in-flight initializations to finish before restoring state") is inaccurate — the wait already completed above the `select` — worth tweaking regardless of the above.

PR #538, opened 2026-09-03
fix: set ErrorReason when typed accessors return TYPE_MISMATCH

### LINKED ISSUE #537: Set ErrorReason when typed accessors return TYPE_MISMATCH
## Summary

Existing typed accessors can keep a successful provider resolution reason when their type assertion fails.

For example, `StringValueDetails`, `FloatValueDetails`, and similar typed accessors can return the typed default value with an error code and error message, while `Reason` remains the provider's successful reason such as `STATIC`.

## Required change

Update the type-mismatch paths in the existing typed accessors so that they set `Reason` to `ErrorReason` when they return the default value because the provider value does not match the accessor type.

Keep the current error code and error message behavior.

Add or update tests for each affected accessor path. The tests must verify that:
- the accessor returns the supplied typed default value;
- the result has the existing type-mismatch error code and error message; and
- `Reason` is `ErrorReason`.

## Rationale

A successful resolution reason does not describe a result that the SDK replaced with the caller-supplied default after a type mismatch. Setting `ErrorReason` makes typed accessor behavior consistent with `Client.evaluate` provider-error handling and with the generic object accessor conversion-failure behavior.

## Affected areas

- Existing typed evaluation accessors, including `StringValueDetails` and `FloatValueDetails`
- Their type-mismatch tests

## Backlinks

- Pull request: https://github.com/open-feature/go-sdk/pull/536
- Review comment: https://github.com/open-feature/go-sdk/pull/536#discussion_r3909749282

Requested by: @sahidvelji

sahidvelji:
Closes #537.

The typed detail accessors copy the provider's `EvaluationDetails` verbatim when their type assertion fails. They overwrite `ErrorCode` and `ErrorMessage` but not `Reason`, so a result holding the caller's *default* value could still report a successful reason such as `STATIC`.

Spec [1.3.4](https://openfeature.dev/specification/sections/flag-evaluation#requirement-134) calls a type mismatch abnormal execution, and [1.4.9](https://openfeature.dev/specification/sections/flag-evaluation#requirement-149) says `reason` should then indicate an error — which is what `Client.evaluate`, `multi`, and `memprovider` already do.

The four mismatch blocks were identical apart from type and message, so they're now one generic helper that sets `Reason = ErrorReason` in one place. Error codes and messages are unchanged.

## Notes for reviewers

- **These branches are unreachable today.** `evaluate` takes its value from statically typed provider results (`BoolResolutionDetail` wraps a `bool`), so no `FeatureProvider` can make `evalDetails.Value.(bool)` fail. They're defensive code — which is why nothing tested them, and why the test drives the helper rather than a provider mock. Verified it fails on `STATIC` without the fix.
- **The e2e step is dormant.** `theReasonShouldIndicateAnErrorAndTheErrorCodeShouldIndicateATypeMismatchWith` already asserts `ErrorReason`, but reaches TYPE_MISMATCH via the provider-error path, and its scenario is no longer in the current `test-harness` `evaluation.feature`.
- **`ObjectValueDetails` is untouched** — no assertion, so no client-side mismatch to fix.
- **Adjacent gap, not fixed here:** `evaluate`'s other error exits (invalid UTF-8 key, NOT_READY/FATAL, hook errors) leave `Reason` empty rather than `ErrorReason`. Happy to open a separate issue.
- Also filled in the missing `Variant`/`Reason` assertions in the `BooleanValueDetails` happy-path subtest, which its four siblings already had.

`make test`, `make e2e-test` and `make lint` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -373,15 +373,7 @@ func (c *Client) BooleanValueDetails(ctx context.Context, flag string, defaultVa
 
 	value, ok := evalDetails.Value.(bool)
 	if !ok {
-		err := errors.New("evaluated value is not a boolean")
-		boolEvalDetails := BooleanEvaluationDetails{
-			Value:             defaultValue,
-			EvaluationDetails: evalDetails.EvaluationDetails,
-		}
-		boolEvalDetails.ErrorCode = TypeMismatchCode
-		boolEvalDetails.ErrorMessage = err.Error()
-
-		return boolEvalDetails, err
+		return typeMismatchDetails(defaultValue, evalDetails.EvaluationDetails, errors.New("evaluated value is not a boolean"))
 	}
 
 	return BooleanEvaluationDetails{
@@ -417,15 +409,7 @@ func (c *Client) StringValueDetails(ctx context.Context, flag string, defaultVal
 
 	value, ok := evalDetails.Value.(string)
 	if !ok {
-		err := errors.New("evaluated value is not a string")
-		strEvalDetails := StringEvaluationDetails{
-			Value:             defaultValue,
-			EvaluationDetails: evalDetails.EvaluationDetails,
-		}
-		strEvalDetails.ErrorCode = TypeMismatchCode
-		strEvalDetails.ErrorMessage = err.Error()
-
-		return strEvalDetails, err
+		return typeMismatchDetails(defaultValue, evalDetails.EvaluationDetails, errors.New("evaluated value is not a string"))
 	}
 
 	return StringEvaluationDetails{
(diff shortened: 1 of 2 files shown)


openfeature/client.go
```diff
@@ -373,15 +373,7 @@ func (c *Client) BooleanValueDetails(ctx context.Context, flag string, defaultVa
...
-		err := errors.New("evaluated value is not a boolean")
-		boolEvalDetails := BooleanEvaluationDetails{
-			Value:             defaultValue,
-			EvaluationDetails: evalDetails.EvaluationDetails,
-		}
-		boolEvalDetails.ErrorCode = TypeMismatchCode
-		boolEvalDetails.ErrorMessage = err.Error()
-
-		return boolEvalDetails, err
+		return typeMismatchDetails(defaultValue, evalDetails.EvaluationDetails, errors.New("evaluated value is not a boolean"))
```
erka: It would be nice to move these `errors.New("evaluated value is not a <type>")` to package level vars

erka (approved):

openfeature/client.go
sahidvelji: Done in 9ed9292 — the four messages are now package-level `errNotBoolean` / `errNotString` / `errNotFloat64` / `errNotInt64` vars next to the interface guard. Kept them unexported to match `errNilProvider`; happy to export if you want callers matching with `errors.Is`.

JamieSinn (approved):

toddbaert (approved):

PR #539, opened 2026-09-03
feat: support updating the in-memory provider's flag set

### LINKED ISSUE #530: in-memory provider cannot update its flag set or emit PROVIDER_CONFIGURATION_CHANGED
## Summary

`memprovider.InMemoryProvider` has no way to update its flag set and emits no events. [Appendix A][appendix-a] requires both:

> The provider **must** support a means of updating the `flag set`, resulting in the emission of `PROVIDER_CONFIGURATION_CHANGED` events.
>
> - The change event should consider all flags changed; a union of all previous and all new flag keys should be supplied in the `flags changed` field.

The provider's whole surface is `Metadata`, the five evaluation methods, `Hooks` and `Track`. It implements neither `openfeature.EventHandler` nor `openfeature.StateHandler`, so there is nothing to call and nothing to emit.

## Comparison with the other SDKs

| SDK | update method | emits `PROVIDER_CONFIGURATION_CHANGED` |
|---|---|---|
| JavaScript | `putConfiguration()` | yes |
| Java | `updateFlag()` | yes |
| **Go** | **none** | **no** |

## Why this matters beyond the missing method

Appendix A also says:

> E2E tests must utilize [in-memory provider](#in-memory-provider) defined within the SDK and must be self-contained.

So the gap propagates: go-sdk's own [Appendix B][appendix-b] suite cannot cover configuration-change events, because the provider it is required to use cannot produce one. Any consumer wanting to test their own reaction to `PROVIDER_CONFIGURATION_CHANGED` has to write their own provider.

## How it surfaced

Building the Go implementation of the provider conformance suite ([go-sdk-contrib#938](https://github.com/open-feature/go-sdk-contrib/issues/938), part of [spec#417](https://github.com/open-feature/spec/issues/417)). The suite's in-memory self-test has to leave the `@configuration-change` capability undeclared and report the scenario as skipped.

## Suggested shape

Something equivalent to the Java and JS providers:

```go
// UpdateFlags replaces the flag set and emits PROVIDER_CONFIGURATION_CHANGED
// naming the union of the previous and new flag keys.
func (i *InMemoryProvider) UpdateFlags(flags map[string]InMemoryFlag)
```

Two constraints worth calling out, both learned while working around this:

1. **It has to be a pointer type.** The SDK compares providers with `reflect.DeepEqual` unless the dynamic type is a pointer (`providerReference.equals`). `NewInMemoryProvider` currently returns a value, so two providers holding equal flag sets compare as the same provider. Adding an event channel to a value type would also copy the channel around.
2. **The flag map must not be mutated in place.** `InMemoryProvider` keeps the map it was constructed with by reference and reads it without synchronisation, so mutating it while an evaluation is in flight is a data race. Rebuilding over a copied map keeps readers on an immutable snapshot.

A working reference that does both, wrapping rather than replacing this provider, is `tck.ControllableProvider` in the linked PR — every resolution decision is still made by `memprovider`, only update-and-emit is added. Happy to open a PR here instead if that shape looks right.

[appendix-a]: https://github.com/open-feature/spec/blob/main/specification/appendix-a-included-utilities.md
[appendix-b]: https://github.com/open-feature/spec/blob/main/specification/appendix-b-gherkin-suites.md

sahidvelji:
Fixes #530

Appendix A requires the in-memory provider to support updating its flag set and emit `PROVIDER_CONFIGURATION_CHANGED` with the union of the previous and new flag keys. Go had neither, unlike JS (`putConfiguration()`) and Java (`updateFlag()`), which blocks the conformance suite from covering configuration-change scenarios.

Adds `UpdateFlags` and `EventChannel`, guards the flag set and tracking events with a mutex, and copies the flag map on construction and update so readers hold an immutable snapshot.

## ⚠️ Breaking change

`NewInMemoryProvider` now returns `*InMemoryProvider`, with all methods on pointer receivers. It has to be a pointer: `providerReference.equals` falls back to `reflect.DeepEqual` for non-pointer providers, so two providers with equal flag sets compared as the same provider — and that `DeepEqual` walks the flag map unsynchronised while `UpdateFlags` mutates it.

Migration affects only code that names the type explicitly:

```diff
-var p memprovider.InMemoryProvider = memprovider.NewInMemoryProvider(flags)
+var p *memprovider.InMemoryProvider = memprovider.NewInMemoryProvider(flags)
```

`p := memprovider.NewInMemoryProvider(flags)` is unaffected. Nothing in this repo needed changing.

## Notes

- The event is dropped rather than blocking when nothing drains the channel (provider not registered); a blocking send would deadlock the caller.
- The channel is never closed — the SDK's forwarding goroutine exits on its own shutdown signal, and closing would panic a later `UpdateFlags`.
- Construction no longer aliases the caller's map. That aliasing was unsynchronised, so relying on it was never safe.
- `StateHandler` is deliberately left for a separate change.

Tests cover the flag swap, removal, the sorted union, event type and provider name, constructor copy semantics, non-blocking emission, concurrent update/evaluate/track under `-race`, and one end-to-end case asserting a registered handler fires.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/openfeature/memprovider/in_memory_provider.go
+++ b/openfeature/memprovider/in_memory_provider.go
@@ -187,20 +187,15 @@ func (i *InMemoryProvider) find(flag string) (*InMemoryFlag, *openfeature.Provid
 // with an API.
 func (i *InMemoryProvider) UpdateFlags(flags map[string]InMemoryFlag) {
 	i.mu.Lock()
-	changed := make([]string, 0, len(i.flags)+len(flags))
-	for key := range i.flags {
-		changed = append(changed, key)
-	}
-	for key := range flags {
-		if _, ok := i.flags[key]; !ok {
-			changed = append(changed, key)
-		}
-	}
+	changed := slices.AppendSeq(slices.Collect(maps.Keys(i.flags)), maps.Keys(flags))
 	// Readers keep an immutable snapshot; the caller's map is never retained.
 	i.flags = maps.Clone(flags)
 	i.mu.Unlock()
 
+	// Sorting makes the union deterministic and puts any key present in both
+	// the old and the new set next to its duplicate for Compact to drop.
 	slices.Sort(changed)
+	changed = slices.Compact(changed)
 
 	select {
 	case i.events <- openfeature.Event{
--- a/openfeature/memprovider/in_memory_provider.go
+++ b/openfeature/memprovider/in_memory_provider.go
@@ -4,6 +4,9 @@ package memprovider
 import (
 	"context"
 	"fmt"
+	"maps"
+	"slices"
+	"sync"
 
 	"github.com/open-feature/go-sdk/openfeature"
 )
@@ -13,25 +16,40 @@ const (
 	Disabled State = "DISABLED"
 )
 
+const providerName = "InMemoryProvider"
+
+// eventChannelBuffer absorbs bursts of events. A registered provider is drained
+// continuously by the SDK's event executor.
+const eventChannelBuffer = 5
+
+var (
+	_ openfeature.FeatureProvider = (*InMemoryProvider)(nil)
+	_ openfeature.EventHandler    = (*InMemoryProvider)(nil)
+	_ openfeature.Tracker         = (*InMemoryProvider)(nil)
+)
+
 type InMemoryProvider struct {
+	mu             sync.RWMutex
 	flags          map[string]InMemoryFlag
 	trackingEvents map[string][]InMemoryEvent
+	events         chan openfeature.Event
 }
 
-func NewInMemoryProvider(from map[string]InMemoryFlag) InMemoryProvider {
-	return InMemoryProvider{
-		flags:          from,
+func NewInMemoryProvider(from map[string]InMemoryFlag) *InMemoryProvider {
+	return &InMemoryProvider{
+		flags:          maps.Clone(from),
 		trackingEvents: map[string][]InMemoryEvent{},
+		events:         make(chan openfeature.Event, eventChannelBuffer),
 	}
 }
 
-func (i InMemoryProvider) Metadata() openfeature.Metadata {
+func (i *InMemoryProvider) Metadata() openfeature.Metadata {
 	return openfeature.Metadata{
-		Name: "InMemoryProvider",
+		Name: providerName,
 	}
 }
 
-func (i InMemoryProvider) BooleanEvaluation(ctx context.Context, flag string, defaultValue bool, flatCtx openfeature.FlattenedContext) openfeature.BoolResolutionDetail {
+func (i *InMemoryProvider) BooleanEvaluation(ctx context.Context, flag string, defaultValue bool, flatCtx openfeature.FlattenedContext) openfeature.BoolResolutionDetail {
 	memoryFlag, details, ok := i.find(flag)
 	if !ok {
 		return openfeature.BoolResolutionDetail{
@@ -49,7 +67,7 @@ func (i InMemoryProvider) BooleanEvaluation(ctx context.Context, flag string, de
 	}
 }
 
-func (i InMemoryProvider) StringEvaluation(ctx context.Context, flag string, defaultValue string, flatCtx openfeature.FlattenedContext) openfeature.StringResolutionDetail {
+func (i *InMemoryProvider) StringEvaluation(ctx context.Context, flag string, defaultValue string, flatCtx openfeature.FlattenedContext) openfeature.StringResolutionDetail {
 	memoryFlag, details, ok := i.find(flag)
 	if !ok {
 		return openfeature.StringResolutionDetail{
@@ -67,7 +85,7 @@ func (i InMemoryProvider) StringEvaluation(ctx context.Context, flag string, def
 	}
 }
 
-func (i InMemoryProvider) FloatEvaluation(ctx context.Context, flag string, defaultValue float64, flatCtx openfeature.FlattenedContext) openfeature.FloatResolutionDetail {
+func (i *InMemoryProvider) FloatEvaluation(ctx context.Context, flag string, defaultValue float64, flatCtx openfeature.FlattenedContext) openfeature.FloatResolutionDetail {
 	memoryFlag, details, ok := i.find(flag)
 	if !ok {
 		return openfeature.FloatResolutionDetail{
@@ -85,7 +103,7 @@ func (i InMemoryProvider) FloatEvaluation(ctx context.Context, flag string, defa
 	}
 }
 
-func (i InMemoryProvider) IntEvaluation(ctx context.Context, flag string, defaultValue int64, flatCtx openfeature.FlattenedContext) openfeature.IntResolutionDetail {
+func (i *InMemoryProvider) IntEvaluation(ctx context.Context, flag string, defaultValue int64, flatCtx openfeature.FlattenedContext) openfeature.IntResolutionDetail {
 	memoryFlag, details, ok := i.find(flag)
 	if !ok {
 		return openfeature.IntResolutionDetail{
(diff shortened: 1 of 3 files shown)


erka (commented):
It looks good, and `maps.Copy` might be a bit overkill. `flag.Variants` could still be mutated, and I’m not sure we want to do a deep copy. I wonder if a simpler option would be to just document that the provider expects to take ownership of the flags' data.

sahidvelji:
> It looks good, and `maps.Copy` might be a bit overkill. `flag.Variants` could still be mutated, and I’m not sure we want to do a deep copy. I wonder if a simpler option would be to just document that the provider expects to take ownership of the flags' data.

Agreed on no deep copy — it isn't reachable anyway, since `ObjectEvaluation` returns the variant value itself and an object flag's map escapes to every evaluator regardless.

I'd keep the shallow `maps.Clone`, though. It's what JS (`{ ...flagConfiguration }`) and Java (`new ConcurrentHashMap<>(flags)`, `putAll`) already do on both construction and update, so the current Go aliasing is the outlier. It's also what makes `UpdateFlags` the only path that changes the flag set — otherwise a caller mutating their retained map changes resolution with no `PROVIDER_CONFIGURATION_CHANGED` event.

Taking the doc suggestion on top: I'll note on `NewInMemoryProvider`, `UpdateFlags` and in the README that the provider takes ownership of each flag's `Variants` and `ContextEvaluator`.

openfeature/memprovider/in_memory_provider.go
```diff
@@ -154,10 +185,56 @@ func (i InMemoryProvider) find(flag string) (*InMemoryFlag, *openfeature.Provide
...
+// still delivered once the SDK attaches a listener. Once the buffer is full the
+// event is dropped rather than blocking the caller.
+func (i *InMemoryProvider) UpdateFlags(flags map[string]InMemoryFlag) {
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	changed := slices.AppendSeq(slices.Collect(maps.Keys(i.flags)), maps.Keys(flags))
+	// Readers keep an immutable snapshot; the caller's map is never retained.
+	i.flags = maps.Clone(flags)
+
```
erka, line 208:
```suggestion
	// Readers keep an immutable snapshot; the caller's map is never retained.
	flags = maps.Clone(flags)
	i.mu.Lock()
	oldKeys := slices.Collect(maps.Keys(i.flags))
	i.flags = flags
	i.mu.Unlock()

	changed := slices.AppendSeq(oldKeys, maps.Keys(flags))
```

openfeature/memprovider/in_memory_provider_test.go
```diff
@@ -429,3 +360,228 @@ func TestInMemoryProvider_Track(t *testing.T) {
...
+	})
+
+	// Drain events so the writer is never throttled by a full buffer.
+	stop := make(chan struct{})
+	defer close(stop)
+	go func() {
+		for {
+			select {
+			case <-memoryProvider.EventChannel():
+			case <-stop:
```
erka:
```suggestion
			case <-t.Context().Done():
```

openfeature/memprovider/in_memory_provider_test.go
```diff
@@ -429,3 +360,228 @@ func TestInMemoryProvider_Track(t *testing.T) {
...
+}
+
+func TestInMemoryProvider_ConcurrentUpdateAndEvaluation(t *testing.T) {
+	memoryProvider := NewInMemoryProvider(map[string]InMemoryFlag{
+		"flagA": boolFlag("flagA", "true"),
+	})
+
+	// Drain events so the writer is never throttled by a full buffer.
+	stop := make(chan struct{})
+	defer close(stop)
```
erka:
```suggestion
```

openfeature/memprovider/in_memory_provider_test.go
```diff
@@ -429,3 +360,228 @@ func TestInMemoryProvider_Track(t *testing.T) {
...
+	t.Cleanup(openfeature.Shutdown)
+
+	received := make(chan openfeature.EventDetails, 1)
+	callback := func(details openfeature.EventDetails) {
+		received <- details
+	}
+	openfeature.AddHandler(openfeature.ProviderConfigChange, &callback)
+	t.Cleanup(func() {
+		openfeature.RemoveHandler(openfeature.ProviderConfigChange, &callback)
+	})
```
erka:
```suggestion
	api := isolated.NewAPI()
	t.Cleanup(func() {
		_ = api.Shutdown(context.Background()) //nolint:usetesting
	})

	if err := api.SetProviderAndWait(t.Context(), memoryProvider); err != nil {
		t.Fatal(err)
	}

	received := make(chan openfeature.EventDetails, 1)
	callback := func(details openfeature.EventDetails) {
		received <- details
	}
	api.AddHandler(openfeature.ProviderConfigChange, &callback)
	t.Cleanup(func() {
		api.RemoveHandler(openfeature.ProviderConfigChange, &callback)
	})
```

erka (approved):
Great work! Some small nits.

While I understand why this could/should be viewed as a breaking change but we really don't want `!` in the title or `breaking changes` in the description. Release-Please action will bump version to `v2` and this means that go package name should be changed to `../v2` as well.

openfeature/memprovider/in_memory_provider.go
sahidvelji: Took the clone out of the critical section, but kept the event construction and send under the lock. Unlocking first is what b822789 fixed: two concurrent `UpdateFlags` calls can enqueue their events in the opposite order from the flag swaps, and the events carry no version for a consumer to recover the order. The send is non-blocking, so holding the lock across it can't deadlock.

JamieSinn (approved):

openfeature/memprovider/in_memory_provider.go
```diff
@@ -129,20 +154,26 @@ func (i InMemoryProvider) ObjectEvaluation(ctx context.Context, flag string, def
...
+func (i *InMemoryProvider) Track(ctx context.Context, trackingEventName string, evalCtx openfeature.EvaluationContext, details openfeature.TrackingEventDetails) {
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
 	i.trackingEvents[trackingEventName] = append(i.trackingEvents[trackingEventName], InMemoryEvent{
 		Value:             details.Value(),
 		Data:              details.Attributes(),
 		ContextAttributes: evalCtx.Attributes(),
 	})
 }
```
kripa-sindhu-007, line 171: The mutex here does the hard half of #545. What's left is that `trackingEvents` has no way out ; declaration , init, this one writen and `TestInMemoryProvider_Track` asserts nothing. Since the lock is yours, folding the getter in here beats a follow-up that has to wait for it:

```suggestion

// TrackingEvents returns the events Track recorded for trackingEventName, in
// the order they were recorded, and an empty slice for a name never tracked.
//
// The events are copies, so a caller reading them back cannot disturb what the
// provider recorded.
func (i *InMemoryProvider) TrackingEvents(trackingEventName string) []InMemoryEvent {
      i.mu.RLock()
      defer i.mu.RUnlock()

      recorded := i.trackingEvents[trackingEventName]

      events := make([]InMemoryEvent, len(recorded))
      for n, event := range recorded {
              // the maps travel with the event, so cloning the slice alone still aliases, see #545
              event.Data = maps.Clone(event.Data)
              event.ContextAttributes = maps.Clone(event.ContextAttributes)
              events[n] = event
      }

      return events
}
```

`Data` and `ContextAttributes` are maps, so copying the slice alone still hands the caller a live handle on what was recorded; I have a test that fails without those two clones.

@sahidvelji Would you rather take it here, or should I keep it as a follow-up on #545 so this PR stays scoped to #530? Per-name is what a test usually wants; a whole-map snapshot is the other shape. Java and JS implement no tracking on their in-memory providers, so there's no sibling to follow.

PR #565, opened 2026-09-05
fix: coerce int types to float64 in memprovider float evaluation

### LINKED ISSUE #564: [DOCS] memprovider: genericResolve's comment claims int-to-float64 coercion that isn't implemented
## Observed behavior

`genericResolve`'s `float64` branch has a comment promising integer coercion that isn't implemented:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/memprovider/in_memory_provider.go#L159-L197

The comment reads "Convert float32 to float64 **and int types to float64**", but only `float32` is handled. So a variant declared as `1` (a Go `int`) evaluated via `FloatValue` yields TYPE_MISMATCH — asymmetric with the `int64` branch, which does accept `int8`/`int16`/`int32`/`int`.

```
FloatEvaluation(int variant 1)   -> value=0.1 reason=ERROR errorCode="TYPE_MISMATCH"
FloatEvaluation(float32 1.5)     -> value=1.5 reason=STATIC errorCode=""
```

## Expected Behavior

Either implement the int→float64 coercion the comment describes, or drop that clause from the comment.

Given the function's stated goal — "to provide a more forgiving API for test flag configuration" — accepting integer literals for float flags seems like the intent, since `Variants: map[string]any{"on": 1}` is an easy thing to write by hand and the failure is non-obvious.

## Steps to reproduce

```go
p := memprovider.NewInMemoryProvider(map[string]memprovider.InMemoryFlag{
	"float-flag": {
		Key:            "float-flag",
		State:          memprovider.Enabled,
		DefaultVariant: "one",
		Variants:       map[string]any{"one": 1}, // Go int, not float64
	},
})
// FloatEvaluation returns TYPE_MISMATCH rather than 1.0
```

alorentiar:
The float64 branch in genericResolve only converted float32, even though
its comment says it handles int types as well. In practice a float flag
configured with a plain int variant, e.g. `Variants: map[string]any{"on": 1}`,
came back as TYPE_MISMATCH through FloatEvaluation, while the int64 branch
accepts those same int types fine.

Added the signed int cases (int, int8, int16, int32, int64) to the float64
conversion, mirroring the int64 branch. Unsigned types still fall through to
a type mismatch, which is what the doc comment on the helper states.

Tests cover each coerced type plus one case locking in the unsigned behavior.

Fixes #564

### DIFF
--- a/openfeature/memprovider/in_memory_provider.go
+++ b/openfeature/memprovider/in_memory_provider.go
@@ -157,7 +157,7 @@ func (i InMemoryProvider) find(flag string) (*InMemoryFlag, *openfeature.Provide
 // helpers
 
 // genericResolve is a helper to extract type verified evaluation and fill openfeature.ProviderResolutionDetail.
-// It coerces smaller numeric types to their canonical forms (int* -> int64, float32 -> float64)
+// It coerces smaller numeric types to their canonical forms (int* -> int64, int* | float32 -> float64)
 // to provide a more forgiving API for test flag configuration.
 //
 // Note: Only signed integer types are supported for conversion. Unsigned integer types
--- a/openfeature/memprovider/in_memory_provider.go
+++ b/openfeature/memprovider/in_memory_provider.go
@@ -187,6 +187,16 @@ func genericResolve[T comparable](value any, defaultValue T, detail *openfeature
 		switch v := value.(type) {
 		case float32:
 			return any(float64(v)).(T)
+		case int8:
+			return any(float64(v)).(T)
+		case int16:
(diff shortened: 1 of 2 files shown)


sahidvelji (approved):

erka:
We need to change the godoc for `genericResolve` as well to match the implementation.

```go
 // It coerces smaller numeric types to their canonical forms (int* -> int64, float32 -> float64)
```

erka (approved):
Thanks @alorentiar for your contribution!

PR #566, opened 2026-09-05
fix: return default value from ObjectValueDetails on abnormal execution

### LINKED ISSUE #553: [BUG] ObjectValueDetails returns the provider's value on abnormal execution instead of the default
## Observed behavior

`ObjectValueDetails` returns `c.evaluate(...)` directly:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/client.go#L525-L543

…unlike the other four typed detail methods, which explicitly overwrite `Value: defaultValue` when `err != nil`:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/client.go#L357-L391

Since `evaluate` has already assigned `evalDetails.Value = resolution.Value` by the time an after-hook fails, the provider's value escapes alongside a non-nil error:

```
ObjectValueDetails  -> value=PROVIDER-VALUE reason="STATIC" errorCode="" err=after hook: after hook boom
BooleanValueDetails -> value=false          reason="STATIC" errorCode="" err=after hook: after hook boom
```

A secondary defect is visible on both lines: even the guarded methods carry the provider's **success** reason (`STATIC`) and an empty error code into what is an abnormal result.

## Expected Behavior

Per [1.4.10](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1410) and [4.4.7](https://openfeature.dev/specification/sections/hooks#requirement-447), abnormal execution should yield the default value, with an error code set and the reason indicating an error.

The siblings replace the whole details struct on any thrown stage:

- **Java** — `OpenFeatureClient.evaluateFlag`'s catch calls `enrichDetailsWithErrorDefaults`, setting `value = defaultValue` and `reason = Reason.ERROR`. The object path shares this, so an after-hook throw returns the default regardless of type.
- **JS** — `getErrorEvaluationDetails(..., defaultValue, ...)` from the catch around `afterHooks`.
- **Python** — the after-hook throw propagates to the handler that sets `value=default_value`, `reason=Reason.ERROR`.

Related but narrower: #537 / #538 cover the reason on the *type-mismatch* paths. This is the after-hook / abnormal-return path, and additionally the `Value` leak unique to `ObjectValueDetails`.

## Steps to reproduce

```go
// provider returns "PROVIDER-VALUE" from ObjectEvaluation
// failAfterHook.After returns an error
client.AddHooks(failAfterHook{})

d, err := client.ObjectValueDetails(ctx, "f", "MY-DEFAULT", of.EvaluationContext{})
// d.Value == "PROVIDER-VALUE", err != nil
// expected d.Value == "MY-DEFAULT"
```

kripa-sindhu-007:
## This PR

- returns the default value from `ObjectValueDetails` when evaluation ends abnormally, instead of the value the provider resolved
- adds an `Object with erroring after hook` subtest to `TestRequirement_1_4_9` covering the hook-error path

`ObjectValueDetails` returned `c.evaluate(...)` directly, so on abnormal execution the caller received whatever the provider resolved. The five typed accessors don't behave this way — `BooleanValueDetails` and friends reset `Value` to the supplied default before returning the error.

This is reachable through a failing hook. With an after hook that errors:

```
object: value=map[from:provider]  err=after hook: after hook boom
bool  : value=false               err=after hook: after hook boom
```

Requirement 1.4.10 states that flag evaluation calls must always return the default value in the event of abnormal execution.

### Related Issues

Fixes #553

### Notes

- The method's existing doc comment already documents this behaviour ("defaultValue is returned if an error occurs"), so this makes the code honour a contract it already advertises.
- `ObjectValue` was already correct — it guards on the error and returns the default. Only the `...Details` variant was affected.
-  The reset happens on any error rather than only on the hook path, matching the typed accessors. The other error paths already return the default, so this is a no-op for them today — but only because `evalDetails.Value = resolution.Value` happens to sit below their return points. Guarding unconditionally keeps the accessor correct regardless of where that assignment moves.
- The existing `Object` subtest drives abnormal execution through a resolution error, which returns before the resolved value is assigned. A failing after hook returns after it, so that path was previously uncovered.

### How to test

```
go test -tags testtools -race -count=1 -v -run 'TestRequirement_1_4_9' ./openfeature/
```

### DIFF
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -766,7 +766,7 @@ func (c *Client) evaluate(
 		c.errorHooks(ctx, hookCtx, hooks, err, options)
 		evalDetails.Reason = ErrorReason
 		evalDetails.ErrorCode = GeneralCode
-		evalDetails.ErrorMessage = err.Error()
+		evalDetails.ErrorMessage = "error executing after hooks"
 		return evalDetails, err
 	}
 
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -764,6 +764,9 @@ func (c *Client) evaluate(
 	if err := c.afterHooks(ctx, hookCtx, hooks, evalDetails, options); err != nil {
 		err = fmt.Errorf("after hook: %w", err)
 		c.errorHooks(ctx, hookCtx, hooks, err, options)
+		evalDetails.Reason = ErrorReason
+		evalDetails.ErrorCode = GeneralCode
+		evalDetails.ErrorMessage = err.Error()
 		return evalDetails, err
 	}
 
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -539,7 +539,15 @@ func (c *Client) ObjectValueDetails(ctx context.Context, flag string, defaultVal
 		option(evalOptions)
 	}
 
-	return c.evaluate(ctx, flag, Object, defaultValue, evalCtx, *evalOptions)
+	evalDetails, err := c.evaluate(ctx, flag, Object, defaultValue, evalCtx, *evalOptions)
+	if err != nil {
+		return InterfaceEvaluationDetails{
+			Value:             defaultValue,
+			EvaluationDetails: evalDetails.EvaluationDetails,
+		}, err
+	}
+
+	return evalDetails, nil
 }
 
 // Boolean performs a flag evaluation that returns a boolean. Any error
--- a/openfeature/client.go
+++ b/openfeature/client.go
@@ -764,9 +764,6 @@ func (c *Client) evaluate(
 	if err := c.afterHooks(ctx, hookCtx, hooks, evalDetails, options); err != nil {
 		err = fmt.Errorf("after hook: %w", err)
 		c.errorHooks(ctx, hookCtx, hooks, err, options)
-		evalDetails.Reason = ErrorReason
-		evalDetails.ErrorCode = GeneralCode
-		evalDetails.ErrorMessage = "error executing after hooks"
 		return evalDetails, err
 	}
 
--- a/openfeature/client_test.go
+++ b/openfeature/client_test.go
@@ -654,6 +654,41 @@ func TestRequirement_1_4_9(t *testing.T) {
 			t.Errorf("expected default value from ObjectValueDetails, got %v", value)
 		}
 	})
+
+	// A hook error is abnormal execution too. The existing subtests above drive
+	// abnormal execution through a resolution error, which returns before the
+	// resolved value is assigned; a failing after hook returns after it.
+	t.Run("Object with erroring after hook", func(t *testing.T) {
+		t.Cleanup(resetSingleton)
+
+		mocks := hydratedMocksForClientTests(t, 1)
+		client := newClient("test-client", mocks.providerBinding, mocks.clientHandlerAPI)
+
+		type obj struct {
+			foo string
+		}
+		defaultValue := obj{foo: "bar"}
+		resolvedValue := obj{foo: "resolved"}
+
+		mocks.providerAPI.EXPECT().ObjectEvaluation(t.Context(), flagKey, defaultValue, flatCtx).
+			Return(InterfaceResolutionDetail{Value: resolvedValue})
+
+		mockHook := NewMockHook(gomock.NewController(t))
+		mockHook.EXPECT().Before(gomock.Any(), gomock.Any(), gomock.Any())
+		mockHook.EXPECT().Error(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
+			After(mockHook.EXPECT().After(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
+				Return(errors.New("forced")))
+		mockHook.EXPECT().Finally(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
+
+		valueDetails, err := client.ObjectValueDetails(t.Context(), flagKey, defaultValue, evalCtx, WithHooks(mockHook))
+		if err == nil {
+			t.Error("expected ObjectValueDetails to return an error, got nil")
+		}
+
+		if valueDetails.Value.(obj) != defaultValue {
+			t.Errorf("expected default value from ObjectValueDetails, got %v", valueDetails.Value)
+		}
+	})
 }
 
 // TODO Requirement_1_4_10
--- a/openfeature/client_test.go
+++ b/openfeature/client_test.go
@@ -688,6 +688,14 @@ func TestRequirement_1_4_9(t *testing.T) {
 		if valueDetails.Value.(obj) != defaultValue {
 			t.Errorf("expected default value from ObjectValueDetails, got %v", valueDetails.Value)
 		}
+
+		if valueDetails.Reason != ErrorReason {
+			t.Errorf("expected reason %s, got %s", ErrorReason, valueDetails.Reason)
+		}
+
+		if valueDetails.ErrorCode != GeneralCode {
+			t.Errorf("expected error code %s, got %s", GeneralCode, valueDetails.ErrorCode)
+		}
 	})
 }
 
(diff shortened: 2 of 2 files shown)


kripa-sindhu-007:
As per the coderabbit review comment I have fixed the part as correctly pointed out.

openfeature/client.go
```diff
@@ -756,6 +764,9 @@ func (c *Client) evaluate(
 	if err := c.afterHooks(ctx, hookCtx, hooks, evalDetails, options); err != nil {
 		err = fmt.Errorf("after hook: %w", err)
 		c.errorHooks(ctx, hookCtx, hooks, err, options)
+		evalDetails.Reason = ErrorReason
+		evalDetails.ErrorCode = GeneralCode
+		evalDetails.ErrorMessage = err.Error()
```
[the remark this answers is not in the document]
kripa-sindhu-007: I have sanitised it in the latest commit

openfeature/client.go
```diff
@@ -756,6 +764,9 @@ func (c *Client) evaluate(
 	if err := c.afterHooks(ctx, hookCtx, hooks, evalDetails, options); err != nil {
 		err = fmt.Errorf("after hook: %w", err)
 		c.errorHooks(ctx, hookCtx, hooks, err, options)
+		evalDetails.Reason = ErrorReason
+		evalDetails.ErrorCode = GeneralCode
+		evalDetails.ErrorMessage = "error executing after hooks"
 		return evalDetails, err
```
erka, line 767: The spec [4.4.7](https://openfeature.dev/specification/sections/hooks/#requirement-447) only states that errors in `beforeHook` is abnormal execution. I see it as `after` hook should not impact the evaluation result and their errors only should be reported to `errorHook`.

kripa-sindhu-007: You're right that 4.4.7 names only before, and 4.4.5 does say after-hook errors go to the error hooks. It's just I have understood the issue as resting on 1.4.10 instead, so the question is whether an after-hook error counts as abnormal execution. The spec doesn't say either way.

kripa-sindhu-007: @erka One more thing I need to clarify. If after hooks genuinely shouldn't affect the result, should evaluate stop returning that error too and would the typed accessors resetting Value to the default then be the actual bug, rather than ObjectValueDetails?

erka: I guess `return evalDetails, err` is another bug in the current implementation. We probably need just to remove this return statement and fallback to return in the end of the function.

kripa-sindhu-007: Both js-sdk and java-sdk treat an after hook error as abnormal execution and return the default, so removing the return would make go the only SDK where a failing after hook is invisible to the caller.

sahidvelji:
I'd push back on the CodeRabbit `ErrorMessage` finding rather than act on it.

The sanitization in 891c263 doesn't actually contain anything — `evaluate` still returns `fmt.Errorf("after hook: %w", err)`, so the hook's original text goes back to the caller anyway. Meanwhile it costs us two things: every other error path in `evaluate` reports the real message, and `telemetry.go:92` maps `ErrorMessage` to `error.message`, so a fixed string makes every after-hook failure look identical in telemetry. It's also outside the scope of #553.

Suggestion: revert it and keep this PR to the `ObjectValueDetails` fix. That also fixes CI — `client_test.go:701` references an undefined `afterHookErrorMessage`, so the test package currently doesn't compile.

kripa-sindhu-007:
> I'd push back on the CodeRabbit `ErrorMessage` finding rather than act on it.
> 
> The sanitization in 891c263 doesn't actually contain anything — `evaluate` still returns `fmt.Errorf("after hook: %w", err)`, so the hook's original text goes back to the caller anyway. Meanwhile it costs us two things: every other error path in `evaluate` reports the real message, and `telemetry.go:92` maps `ErrorMessage` to `error.message`, so a fixed string makes every after-hook failure look identical in telemetry. It's also outside the scope of #553.
> 
> Suggestion: revert it and keep this PR to the `ObjectValueDetails` fix. That also fixes CI — `client_test.go:701` references an undefined `afterHookErrorMessage`, so the test package currently doesn't compile.
> 

@sahidvelji noted. I was thinking the same when i got the CI failing mail. Will fix this tomorrow morning. Me in GMT +5:30 timezone.😴

openfeature/client.go
sahidvelji: I checked the other SDKs at HEAD:

- **JS** (server + web) — `afterHooks` runs inside the `try`; a throw falls to the `catch` and returns `getErrorEvaluationDetails(...)` → default value, `reason=ERROR`, `errorCode=GENERAL`.
- **Java** — same shape; `catch (Exception e)` calls `enrichDetailsWithErrorDefaults`, which sets value to the default and reason to `ERROR`.
- **Python** — same; both `except` branches build details with `value=default_value, reason=ERROR`.
- **Rust** — builds `EvaluationDetails::error_reason(flag_key, T::default())` on after-hook error, but then returns the original `Ok` result, so the caller gets the resolved value and no error. The discarded details look accidental rather than deliberate.

So three SDKs treat an after-hook error as abnormal execution and return the default. That matches the direction of this PR.

That said, I don't think we should settle it from precedent alone — @erka's reading of 4.4.7 is correct on the letter, and it's entirely possible the other SDKs are the ones that drifted. Worth getting input from the wider group on whether the JS/Java/Python behaviour is intended or a shared bug before we decide which way Go should go.

erka: cc @toddbaert

kripa-sindhu-007:
@sahidvelji  Reverted in 25de9c2. Both commits are gone, so the branch is back to just the ObjectValueDetails change and the test package compiles again. Agreed on the telemetry point, I had not considered telemetry.go:92 mapping ErrorMessage to error.message. make test, make lint and make e2e-test are green locally.

openfeature/client.go
kripa-sindhu-007: One more data point for whoever picks this up: TestRequirement_4_3_6 in hooks_test.go:469 already asserts a non-nil error when an after hook fails, and that predates this PR. So the current Go behaviour is covered by a test rather than incidental, which is worth knowing if the conclusion is that it should change.

I have reverted the after-hook changes out of this PR either way, so it is now only the ObjectValueDetails fix and does not depend on how this lands.

kripa-sindhu-007:
@sahidvelji @erka The ci workflow on this and on #573 is sitting at action_required — the first-time contributor gate, so no run has actually executed on either. Could someone approve them when you get a chance? Both are green locally on make test, make lint and make e2e-test.

openfeature/client.go
kripa-sindhu-007: @erka  @sahidvelji The ci run went through overnight, so this is green now at 5afdf7e — test, lint, e2e and codecov all passing.

The only open thread left is the after-hook one, and GitHub has marked it outdated: those commits are reverted, so nothing in the current diff turns on how that question lands. It stays parked with @toddbaert either way. Could this thread be resolved so it is not sitting on this PR as an open point, and could one of you take a look at the review when you get a chance?

toddbaert: Ya it's a bit of an oversight. It's the intent though. One of the explicit use-cases from the early days were validating after hooks that can allow you to bail out of invalid resolutions by throwing/returning an error ([relevant blog post](https://openfeature.dev/blog/creating-a-hook-for-the-go-sdk)).

4.4.3, 4.4.4, and especially 4.4.5 all sort of imply this, it should just be made explicit in 4.4.7 or a new point. I'll create an issue.

toddbaert: Opened https://github.com/open-feature/spec/issues/427 - any of you are welcome to take a stab at this - should be a pretty simple addition.

sahidvelji (approved):

erka (approved):

PR #541, opened 2026-09-03
fix: populate EvaluationDetails on NOT_READY/FATAL short-circuit

typotter:
## Motivation & Context

Fixes #540

When the provider state is `NOT_READY` or `FATAL`, the `evaluate()` function in `client.go` short-circuits and returns an `EvaluationDetails` struct without populating `Reason`, `ErrorCode`, or `ErrorMessage`. The error information is only available in the separate `error` return value. Consumers using `BooleanValueDetails` (and similar typed detail methods) receive an empty `ResolutionDetail`, which is inconsistent with the spec and with how other error paths in the same function behave.

## Change

Set `Reason`, `ErrorCode`, and `ErrorMessage` on the `evalDetails` struct before returning in both the `NotReadyState` and `FatalState` short-circuit blocks.

## Testing

Added two tests (`TestEvaluationDetails_NotReady`, `TestEvaluationDetails_Fatal`) that call `BooleanValueDetails` when the provider is in NOT_READY or FATAL state and assert that `Reason`, `ErrorCode`, and `ErrorMessage` are populated on the returned details struct. Full test suite passes.

Issue #540, opened 2026-09-03
evaluate() does not populate EvaluationDetails fields on NOT_READY/FATAL short-circuit


typotter:
## Description

When the provider is in `NOT_READY` or `FATAL` state, `client.evaluate()` short-circuits and returns `evalDetails` with empty `Reason`, `ErrorCode`, and `ErrorMessage` fields. The error is only communicated via the separate `error` return value.

This violates the OpenFeature spec (section 1.4.7): "In the case of abnormal execution, the evaluation details structure's error code field MUST contain an error code."

## Affected code

https://github.com/open-feature/go-sdk/blob/main/openfeature/client.go#L680-L718

```go
// L680-686: evalDetails initialized with no Reason/ErrorCode
evalDetails := InterfaceEvaluationDetails{
    Value: defaultValue,
    EvaluationDetails: EvaluationDetails{
        FlagKey:  flag,
        FlagType: flagType,
    },
}

// ...

// L716-718: returned as-is — Reason, ErrorCode, ErrorMessage all zero-value
if c.State() == NotReadyState {
    c.errorHooks(ctx, hookCtx, providerInvocationClientAPIHooks, ProviderNotReadyError, options)
    return evalDetails, ProviderNotReadyError
}

// L722-724: same issue
if c.State() == FatalState {
    c.errorHooks(ctx, hookCtx, providerInvocationClientAPIHooks, ProviderFatalError, options)
    return evalDetails, ProviderFatalError
}
```

## Impact

Callers that inspect `EvaluationDetails` (e.g. for logging or observability) see empty `Reason`/`ErrorCode`/`ErrorMessage`, making it impossible to distinguish NOT_READY from a successful evaluation returning the default. The error is only available via the separate `error` return, which many wrappers and hooks don't propagate into structured logs.

## Spec references

- [Requirement 1.4.7](https://openfeature.dev/specification/sections/flag-evaluation#requirement-147) — error code MUST be set on abnormal execution
- [Requirement 1.4.8](https://openfeature.dev/specification/sections/flag-evaluation#requirement-148) — reason MUST indicate ERROR on abnormal execution

PR #568, opened 2026-09-07
fix: populate error fields on the UTF-8 and before hook returns

### LINKED ISSUE #567: [BUG] evaluate() returns errors without an ErrorCode or Reason on the UTF-8 and before-hook paths
## Observed behavior

Two of `Client.evaluate`'s early returns hand back an error with `ErrorCode` and `Reason` both empty.

**Invalid UTF-8 flag key** — returns a parse error, but nothing is written to `evalDetails`, and the error hooks don't run either:

https://github.com/open-feature/go-sdk/blob/a0892c296bd8b6d55188687e49dcc7692594f015/openfeature/client.go#L665-L666

**Before-hook error** — the error hooks do run, but the fields are still unset:

https://github.com/open-feature/go-sdk/blob/a0892c296bd8b6d55188687e49dcc7692594f015/openfeature/client.go#L712-L715

Contrast the NOT_READY and FATAL short-circuits just above, which set `Reason`, `ErrorCode` and `ErrorMessage` before returning.

## Expected Behavior

Per [1.4.7](https://openfeature.dev/specification/sections/flag-evaluation#requirement-147), in cases of abnormal execution the evaluation details' error code field **MUST** contain an error code; [1.4.8](https://openfeature.dev/specification/sections/flag-evaluation#requirement-148) adds that the reason **SHOULD** indicate an error.

A before-hook error is unambiguously abnormal execution under [4.4.7](https://openfeature.dev/specification/sections/hooks#requirement-447), so 1.4.7 applies squarely there. The UTF-8 case is a parse failure before evaluation begins, and `ParseErrorCode` is the natural code.

The error hooks not running on the UTF-8 path looks like a separate defect on the same return — worth confirming whether that's deliberate.

## Scope

**Not** the after-hook path, which has the same symptom (`Reason` stays `STATIC`, `ErrorCode` empty) but isn't decidable yet: whether an after-hook error counts as abnormal execution at all is under discussion in #566. That path should be revisited once that settles.

## Related

- #542 — the `FlagMetadata` gap on these same returns. Same lines, separate fix; keeping them in different PRs.
- #537 / #538 — the same class of defect on the typed accessors' TYPE_MISMATCH paths.

fredrikblau:
## Motivation & Context

Fixes #567

Two of `Client.evaluate`'s early returns hand back an error with `Reason` and `ErrorCode` left empty on the evaluation details:

- an invalid UTF-8 flag key, which returns a parse error before the provider is bound;
- a before hook that fails.

The `NOT_READY` and `FATAL` short-circuits directly above already populate both, so a caller reading `BooleanValueDetails` (or any typed detail method) rather than the returned `error` gets nothing on these two paths. Per [1.4.7](https://openfeature.dev/specification/sections/flag-evaluation#requirement-147) the error code field MUST be populated in cases of abnormal execution, and [1.4.8](https://openfeature.dev/specification/sections/flag-evaluation#requirement-148) says the reason SHOULD indicate an error. A before hook error is abnormal execution under [4.4.7](https://openfeature.dev/specification/sections/hooks#requirement-447).

This is the same defect class as #540, fixed in #541.

## Change

Set `Reason`, `ErrorCode` and `ErrorMessage` on `evalDetails` before returning from both paths.

The UTF-8 path uses `PARSE_ERROR`, taken from the `ResolutionError` it already constructs. For the before hook path the code comes from the hook's own error when that error is a `ResolutionError`, and falls back to `GENERAL` otherwise — a hook may return an arbitrary error, and the Java and JS SDKs both preserve an OpenFeature error's own code rather than flattening every hook failure to `GENERAL`. That is what the small `errorCodeOf` helper does; happy to reduce it to a plain `GeneralCode` if you would rather not read the hook's error.

Deliberately out of scope, both called out on the issue:

- the after hook path, which has the same symptom — whether an after hook error counts as abnormal execution is still being settled in #566;
- the UTF-8 return not running the error hooks, which looks like a separate defect on the same line.

## Testing

Two new tests in `client_test.go`:

- `TestEvaluationDetails_InvalidUTF8FlagKey` — asserts the default value comes back with `ERROR`, `PARSE_ERROR` and a non-empty message, and that no provider binding happens.
- `TestEvaluationDetails_BeforeHookError` — two subtests, one with an arbitrary error expecting `GENERAL` and one with a `ResolutionError` expecting `TARGETING_KEY_MISSING`, both asserting the reason, code, message and default value, with the error and finally hooks still running.

All three fail on `main` with empty `Reason` and `ErrorCode` and pass with the change. `make test` and `make lint` are clean.

### DIFF
--- a/openfeature/client_test.go
+++ b/openfeature/client_test.go
@@ -477,6 +477,91 @@ func TestRequirement_1_4_8(t *testing.T) {
 	}
 }
 
+// An invalid UTF-8 flag key short-circuits evaluation before a provider is bound, so 1.4.7 and
+// 1.4.8 have to be satisfied on that return rather than by the provider's resolution.
+func TestEvaluationDetails_InvalidUTF8FlagKey(t *testing.T) {
+	t.Cleanup(resetSingleton)
+	mocks := hydratedMocksForClientTests(t, 0)
+	client := newClient("test-client", mocks.providerBinding, mocks.clientHandlerAPI)
+
+	defaultValue := true
+	res, err := client.evaluate(
+		t.Context(), string([]byte{0xff}), Boolean, defaultValue, EvaluationContext{}, EvaluationOptions{},
+	)
+	if err == nil {
+		t.Fatal("expected err, got nil")
+	}
+
+	if res.Value != defaultValue {
+		t.Errorf("expected default value %v, got %v", defaultValue, res.Value)
+	}
+	if res.Reason != ErrorReason {
+		t.Errorf("expected reason to be '%s', got '%s'", ErrorReason, res.Reason)
+	}
+	if res.ErrorCode != ParseErrorCode {
+		t.Errorf("expected error code to be '%s', got '%s'", ParseErrorCode, res.ErrorCode)
+	}
+	if res.ErrorMessage == "" {
+		t.Error("expected a non-empty error message")
+	}
+}
+
+// A before hook error is abnormal execution under 4.4.7, so 1.4.7 and 1.4.8 apply to the
+// evaluation details returned from that path. The code comes from the hook's own error when it
+// is a ResolutionError, and falls back to GENERAL otherwise.
+func TestEvaluationDetails_BeforeHookError(t *testing.T) {
+	for _, tt := range []struct {
+		name         string
+		hookErr      error
+		expectedCode ErrorCode
+	}{
+		{
+			name:         "arbitrary error",
+			hookErr:      errors.New("forced"),
+			expectedCode: GeneralCode,
+		},
+		{
+			name:         "resolution error",
+			hookErr:      NewTargetingKeyMissingResolutionError("no targeting key"),
+			expectedCode: TargetingKeyMissingCode,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Cleanup(resetSingleton)
+			ctrl := gomock.NewController(t)
+			mocks := hydratedMocksForClientTests(t, 1)
+			client := newClient("test-client", mocks.providerBinding, mocks.clientHandlerAPI)
+
+			mockHook := NewMockHook(ctrl)
+			mockHook.EXPECT().Before(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, tt.hookErr)
+			mockHook.EXPECT().Error(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
+			mockHook.EXPECT().Finally(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
+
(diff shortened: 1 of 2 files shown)


openfeature/client_test.go
```diff
@@ -477,6 +477,91 @@ func TestRequirement_1_4_8(t *testing.T) {
 	}
 }
 
+// An invalid UTF-8 flag key short-circuits evaluation before a provider is bound, so 1.4.7 and
+// 1.4.8 have to be satisfied on that return rather than by the provider's resolution.
+func TestEvaluationDetails_InvalidUTF8FlagKey(t *testing.T) {
+	t.Cleanup(resetSingleton)
```
erka:
```suggestion
```

openfeature/client_test.go
```diff
@@ -477,6 +477,91 @@ func TestRequirement_1_4_8(t *testing.T) {
...
+			expectedCode: GeneralCode,
+		},
+		{
+			name:         "resolution error",
+			hookErr:      NewTargetingKeyMissingResolutionError("no targeting key"),
+			expectedCode: TargetingKeyMissingCode,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Cleanup(resetSingleton)
```
erka:
```suggestion
```

openfeature/client_test.go
```diff
@@ -477,6 +477,91 @@ func TestRequirement_1_4_8(t *testing.T) {
...
+			if res.Value != defaultValue {
+				t.Errorf("expected default value %v, got %v", defaultValue, res.Value)
+			}
+			if res.Reason != ErrorReason {
+				t.Errorf("expected reason to be '%s', got '%s'", ErrorReason, res.Reason)
+			}
+			if res.ErrorCode != tt.expectedCode {
+				t.Errorf("expected error code to be '%s', got '%s'", tt.expectedCode, res.ErrorCode)
+			}
+			if res.ErrorMessage == "" {
```
erka: could we check the error?

openfeature/client_test.go
```diff
@@ -477,6 +477,91 @@ func TestRequirement_1_4_8(t *testing.T) {
...
+	if res.Value != defaultValue {
+		t.Errorf("expected default value %v, got %v", defaultValue, res.Value)
+	}
+	if res.Reason != ErrorReason {
+		t.Errorf("expected reason to be '%s', got '%s'", ErrorReason, res.Reason)
+	}
+	if res.ErrorCode != ParseErrorCode {
+		t.Errorf("expected error code to be '%s', got '%s'", ParseErrorCode, res.ErrorCode)
+	}
+	if res.ErrorMessage == "" {
```
erka: could we check the error?

PR #569, opened 2026-09-07
fix: merge before hook evaluation contexts instead of replacing them

### LINKED ISSUE #549: [BUG] before hooks replace the accumulated evaluation context instead of merging it
## Observed behavior

In `Client.beforeHooks`, each hook's returned evaluation context **wholly replaces** the accumulated one rather than merging into it:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/client.go#L776-L792

So every attribute contributed by an earlier hook is silently dropped. With two hooks each contributing one attribute, the provider sees only the last one:

```
provider saw context = map[from-hook-B:B]     // from-hook-A lost
```

The loss is twofold: because the assignment writes into `hookCtx.evaluationContext`, hook B never *sees* hook A's contribution either. That makes it a violation of both [4.3.4](https://openfeature.dev/specification/sections/hooks#requirement-434) (context returned from a before hook MUST be passed to subsequent before hooks) and [4.3.5](https://openfeature.dev/specification/sections/hooks#requirement-435) (the resulting context MUST be merged with the existing one).

## Expected Behavior

Before-hook contexts should accumulate cumulatively, so hook N observes every prior hook's contribution and the provider receives the union. All three sibling SDKs do this:

- **JS** — `packages/server/src/client/internal/open-feature-client.ts`: `Object.assign(accumulatedContext, hookResult)`, then assigns back into `hookContext.context`.
- **Java** — `HookSupport.java` calls `putHookContext`, and `LayeredEvaluationContext` *appends* to an `ArrayList<EvaluationContext>` rather than replacing.
- **Python** — `openfeature/hook/_hook_support.py`: `reduce(lambda a, b: a.merge(b), filtered_hooks)`.

Worth noting this repo already has a correct implementation in the `multi` package, which accumulates into a slice and merges all of them:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/multi/isolation.go#L241-L256

So the two hook engines in the repo currently disagree.

## Steps to reproduce

```go
// ctxHook.Before returns of.NewTargetlessEvaluationContext(map[string]any{h.key: h.val})
client.AddHooks(
	ctxHook{key: "from-hook-A", val: "A"},
	ctxHook{key: "from-hook-B", val: "B"},
)
client.Boolean(ctx, "some-flag", false, of.EvaluationContext{})

// provider receives map[from-hook-B:B]
// expected: both from-hook-A and from-hook-B
```

fredrikblau:
## Motivation & Context

Fixes #549

In `Client.beforeHooks`, each hook's returned evaluation context wholly replaces the accumulated one:

```go
if resultEvalCtx != nil {
    hookCtx.evaluationContext = *resultEvalCtx
}
```

So with two hooks each contributing an attribute, everything the earlier hook added is dropped, and the later hook never sees it either. The provider ends up with only the last hook's contribution.

That misses [4.3.3](https://openfeature.dev/specification/sections/hooks#requirement-433) — a context returned from a before hook MUST be passed to subsequent before hooks — since what gets passed on is no longer the accumulation of what came before.

The other SDKs all accumulate: JS `Object.assign(accumulatedContext, hookResult)`, Java's `LayeredEvaluationContext` appends rather than replaces, Python reduces with `merge`. This repo's own `multi` package already does it correctly in `hookIsolator.beforeHooks`, so the two hook engines here currently disagree.

## Change

Merge the hook's result into the accumulated context instead of overwriting it. `mergeContexts` gives precedence to its first argument, so passing the hook result first keeps the newest contribution winning while retaining everything earlier hooks added — the same precedence as the JS `Object.assign`.

The existing merge with the invocation context on the way out is unchanged, so before-hook values still outrank invocation, client and API context.

## Testing

- New `TestBeforeHooksAccumulateEvaluationContext`: two hooks each contribute one attribute; asserts the second hook observes the first one's contribution and that the provider receives both. On `main` the provider sees only `map[from-hook-b:b]`.
- `TestRequirement_4_3_3` asserted the old replace behaviour — that the second hook receives exactly the first hook's returned context, with the invocation context's attributes dropped. Updated to expect the merged context, which is what the requirement it covers actually calls for. Its provider-level assertion is unchanged and still passes.

`make test` and `make lint` are clean.

### DIFF
--- a/openfeature/hooks_test.go
+++ b/openfeature/hooks_test.go
@@ -235,9 +235,10 @@ func TestRequirement_4_3_3(t *testing.T) {
 		TargetingKey: "mockHook1",
 	})
 
-	// assert that the evaluation context returned by the first hook is passed into the second hook
+	// assert that the evaluation context returned by the first hook is merged into the context
+	// passed to the second hook, rather than replacing it
 	hook2Ctx := hook1Ctx
-	hook2Ctx.evaluationContext = *hook1EvalCtxResult
+	hook2Ctx.evaluationContext = mergeContexts(*hook1EvalCtxResult, evalCtx)
 	mockHook2.EXPECT().Before(gomock.Any(), hook2Ctx, gomock.Any())
 
 	mockHook1.EXPECT().After(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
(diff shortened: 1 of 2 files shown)


openfeature/hooks_test.go
```diff
@@ -252,6 +253,59 @@ func TestRequirement_4_3_3(t *testing.T) {
 }
 
 // When `before` hooks have finished executing, any resulting `evaluation context` MUST be merged with the existing
+
+// A before hook's evaluation context is merged into the accumulated one rather than replacing it,
+// so each hook observes every prior hook's contribution and the provider receives the union.
+func TestBeforeHooksAccumulateEvaluationContext(t *testing.T) {
+	t.Cleanup(resetSingleton)
```
erka: could you please use new isolated API? I hope we will be able to run tests in parallel.

erka (approved):
Thanks @fredrikblau, one nit from me.

sahidvelji (approved):

PR #571, opened 2026-09-07
docs: align genericResolve godoc with the implementation

alorentiar:
Follow-up from the review on #565: the godoc said int values convert to float64 as a primary form, but the canonical mapping is int* to int64 and float32 to float64.


erka:
Hey @alorentiar 

I think that the godoc is correct right now in the main. int* or float32 will be converted to float64.

alorentiar:
Ah, that makes sense. int values do convert to float64 in the float branch, which was the point of the original change in the first place, so the existing wording was accurate. I misread it, sorry for the noise. Closing this.

PR #573, opened 2026-09-07
fix: copy the caller's map in NewHookHints

### LINKED ISSUE #562: [BUG] NewHookHints stores the caller's map without copying
## Observed behavior

`NewHookHints` stores the caller's map by reference, with no copy:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/hooks.go#L15-L29

So the caller can mutate the hints after construction, and every hook observes the change:

```
HookHints.Value("k") after the caller mutated its own map = "after"   (want "before")
```

The `Value` doc comment — "Maintains immutability of the map" — is misleading as a result: it protects the *reference*, not the contents.

`NewEvaluationContext` in the same package does copy, for exactly this reason:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/evaluation_context.go#L38-L51

## Expected Behavior

A defensive `maps.Copy` in `NewHookHints`, matching `NewEvaluationContext`, so the doc comment becomes accurate.

[4.5.3](https://openfeature.dev/specification/sections/hooks#requirement-453) is unconditional: *"The hook **MUST NOT** alter the `hook hints` structure."* A copy makes that enforceable rather than advisory.

(Citing 4.5.3 rather than [4.2.2.1](https://openfeature.dev/specification/sections/hooks#conditional-requirement-4221) — "hook hints MUST be immutable" — since the latter is conditional on the language supporting a mechanism for marking data as immutable, which Go arguably doesn't.)

The siblings all defend here:

- **Java** — `Collections.unmodifiableMap(hookHints)`.
- **JS** — `Object.freeze(options.hookHints)`.
- **Python** — hints typed as a read-only `Mapping`, with a `MappingProxyType` helper shipped.

## Steps to reproduce

```go
src := map[string]any{"k": "before"}
hh := of.NewHookHints(src)
src["k"] = "after"
fmt.Println(hh.Value("k")) // "after"
```

kripa-sindhu-007:
This PR

- copies the caller's map in NewHookHints, matching NewEvaluationContext in the same package
- moves the immutability claim off Value onto the constructor, and states that the copy is shallow
- extends TestRequirement_4_2_2_1 with an assertion that actually fails without the fix

NewHookHints stored the map by reference, so the caller could mutate the hints after construction and every hook would observe the change:

Value("k") = after                 (want "before")
Value("added-later") = surprise    (want nil)

Related Issues

Fixes #562

Notes

- The assertion goes in TestRequirement_4_2_2_1 rather than 4.5.3, per the discussion on the issue: immutability of the hints structure is the property the copy establishes, while 4.5.3 constrains the hook rather than the caller. TestRequirement_4_5_3 is left as it is.
- Both of those tests previously asserted only that mapOfHints is unexported. That keeps the reference out of reach from outside the package, but not the contents, since the caller still holds the map it passed in — so neither test could fail today, and neither would have caught this.
- The copy is shallow, so values that are themselves mutable stay shared. That matches Collections.unmodifiableMap in the Java SDK and Object.freeze in the JS SDK, so it is parity rather than a shortfall. I verified the boundary behaves as documented but did not add a test for it, since asserting it would lock the shallowness in.
- The doc comment on Value claimed to maintain immutability of the map, which only became true once the constructor copies. Moved rather than deleted.

How to test

go test -tags testtools -race -count=1 -run 'TestRequirement_4_2_2_1' ./openfeature/

### DIFF
--- a/openfeature/hooks.go
+++ b/openfeature/hooks.go
@@ -1,6 +1,9 @@
 package openfeature
 
-import "context"
+import (
+	"context"
+	"maps"
+)
 
 // Hook allows application developers to add arbitrary behavior to the flag evaluation lifecycle.
 // They operate similarly to middleware in many web frameworks.
@@ -18,12 +21,19 @@ type HookHints struct {
 }
 
 // NewHookHints constructs HookHints
+//
+// mapOfHints - arbitrary data made available to hooks. The map is copied, so
+// changes the caller makes to it afterwards are not reflected in the hints.
+// The copy is shallow: values that are themselves mutable remain shared.
 func NewHookHints(mapOfHints map[string]any) HookHints {
-	return HookHints{mapOfHints: mapOfHints}
+	// copy hints to new map to avoid reference being externally available, thereby enforcing immutability
+	hints := make(map[string]any, len(mapOfHints))
+	maps.Copy(hints, mapOfHints)
+
+	return HookHints{mapOfHints: hints}
 }
 
 // Value returns the value at the given key in the underlying map.
-// Maintains immutability of the map.
 func (h HookHints) Value(key string) any {
 	return h.mapOfHints[key]
 }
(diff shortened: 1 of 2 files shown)


erka (approved):

sahidvelji (approved):

erka:
Thanks @kripa-sindhu-007

PR #574, opened 2026-09-07
fix: resolve disabled flags without an error

### LINKED ISSUE #552: [BUG] memprovider: a disabled flag surfaces a GENERAL error instead of reason DISABLED
## Observed behavior

`InMemoryFlag.Resolve` attaches a resolution **error** to the disabled case:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/memprovider/in_memory_provider.go#L217-L226

`ProviderResolutionDetail.Error()` returns non-nil whenever a code is set:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/provider.go#L171-L176

So `Client.evaluate` treats a disabled flag as abnormal execution, **overwrites** `DisabledReason` with `ErrorReason`, and hands the application a GENERAL error:

```
disabled flag: value=false reason="ERROR" errorCode="GENERAL"
               errorMessage="flag is disabled" err=error code: GENERAL: flag is disabled
```

## Expected Behavior

Reason `DISABLED`, the default value, and **no** error.

That's what the Gherkin suite this SDK is required to run asserts — `specification/assets/gherkin/evaluation_v2.feature`, scenario tagged `@spec-1.4.7` ("DISABLED reason"):

```gherkin
Then the resolved details value should be "<default>"
And the reason should be "DISABLED"
```

…across `boolean-`, `string-`, `integer-`, `float-` and `object-disabled-flag`.

Both siblings return DISABLED with no error:

- **Java** `InMemoryProvider` builds `.reason(Reason.DISABLED.name()).value(defaultValue)` and returns normally, with no throw.
- **JS** `in-memory-provider.ts` returns `{ value: defaultValue, reason: StandardResolutionReasons.DISABLED }`.

The fix is to drop the `ResolutionError` from that branch, keeping `Reason: DisabledReason`.

Note the existing `TestInMemoryProvider_Disabled` passes because it calls the **provider** directly and never goes through the client, so it doesn't observe the reason being overwritten — worth extending to a client-level assertion.

## Steps to reproduce

```go
p := memprovider.NewInMemoryProvider(map[string]memprovider.InMemoryFlag{
	"disabled-flag": {
		Key:            "disabled-flag",
		State:          memprovider.Disabled,
		DefaultVariant: "on",
		Variants:       map[string]any{"on": true, "off": false},
	},
})
// SetProviderAndWait(p), then:
d, err := client.BooleanValueDetails(ctx, "disabled-flag", false, of.EvaluationContext{})
// d.Reason == "ERROR" (want "DISABLED"), d.ErrorCode == "GENERAL", err != nil (want nil)
```

kripa-sindhu-007:
This PR

- drops the ResolutionError from InMemoryFlag.Resolve's disabled branch, so a disabled flag resolves to the default with reason DISABLED and no error
- extends TestInMemoryProvider_Disabled with an error assertion and a client-level subtest

ProviderResolutionDetail.Error() is non-nil whenever a code is set, so Client.evaluate treated a disabled flag as abnormal execution, overwrote DisabledReason with ErrorReason and returned a GENERAL error to the caller:

value=false reason="ERROR" errorCode="GENERAL" errorMessage="flag is disabled"
err=error code: GENERAL: flag is disabled

After the fix, reason DISABLED, the default value, and no error.

Related Issues

Fixes #552

Notes

- Dropping the ResolutionError is the whole fix. Resolve already returns defaultValue on that branch and genericResolve asserts it against itself, so the value and the reason come out right on their own.
- The existing test disabled flag subtest asserts only Value and Reason, both of which are already correct at the provider level, so it passes unchanged before and after this change. I have added an error assertion to it, otherwise nothing in the suite would catch a regression here.
- The client-level subtest is the one that matters, since the reason is overwritten in Client.evaluate and a provider-level assertion cannot observe that. It uses isolated.NewAPI() so the global singleton is left alone — this package's tests have not driven the client before, and I did not want to introduce shared provider state into them.
- Worth recording for anyone reading the issue: the DISABLED scenarios cited there are in evaluation_v2.feature in the spec repo. This repo's test-harness submodule is the flagd harness, and e2e runs only evaluation.feature from it, which has no disabled-flag scenarios. So e2e does not cover this today, which is likely why it went unnoticed. The parity argument with the Java and JS in-memory providers, both of which return DISABLED with no error, is what the fix rests on.
- One line outside the strict fix: the reason assertion's failure message printed ErrorReason as its expectation while checking DisabledReason. It is inside the block being changed, so I corrected it rather than leave a misleading message next to a new assertion. Happy to split it out if you would rather.
- On sequencing with #539: the test uses := throughout and never names memprovider.InMemoryProvider explicitly, so it survives whichever of the two merges first.

How to test

go test -tags testtools -race -count=1 -run 'TestInMemoryProvider_Disabled' ./openfeature/memprovider/

### DIFF
--- a/openfeature/memprovider/in_memory_provider.go
+++ b/openfeature/memprovider/in_memory_provider.go
@@ -230,8 +230,7 @@ func (flag *InMemoryFlag) Resolve(defaultValue any, flatCtx openfeature.Flattene
 	// check the state
 	if flag.State == Disabled {
 		return defaultValue, openfeature.ProviderResolutionDetail{
-			ResolutionError: openfeature.NewGeneralResolutionError("flag is disabled"),
-			Reason:          openfeature.DisabledReason,
+			Reason: openfeature.DisabledReason,
 		}
 	}
 
--- a/openfeature/memprovider/in_memory_provider_test.go
+++ b/openfeature/memprovider/in_memory_provider_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/open-feature/go-sdk/openfeature"
+	"github.com/open-feature/go-sdk/openfeature/isolated"
 )
 
 func TestInMemoryProvider_boolean(t *testing.T) {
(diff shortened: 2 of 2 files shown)


sahidvelji (approved):

erka (approved):
Thank you @kripa-sindhu-007

PR #575, opened 2026-09-07
fix: return an empty flag metadata record on evaluate's early returns

### LINKED ISSUE #542: [BUG] evaluate() early returns leave FlagMetadata nil instead of an empty record
## Observed behavior

`Client.evaluate` initializes `evalDetails` without a `FlagMetadata`, so every early return yields `FlagMetadata == nil`: the NOT_READY and FATAL short-circuits, the UTF-8 flag key check, and the before/after-hook error returns.

The normal provider path doesn't have this problem — `ProviderResolutionDetail.ResolutionDetail()` substitutes `FlagMetadata{}` when the provider supplies none, and `TestRequirement_1_4_13/No Metadata` asserts it. So the guarantee holds there but is dropped on `evaluate`'s early returns.

## Expected Behavior

Per [spec 1.4.14](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1414), flag metadata "**MUST** contain an empty record" when the provider sets none.

Reads from a nil map are safe, so this isn't a panic risk for read-only consumers. It matters for code comparing against `FlagMetadata{}`, serialization where nil and `{}` differ, or writes into the map — `multi` does `resolution.FlagMetadata[MetadataProviderName] = ...`, and assigning into a nil map panics.

Likely fix is to initialize it once at the top of `evaluate` alongside `FlagKey`/`FlagType`; the later `evalDetails.ResolutionDetail = resolution.ResolutionDetail()` still overwrites wholesale, so provider-supplied metadata keeps winning. `multi`'s `hookIsolator.evaluate` has the same gap on its before-hook error path.

## Steps to reproduce

```go
// with the provider in NOT_READY
details, _ := client.BooleanValueDetails(ctx, "a-flag", true, openfeature.EvaluationContext{})
fmt.Println(details.FlagMetadata == nil) // true; expected empty FlagMetadata{}
```

## Notes

Surfaced while reviewing #541, but pre-existing and broader than that PR. May belong under #526.

kripa-sindhu-007:
## This PR

- initialises `FlagMetadata` once at the top of `evaluate`, so the early returns yield an empty record instead of nil
- fixes the same gap in `multi`'s `hookIsolator`, which set `FlagMetadata: nil` explicitly on its before-hook error path
- adds subtests covering all four returns, and renames the test they live in to `TestRequirement_1_4_14`

Four of `evaluate`'s returns never reach a provider resolution, so they had no metadata to inherit and handed back nil: the invalid UTF-8 flag key check, the NOT_READY and FATAL short-circuits, and the before-hook error. Per 1.4.14 the field must contain an empty record when the provider sets none.

### Related Issues

Fixes #542

### Notes

- Initialising at the top rather than per-path is deliberate. `evalDetails.ResolutionDetail = resolution.ResolutionDetail()` further down assigns wholesale, so provider-supplied metadata still wins and nothing on the normal path changes.
- The after-hook and resolution-error returns already yielded an empty record, because both sit below that assignment. So it is four returns, not five — the issue body says five.
- Scope is metadata only. The error-code gap on these same returns is #567, being handled in #568.
- On overlap: #568 edits the UTF-8 and before-hook return blocks, and #572 is in `multi/isolation.go`. Neither touches metadata. I have kept the new tests as subtests of `TestRequirement_1_4_14` rather than new top-level functions, since #568 is adding those to `client_test.go`, and extended the existing `Test_HookIsolator_ExecutesHooksDuringEvaluation_BeforeErrorAbortsExecution` rather than adding a test to `isolation_test.go`, which #572 also touches.
- The test these subtests live in was named `TestRequirement_1_4_13` while its doc comment quoted 1.4.14. Renamed to `TestRequirement_1_4_14` per review: 1.4.13 is the error message MAY, and the empty-record MUST is 1.4.14. That is one entry off the numbering drift tracked in #526.
- The new assertions use `%#v`. With `%v` a nil map and an empty map both print as `map[]`, which would make the one distinction these tests exist to check invisible in a failure. The pre-existing `No Metadata` subtest has the same issue; I left it alone as out of scope.

### How to test

```
go test -tags testtools -race -count=1 -run 'TestRequirement_1_4_14' ./openfeature/
go test -tags testtools -race -count=1 -run 'Test_HookIsolator' ./openfeature/multi/
```

### DIFF
--- a/openfeature/client_test.go
+++ b/openfeature/client_test.go
@@ -695,11 +695,11 @@ func TestRequirement_1_4_12(t *testing.T) {
 	}
 }
 
-// Requirement_1_4_13
+// Requirement_1_4_14
 // If the `flag metadata` field in the `flag resolution` structure returned by the configured `provider` is set,
 // the `evaluation details` structure's `flag metadata` field MUST contain that value. Otherwise,
 // it MUST contain an empty record.
-func TestRequirement_1_4_13(t *testing.T) {
+func TestRequirement_1_4_14(t *testing.T) {
 	flagKey := "flag-key"
 	evalCtx := EvaluationContext{}
 	flatCtx := flattenContext(evalCtx)
--- a/openfeature/client_test.go
+++ b/openfeature/client_test.go
@@ -730,6 +730,112 @@ func TestRequirement_1_4_13(t *testing.T) {
 		}
 	})
 
+	// early returns that never reach a provider resolution, see #542
+	t.Run("No metadata on the invalid UTF-8 flag key return", func(t *testing.T) {
+		t.Cleanup(resetSingleton)
+
+		mocks := hydratedMocksForClientTests(t, 0)
+		client := newClient("test-client", mocks.providerBinding, mocks.clientHandlerAPI)
+
+		evDetails, err := client.BooleanValueDetails(t.Context(), "invalid\xf0\x28", true, EvaluationContext{})
+		if err == nil {
+			t.Error("expected an error for an invalid UTF-8 flag key, got nil")
+		}
+		if !reflect.DeepEqual(evDetails.FlagMetadata, FlagMetadata{}) {
+			// %#v: nil and an empty map both print as "map[]"
+			t.Errorf("expected %#v, got %#v", FlagMetadata{}, evDetails.FlagMetadata)
+		}
+	})
+
+	t.Run("No metadata on the before hook error return", func(t *testing.T) {
+		t.Cleanup(resetSingleton)
+
+		mocks := hydratedMocksForClientTests(t, 1)
+		client := newClient("test-client", mocks.providerBinding, mocks.clientHandlerAPI)
+
+		mockHook := NewMockHook(gomock.NewController(t))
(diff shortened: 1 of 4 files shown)


openfeature/client_test.go
sahidvelji:
```suggestion
// Requirement_1_4_14
// If the `flag metadata` field in the `flag resolution` structure returned by the configured `provider` is set,
// the `evaluation details` structure's `flag metadata` field MUST contain that value. Otherwise,
// it MUST contain an empty record.
func TestRequirement_1_4_14(t *testing.T) {
```

This is pre-existing, but the name of the test does not match the intent.

1.4.13 is the error message MAY requirement. The empty-record MUST is 1.4.14:

https://openfeature.dev/specification/sections/flag-evaluation/#requirement-1414

sahidvelji (approved):

openfeature/client_test.go
kripa-sindhu-007: Applied in 62a7ba0

sahidvelji (approved):

erka (approved):

PR #576, opened 2026-09-07
fix: omit empty feature flag context id

### LINKED ISSUE #563: [BUG] telemetry: feature_flag.context.id is emitted as an empty string when no targeting key is set
## Observed behavior

`CreateEvaluationEvent` assigns the context ID unconditionally from the targeting key, then lets flag metadata override it:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/telemetry/telemetry.go#L66-L70

With no targeting key and no `contextId` metadata, the event carries `feature_flag.context.id: ""` rather than omitting the attribute.

## Expected Behavior

[Appendix D](https://openfeature.dev/specification/appendix-d) types this attribute as `string` and says the targeting key is used only *"if not available"* — i.e. as a fallback for `contextId`.

All three siblings guard the assignment, and all three check **metadata first**, falling back to the targeting key:

- **JS** — `if (contextId)`.
- **Python** — `if context_id:`.
- **Java** — `if (contextId != null) … else …` (though Java's else-branch can still pass a null targeting key, so it's only partially better).

Go's inverted precedence is what produces the empty string; matching the sibling ordering fixes both the empty-attribute case and the precedence difference at once.

## Steps to reproduce

```go
// empty targeting key, no contextId in flag metadata
event := telemetry.CreateEvaluationEvent(hookContext, details)
v, ok := event.Attributes["feature_flag.context.id"]
fmt.Println(ok, fmt.Sprintf("%q", v)) // true ""  — want the attribute omitted
```

vbrekher:
Fixes #563.

`feature_flag.context.id` was being added as an empty string when there was no targeting key.

This uses `contextId` from flag metadata first, falls back to the targeting key, and leaves the attribute out when neither is available.

Added tests for these cases.

### DIFF
--- a/openfeature/telemetry/telemetry_context_id_test.go
+++ b/openfeature/telemetry/telemetry_context_id_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/open-feature/go-sdk/openfeature"
 )
 
+// TestCreateEvaluationEvent_ContextID verifies context ID precedence, fallback, and omission.
 func TestCreateEvaluationEvent_ContextID(t *testing.T) {
 	tests := []struct {
 		name          string
@@ -40,9 +41,35 @@ func TestCreateEvaluationEvent_ContextID(t *testing.T) {
 			wantAttribute: true,
 		},
 		{
+			name:         "nil metadata context uses targeting key fallback",
+			targetingKey: "targeting-context",
+			flagMetadata: openfeature.FlagMetadata{
+				flagMetaContextIDKey: nil,
+			},
+			wantContextID: "targeting-context",
+			wantAttribute: true,
+		},
+		{
+			name:         "non-string metadata context uses targeting key fallback",
+			targetingKey: "targeting-context",
+			flagMetadata: openfeature.FlagMetadata{
+				flagMetaContextIDKey: true,
+			},
+			wantContextID: "targeting-context",
+			wantAttribute: true,
+		},
+		{
 			name:          "context attribute is omitted when unavailable",
 			targetingKey:  "",
 			flagMetadata:  openfeature.FlagMetadata{},
+			wantAttribute: false,
+		},
+		{
+			name:         "invalid metadata is omitted without targeting key",
+			targetingKey: "",
+			flagMetadata: openfeature.FlagMetadata{
+				flagMetaContextIDKey: nil,
+			},
 			wantAttribute: false,
 		},
 	}
--- a/openfeature/telemetry/telemetry_context_id_test.go
+++ b/openfeature/telemetry/telemetry_context_id_test.go
@@ -1,0 +1,86 @@
+package telemetry
+
+import (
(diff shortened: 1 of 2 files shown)


openfeature/telemetry/telemetry_context_id_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}{
+		{
+			name:         "metadata takes precedence",
+			targetingKey: "targeting-context",
+			flagMetadata: openfeature.FlagMetadata{
+				flagMetaContextIDKey: "metadata-context",
+			},
+			wantContextID: "metadata-context",
+			wantAttribute: true,
+		},
```
sahidvelji, line 26: Nit: one uncovered combination — non-empty metadata `contextId` with an empty targeting key. Every metadata-wins case here also supplies a targeting key, so nothing asserts that metadata works standalone. Since the branch under test is `if metadata { … } else if targetingKey { … }`, a bug that accidentally required a targeting key for the metadata path would still pass. One extra row closes it:

```go
{
    name:         "metadata is used without a targeting key",
    targetingKey: "",
    flagMetadata: openfeature.FlagMetadata{
        flagMetaContextIDKey: "metadata-context",
    },
    wantContextID: "metadata-context",
    wantAttribute: true,
},
```

*AI-assisted — Tool: Claude Code; model: anthropic/claude-5-opus[1m]; version: 2.1.219.*

openfeature/telemetry/telemetry_context_id_test.go
sahidvelji, line 1: Nit: the package's existing tests all live in `telemetry_test.go` and are named `TestCreateEvaluationEvent_<spec_requirement>_<Description>` (e.g. `TestCreateEvaluationEvent_1_4_14_WithFlagMetaData`), tying each test back to the requirement it covers. Consider folding this table into `telemetry_test.go` as `TestCreateEvaluationEvent_1_4_14_ContextID` for consistency. That would also surface the small overlap with `TestCreateEvaluationEvent_1_4_14_WithFlagMetaData`, which already asserts the metadata-precedence case.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-5-opus[1m]; version: 2.1.219.*

sahidvelji (approved):

vbrekher:
Good catches. I folded the context-ID table into telemetry_test.go as TestCreateEvaluationEvent_1_4_14_ContextID and added the missing case where metadata supplies a contextId with no targeting key. The nil, non-string, fallback and omission cases are still covered. Focused telemetry tests, the tagged OpenFeature regression, the full short test suite and lint are clean.

erka:
@vbrekher could you please sign the commit to resolve DCO issue?

kripa-sindhu-007 (approved):

PR #578, opened 2026-09-08
fix(multi): stop unchecked type assertions panicking out of flag evaluation

### LINKED ISSUE #546: [BUG] multiprovider: unchecked type assertions panic out of flag evaluation
## Observed behavior

The typed evaluation methods on `multi.Provider` assert the strategy's result without checking:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/multi/multiprovider.go#L308-L342

The same pattern appears in `isolation.go` (`hookIsolator`'s evaluation methods) and in `strategies.go`:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/multi/strategies.go#L158-L162

Three distinct panics are reachable, one of which needs no custom code at all.

**1. `StrategyComparison` with a `nil` object default.** With no custom comparator, `evaluateComparison` calls `reflect.TypeOf(defaultValue)` → `nil`, then `t.Comparable()` → nil-pointer dereference:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/multi/comparison_strategy.go#L86-L101

Confirmed escaping through the public API:

```
PANIC escaped ObjectEvaluation with nil default:
  runtime error: invalid memory address or nil pointer dereference
  multi.evaluateComparison[...].func1   comparison_strategy.go:93
  multi.(*Provider).ObjectEvaluation    multiprovider.go:349
```

**2. A custom strategy returning a zero-value result.** `WithCustomStrategy` is a public, documented extension point, and returning a zero-value `GenericResolutionDetail` is a natural way to express "give up": `interface conversion: multi.FlagTypes is nil, not bool`.

**3. Any strategy path where an inner provider returns a nominal result with `Value: nil`.**

There is no `recover()` anywhere on the evaluation path — the only one in the SDK guards event callbacks in `event_executor.go`.

## Expected Behavior

Per [Requirement 1.4.10](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1410): *"Methods, functions, or operations on the client **MUST NOT** throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution."*

Suggested shape:

- Use checked assertions (`v, ok := res.Value.(bool)`) in the typed evaluation methods, falling back to the default value with `TypeMismatchCode` when the assertion fails.
- Guard the `reflect.TypeOf(defaultValue) == nil` case in `evaluateComparison` and `defaultComparator`.

## Steps to reproduce

```go
mp, _ := multi.NewProvider(multi.StrategyComparison,
	multi.WithProvider("a", memprovider.NewInMemoryProvider(map[string]memprovider.InMemoryFlag{})),
	multi.WithProvider("b", memprovider.NewInMemoryProvider(map[string]memprovider.InMemoryFlag{})))

mp.InitWithContext(context.Background(), of.EvaluationContext{})
mp.ObjectEvaluation(context.Background(), "flag", nil, of.FlattenedContext{}) // panics
```

The same panic surfaces through `client.ObjectValue(ctx, "flag", nil, evalCtx)` with such a provider registered.

kripa-sindhu-007:
## This PR

- guards `reflect.TypeOf(defaultValue)` against a nil default in `evaluateComparison`, so a nil object default resolves instead of panicking
- returns the caller's default when a strategy leaves `Value` unset, at the four typed accessors on `Provider` and in the object branch of `Evaluate`
- does the same for `Provider.ObjectEvaluation`, which asserted nothing and passed the nil straight through to the client
- reports a wrong-typed strategy value as `TYPE_MISMATCH` with an `ERROR` reason, rather than returning the default under a success reason
- adds coverage for each path

### Related Issues

Fixes #546

### Notes

All three panics in the issue reproduce at `9853ca0`, and the last two share a root cause, an unchecked type assertion on a value the strategy may leave nil, so they are fixed by one unexported helper rather than three separate guards:

| # | Panic | Site |
|---|---|---|
| 1 | `reflect.TypeOf(nil).Comparable()` nil deref | `comparison_strategy.go:93` |
| 2 | `interface conversion: multi.FlagTypes is nil, not bool` | `multiprovider.go:312` |
| 3 | `interface conversion: interface is nil, not multi.FlagTypes` | `strategies.go:161` |

Claim 1 is broader than the issue states. The nil default escapes the client too, not just the provider, which makes it a 1.4.10 violation rather than a `multi`-internal panic. Reproduced through `client.ObjectValue`.

The open question about `TYPE_MISMATCH` is settled: @erka  and @sahidvelji both preferred it, so the fallback no longer hides behind the provider's success reason. `resolveTyped` splits the two cases. A nil value means the strategy resolved nothing, so the caller's default stands on its own. A value of the wrong type is a bug in the strategy, so it comes back as the default plus `TYPE_MISMATCH` with an `ERROR` reason. Through the client that reads:

    provider -> Value=true Reason="ERROR" Err="TYPE_MISMATCH: strategy resolved string, expected bool"
    client   -> Value=true Reason="ERROR" ErrorCode="TYPE_MISMATCH"

Review also turned up a fourth site the issue does not list. `Provider.ObjectEvaluation` asserted nothing at all, so an unset strategy value reached `Client.ObjectValue` as nil with no error and no reason to go with it:

    mp.ObjectEvaluation       -> Value=<nil>  Reason=""
    client.ObjectValue        -> value=<nil>  err=<nil>
    client.ObjectValueDetails -> value=<nil>  reason=""  code=""  err=<nil>

`BooleanEvaluation` on the same strategy returns the default, so this is the same 1.4.10 gap as the other claims, only silent. It goes through `resolveTyped` with the rest.

Each new test was checked against a reverted fix, so they fail for the right reason: the panic cases die with the message in the table above, and the two added in review return `nil` and a `STATIC` reason respectively.

### How to test

```go
// claim 1 — a nil object default under StrategyComparison
mp, _ := multi.NewProvider(multi.StrategyComparison,
      multi.WithProvider("a", someProvider), multi.WithProvider("b", otherProvider))
mp.InitWithContext(context.Background(), of.EvaluationContext{})
mp.ObjectEvaluation(context.Background(), "flag", nil, of.FlattenedContext{}) // panics on main

// the same through the client, which is the 1.4.10 case
api := isolated.NewAPI()
api.SetProviderAndWait(context.Background(), mp)
api.NewClient().ObjectValue(context.Background(), "flag", nil, of.EvaluationContext{}) // panics on main

// claim 2 — a custom strategy that gives up
strategy := func(providers []multi.NamedProvider) multi.StrategyFn[multi.FlagTypes] {
      return func(context.Context, string, multi.FlagTypes, of.FlattenedContext) of.GenericResolutionDetail[multi.FlagTypes] {
              return of.GenericResolutionDetail[multi.FlagTypes]{}
      }
}
mp2, _ := multi.NewProvider(multi.StrategyCustom,
      multi.WithCustomStrategy(strategy), multi.WithProvider("p", someProvider))
mp2.BooleanEvaluation(context.Background(), "flag", true, of.FlattenedContext{}) // panics on main

// claim 3 — an inner provider resolving an object flag to nil
mp3, _ := multi.NewProvider(multi.StrategyFirstMatch, multi.WithProvider("nil-object", nilObjectProvider{}))
mp3.ObjectEvaluation(context.Background(), "flag", map[string]any{"d": true}, of.FlattenedContext{}) // panics on main

// the fourth site — the same giving-up strategy, reaching the object accessor
mp2.ObjectEvaluation(context.Background(), "flag", map[string]any{"kept": true}, of.FlattenedContext{})
// on main: Value=<nil>, no error, no reason. client.ObjectValue returns nil too.

// a strategy that resolves the wrong type
wrongType := func(providers []multi.NamedProvider) multi.StrategyFn[multi.FlagTypes] {
      return func(context.Context, string, multi.FlagTypes, of.FlattenedContext) of.GenericResolutionDetail[multi.FlagTypes] {
              return of.GenericResolutionDetail[multi.FlagTypes]{
                      Value:                    "not-a-bool",
                      ProviderResolutionDetail: of.ProviderResolutionDetail{Reason: of.StaticReason},
              }
      }
}
mp4, _ := multi.NewProvider(multi.StrategyCustom,
      multi.WithCustomStrategy(wrongType), multi.WithProvider("p", someProvider))
mp4.BooleanEvaluation(context.Background(), "flag", true, of.FlattenedContext{})
// on main: Value=true with Reason=STATIC and no error. Now: Reason=ERROR, TYPE_MISMATCH.
```

### DIFF
--- a/openfeature/multi/strategies.go
+++ b/openfeature/multi/strategies.go
@@ -109,6 +109,16 @@ func mergeFlagMeta(tags ...of.FlagMetadata) of.FlagMetadata {
 	}
 }
 
+// valueOrDefault returns value when it holds the expected type, and the caller's default
+// otherwise. A strategy may leave Value unset, and a nil has no type to assert, see #546
+func valueOrDefault[T any](value any, defaultValue T) T {
+	if v, ok := value.(T); ok {
+		return v
+	}
+
+	return defaultValue
+}
+
 // BuildDefaultResult should be called when a [StrategyFn] is in a failure state and needs to return a default value.
 // This method will build a resolution detail with the internal provided error set. This method is exported for those
 // writing their own custom [StrategyFn].
@@ -158,7 +168,7 @@ func Evaluate[T FlagTypes](ctx context.Context, provider NamedProvider, flag str
 	default:
 		res := provider.ObjectEvaluation(ctx, flag, defaultVal, flatCtx)
 		resolution.ProviderResolutionDetail = res.ProviderResolutionDetail
-		resolution.Value = res.Value.(T)
+		resolution.Value = valueOrDefault(res.Value, defaultVal)
 	}
 
 	if resolution.FlagMetadata == nil {
--- a/openfeature/multi/strategies.go
+++ b/openfeature/multi/strategies.go
@@ -2,6 +2,7 @@ package multi
 
 import (
 	"context"
+	"fmt"
 	"maps"
 	"regexp"
 	"strings"
@@ -109,14 +110,21 @@ func mergeFlagMeta(tags ...of.FlagMetadata) of.FlagMetadata {
 	}
 }
 
-// valueOrDefault returns value when it holds the expected type, and the caller's default
-// otherwise. A strategy may leave Value unset, and a nil has no type to assert, see #546
-func valueOrDefault[T any](value any, defaultValue T) T {
+// resolveTyped builds the resolution for value, falling back to the caller's default.
+// A nil value means the strategy resolved nothing, so the default stands on its own.
+// A value of the wrong type is a strategy bug and is reported as a type mismatch rather
+// than returned under the provider's success reason.
+func resolveTyped[T any](value any, detail of.ProviderResolutionDetail, defaultValue T) of.GenericResolutionDetail[T] {
+	typed := of.GenericResolutionDetail[T]{Value: defaultValue, ProviderResolutionDetail: detail}
 	if v, ok := value.(T); ok {
-		return v
+		typed.Value = v
+	} else if value != nil {
+		typed.ResolutionError = of.NewTypeMismatchResolutionError(
+			fmt.Sprintf("strategy resolved %T, expected %T", value, defaultValue))
+		typed.Reason = of.ErrorReason
 	}
 
-	return defaultValue
+	return typed
 }
 
 // BuildDefaultResult should be called when a [StrategyFn] is in a failure state and needs to return a default value.
@@ -167,8 +175,7 @@ func Evaluate[T FlagTypes](ctx context.Context, provider NamedProvider, flag str
 		resolution.Value = any(res.Value).(T)
 	default:
 		res := provider.ObjectEvaluation(ctx, flag, defaultVal, flatCtx)
-		resolution.ProviderResolutionDetail = res.ProviderResolutionDetail
-		resolution.Value = valueOrDefault(res.Value, defaultVal)
+		resolution = resolveTyped(res.Value, res.ProviderResolutionDetail, defaultVal)
 	}
 
 	if resolution.FlagMetadata == nil {
--- a/openfeature/multi/comparison_strategy.go
+++ b/openfeature/multi/comparison_strategy.go
@@ -89,8 +89,8 @@ func evaluateComparison[T FlagTypes](providers []NamedProvider, fallbackProvider
 			case int8, int16, int32, int64, int, uint8, uint16, uint32, uint64, uint, uintptr, float32, float64, string, bool:
 				break
 			default:
-				t := reflect.TypeOf(defaultValue)
-				if !t.Comparable() {
+				// a nil default has no type to compare, see #546
+				if t := reflect.TypeOf(defaultValue); t == nil || !t.Comparable() {
 					// Impossible to evaluate strategy with expected result type
 					defaultResult := BuildDefaultResult(StrategyComparison, defaultValue, ErrAggregationNotAllowed)
 					defaultResult.FlagMetadata[MetadataFallbackUsed] = false
--- a/openfeature/multi/comparison_strategy_test.go
+++ b/openfeature/multi/comparison_strategy_test.go
@@ -739,4 +739,22 @@ func Test_ComparisonStrategy_ObjectEvaluation(t *testing.T) {
 		assert.Equal(t, "none", result.FlagMetadata[MetadataSuccessfulProviderName])
 		assert.False(t, result.FlagMetadata[MetadataFallbackUsed].(bool))
 	})
+
+	t.Run("nil default value returns the default instead of panicking", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		fallback := of.NewMockFeatureProvider(ctrl)
+		provider := of.NewMockFeatureProvider(ctrl)
+		strategy := newComparisonStrategy([]NamedProvider{
+			&namedProvider{
+				name:            "test-provider1",
+				FeatureProvider: provider,
+			},
+		}, fallback, nil)
+
+		result := strategy(t.Context(), testFlag, nil, of.FlattenedContext{})
+		assert.Nil(t, result.Value)
+		assert.Equal(t, of.ErrorReason, result.Reason)
+		assert.Equal(t, of.NewGeneralResolutionError(ErrAggregationNotAllowed.Error()), result.ResolutionError)
+		assert.False(t, result.FlagMetadata[MetadataFallbackUsed].(bool))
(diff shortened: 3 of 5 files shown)


erka:
Hey @kripa-sindhu-007

I think your suggestion about TYPE_MISMATCH is a great shout-out.

openfeature/multi/strategies.go
```diff
@@ -109,6 +109,16 @@ func mergeFlagMeta(tags ...of.FlagMetadata) of.FlagMetadata {
 	}
 }
 
+// valueOrDefault returns value when it holds the expected type, and the caller's default
+// otherwise. A strategy may leave Value unset, and a nil has no type to assert, see #546
+func valueOrDefault[T any](value any, defaultValue T) T {
```
sahidvelji: Nil means no value, so the default is right. A wrong-typed value is a strategy bug and shouldn't come back wearing the provider's success reason with no error — that's undiagnosable, unlike the panic it replaces.

Prefer returning the detail so the helper owns the policy:

```go
func resolveTyped[T any](value any, detail of.ProviderResolutionDetail, defaultValue T) of.GenericResolutionDetail[T] {
	typed := of.GenericResolutionDetail[T]{Value: defaultValue, ProviderResolutionDetail: detail}
	if v, ok := value.(T); ok {
		typed.Value = v
	} else if value != nil {
		typed.ResolutionError = of.NewTypeMismatchResolutionError(
			fmt.Sprintf("strategy resolved %T, expected %T", value, defaultValue))
		typed.Reason = of.ErrorReason
	}
	return typed
}
```

Params stay decomposed — passing the whole struct won't compile at both sites (`[any]` vs `[FlagTypes]`). Since the detail types are `=` aliases, each accessor becomes `return resolveTyped(res.Value, res.ProviderResolutionDetail, defaultValue)`.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-5-opus[1m]; version: 2.1.219.*

sahidvelji (commented):
Fixes work — verified against `9853ca0`, `make test` clean under `-race`. Two asks:

1. `valueOrDefault` reports wrong-typed values as successes. Prefer `TYPE_MISMATCH` for non-nil, default for nil (inline below). That's also my answer to your open question.
2. `ObjectEvaluation` (`multiprovider.go:351`) is still unguarded — returns `nil`, no default, and `Client.ObjectValue` passes it through. Deliberate or oversight?

Nits: drop the inline `see #546` (trailers have it); missing period on the doc comment; `assert.InDelta` at `multiprovider_test.go:95` should be `assert.Equal`.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-5-opus[1m]; version: 2.1.219.*

openfeature/multi/strategies.go
kripa-sindhu-007: Applied in b38273a, as you wrote it, so the helper owns both policies: nil keeps the quiet default, a wrong type gets TYPE_MISMATCH with an ERROR reason.

Confirmed through the client, not just the provider:

    provider -> Value=true Reason="ERROR" Err="TYPE_MISMATCH: strategy resolved string, expected bool"
    client   -> Value=true Reason="ERROR" ErrorCode="TYPE_MISMATCH"

The caller still gets their default and now has something to diagnose it with. Since erka said the same above, I have taken the open question as settled and removed it from the PR body.

kripa-sindhu-007:
@sahidvelji All of it is in b38273a. make test, make lint and make e2e-test are green.

On ObjectEvaluation: oversight, not deliberate. I guarded the four typed accessors and never went back to the object one, where T is any and the assertion had been dropped entirely.

It is also quieter than "returns nil" suggests. There is no error and no reason either:

    mp.ObjectEvaluation       -> Value=<nil>  Reason=""
    client.ObjectValue        -> value=<nil>  err=<nil>
    client.ObjectValueDetails -> value=<nil>  reason=""  code=""  err=<nil>

BooleanEvaluation on the same strategy returns the default. So a caller has no way to detect this one, which puts it in the same 1.4.10 family as claims 2 and 3 rather than alongside them. It now goes through resolveTyped like the rest.

Two tests added: one for the object accessor dropping the default, one asserting the ERROR reason and TYPE_MISMATCH code on a wrong-typed value.

Nits are all in. assert.Equal at multiprovider_test.go:95, since the float is returned verbatim and there was nothing to give a tolerance to. Period on the doc comment, and the see #546 is gone.

Thanks for the review.

erka (approved):

sahidvelji (approved):

PR #580, opened 2026-09-12
fix(multi): preserve fatal and not-ready provider states

### LINKED ISSUE #544: [BUG] multiprovider: NOT_READY and FATAL are absent from the state precedence table, so they resolve to READY
## Observed behavior

The state precedence tables in `openfeature/multi/multiprovider.go` define only three states:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/multi/multiprovider.go#L119-L130

`of.NotReadyState` and `of.FatalState` are missing. `evaluateState()` does a bare `stateValues[s]` lookup, so a miss yields the zero value `0`, and `stateTable[0]` maps back to **`ReadyState`**. The two highest-precedence states in Appendix A become the lowest:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/multi/multiprovider.go#L529-L540

This is reachable on the ordinary startup path: `InitWithContext` explicitly sets every inner provider to `NotReadyState` before initializing it, which immediately flips the overall status to READY.

```
status BEFORE Init                    = "NOT_READY"
status DURING Init (inner NOT_READY)  = "READY"      <- want NOT_READY
status AFTER Init                     = "READY"
```

Separately, `FatalState` is **unrepresentable** in this package. `eventTypeToState` maps `of.ProviderError` → `of.ErrorState` unconditionally and ignores `ProviderEventDetails.ErrorCode`, unlike core's `statesMap`, which checks for `ProviderFatalCode`:

https://github.com/open-feature/go-sdk/blob/6b0824c6c7b9b69e40f26aedcc4406ad4d3a5cec/openfeature/openfeature_api.go#L518-L528

```
multi Status() after inner PROVIDER_ERROR{PROVIDER_FATAL} = "ERROR"   <- want FATAL
```

## Expected Behavior

[Appendix A](https://github.com/open-feature/spec/blob/main/specification/appendix-a-included-utilities.md#multi-provider-status) ("Multi-Provider Status") defines the precedence order as **FATAL > NOT_READY > ERROR > STALE > READY**. Both states should be present in the tables, and `eventTypeToState` should map a `PROVIDER_ERROR` carrying error code `PROVIDER_FATAL` to `FatalState`, matching core's behaviour.

This matters beyond status reporting. Appendix A: *"The only statuses which affect evaluation behavior at the SDK client level are FATAL and NOT_READY. If a provider is in either of these states, evaluation will be skipped by the client and the default value will be returned."* Since `Client.evaluate` short-circuits only on `NotReadyState`/`FatalState`, this causes the client to evaluate against a fatal or not-yet-initialized multi-provider instead of returning the default value.

Note that fixing the maps alone is insufficient — `evaluateState` maps precedence back through a fixed `[3]of.State` array, so the array and both maps need to grow together.

## Steps to reproduce

```go
// slowInitProvider.Init sleeps for `delay`, then returns nil
slow := &slowInitProvider{delay: 700 * time.Millisecond}
mp, _ := multi.NewProvider(multi.StrategyFirstMatch, multi.WithProvider("slow", slow))

go mp.InitWithContext(context.Background(), of.EvaluationContext{})
time.Sleep(200 * time.Millisecond)

fmt.Println(mp.Status()) // "READY" while the only inner provider is still NOT_READY
```

For the FATAL half, have an inner provider emit `of.Event{EventType: of.ProviderError, ProviderEventDetails: of.ProviderEventDetails{ErrorCode: of.ProviderFatalCode}}` and observe `mp.Status()` report `ERROR` rather than `FATAL`.

kripa-sindhu-007:
## This PR

  - Preserve `NOT_READY` when any inner provider is still initializing.
  - Map `PROVIDER_FATAL` events to `FATAL` instead of `ERROR`.
  - Apply the complete provider-state precedence order.
  - Add regression tests for initialization and fatal provider events.

  ### Related Issues

  Fixes #544

  ### Notes

  The previous state table only represented `READY`, `STALE`, and `ERROR`, causing `NOT_READY` and `FATAL` states to be lost during
  aggregation.

  This change preserves the full precedence order:

  `READY < STALE < ERROR < NOT_READY < FATAL`

  Only `PROVIDER_ERROR` events carrying the `PROVIDER_FATAL` error code map to `FATAL`; other provider errors retain their existing behavior.

  ### How to test

  - `make test`
  - `make lint`
  - `make e2e-test`

  All checks pass locally.

### DIFF
--- a/openfeature/multi/multiprovider.go
+++ b/openfeature/multi/multiprovider.go
@@ -103,7 +103,7 @@ func (n *namedProvider) unwrap() of.FeatureProvider {
 
 var (
 	stateValues      map[of.State]int
-	stateTable       [3]of.State
+	stateTable       [5]of.State
 	eventTypeToState map[of.EventType]of.State
 
 	// Compile-time interface compliance checks
@@ -118,15 +118,19 @@ var (
 func init() {
 	// used for mapping provider event types & provider states to comparable values for evaluation
 	stateValues = map[of.State]int{
-		of.ReadyState: 0,
-		of.StaleState: 1,
-		of.ErrorState: 2,
+		of.ReadyState:    0,
+		of.StaleState:    1,
+		of.ErrorState:    2,
+		of.NotReadyState: 3,
+		of.FatalState:    4,
 	}
 	// used for mapping
-	stateTable = [3]of.State{
-		of.ReadyState, // 0
-		of.StaleState, // 1
-		of.ErrorState, // 2
+	stateTable = [5]of.State{
+		of.ReadyState,
+		of.StaleState,
+		of.ErrorState,
+		of.NotReadyState,
+		of.FatalState,
 	}
 	eventTypeToState = map[of.EventType]of.State{
 		of.ProviderReady: of.ReadyState,
@@ -507,8 +511,16 @@ func (p *Provider) updateProviderStateFromEvent(e namedEvent) bool {
 	p.providerStatusLock.Lock()
 	previousState := p.providerStatus[e.providerName]
 	p.providerStatusLock.Unlock()
-	logProviderState(p.logger, e, previousState)
-	return p.updateProviderState(e.providerName, eventTypeToState[e.EventType])
+	state := stateFromEvent(e)
+	logProviderState(p.logger, e, state, previousState)
+	return p.updateProviderState(e.providerName, state)
+}
+
+func stateFromEvent(e namedEvent) of.State {
+	if e.EventType == of.ProviderError && e.ErrorCode == of.ProviderFatalCode {
+		return of.FatalState
+	}
+	return eventTypeToState[e.EventType]
 }
 
 // evaluateState Determines the overall state of the provider using the weights specified in Appendix A of the
@@ -524,8 +536,8 @@ func (p *Provider) evaluateState() of.State {
 	return stateTable[maxState]
 }
 
-func logProviderState(l *slog.Logger, e namedEvent, previousState of.State) {
-	switch eventTypeToState[e.EventType] {
+func logProviderState(l *slog.Logger, e namedEvent, state, previousState of.State) {
+	switch state {
 	case of.ReadyState:
 		if previousState != of.NotReadyState {
 			l.LogAttrs(context.Background(), slog.LevelInfo, "provider has returned to ready state",
@@ -538,6 +550,9 @@ func logProviderState(l *slog.Logger, e namedEvent, previousState of.State) {
 			slog.String(MetadataProviderName, e.providerName), slog.String("event-message", e.Message))
 	case of.ErrorState:
 		l.LogAttrs(context.Background(), slog.LevelError, "provider is in an error state",
+			slog.String(MetadataProviderName, e.providerName), slog.String("event-message", e.Message))
+	case of.FatalState:
+		l.LogAttrs(context.Background(), slog.LevelError, "provider is in a fatal state",
 			slog.String(MetadataProviderName, e.providerName), slog.String("event-message", e.Message))
 	}
 }
--- a/openfeature/multi/multiprovider_test.go
+++ b/openfeature/multi/multiprovider_test.go
@@ -252,6 +252,39 @@ func TestMultiProvider_Init(t *testing.T) {
 	assert.Equal(t, of.ReadyState, mp.Status())
 }
 
+func TestMultiProvider_InitReportsNotReady(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	provider := of.NewMockFeatureProvider(ctrl)
+	provider.EXPECT().Metadata().Return(of.Metadata{Name: "MockProvider"})
+	provider.EXPECT().Hooks().Return([]of.Hook{}).MinTimes(1)
+	stateHandler := of.NewMockStateHandler(ctrl)
+	started := make(chan struct{})
+	release := make(chan struct{})
+	stateHandler.EXPECT().Init(gomock.Any()).DoAndReturn(func(of.EvaluationContext) error {
+		close(started)
+		<-release
+		return nil
+	})
+	stateHandler.EXPECT().Shutdown().MaxTimes(1)
+	wrapped := struct {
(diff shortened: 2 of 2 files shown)


sahidvelji:
`TestMultiProvider_InitReportsNotReady` only covers one provider, and the guarantee doesn't hold past that. In `InitWithContext` (not in this diff) each provider is seeded to `NotReadyState` inside the same loop that launches its init goroutine, so a fast provider can reach `ReadyState` while later providers are still absent from `providerStatus` — and `evaluateState` only maxes over what's present, so `Status()` transiently reports READY.

200 no-op providers plus one blocker hit transient READY in 55/200 iterations on this branch; 0/200 with the seeding split into its own pass first:

```go
	for _, provider := range p.providers {
		p.updateProviderState(provider.Name(), of.NotReadyState)
	}

	for _, provider := range p.providers {
		// ... existing goroutine launch, unchanged
	}
```

`make test` passes with that applied. It's pre-existing and the window is narrow at realistic provider counts, but this PR is what makes NOT_READY meaningful in the precedence table. Either way the `// No locks required there are no workers running` comment above it is inaccurate.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.236.*

kripa-sindhu-007:
@sahidvelji  Confirmed, and thanks — the window is exactly where you said it is. Seeding shared the loop with the goroutine launch, so provider *n*'s worker could write `ReadyState` before provider *n+1* was seeded at all, and `evaluateState` only maxes over the keys already in `providerStatus`.

I've split the seeding into its own pass as you suggested, and dropped the `// No locks required there are no workers running` comment — it was describing a loop that does launch workers.

For the test: `TestMultiProvider_InitReportsNotReady` can't catch this with one provider, and I couldn't find a way to force the interleaving deterministically, so the new test reproduces it instead — 128 no-op providers followed by one blocking provider, with a goroutine polling `Status()` for the duration of `InitWithContext`. On `7dd763d` it fails
42 of 50 runs, every failure reporting `READY`; with the seeding split it passes 50 of 50. it can't false-fail on the fixed code, since the status can only leave `NOT_READY` once the test releases the blocker. `make test` and `make lint` are both clean.

sahidvelji:
Follow-up on the state work here rather than the seeding fix: `stateFromEvent` covers the event path, but a fatal error out of a provider's `Init` still lands as ERROR. At 4890099 a provider whose `Init` returns `&of.ProviderInitError{ErrorCode: of.ProviderFatalCode}` leaves `mp.Status()` reporting `ERROR`, so one of the two ways a provider becomes fatal doesn't produce FATAL.

Suggested:

```go
// stateFromInitError maps an error returned by a provider's initialization to the state that
// provider should be recorded in. A fatal init error yields FATAL, everything else ERROR.
func stateFromInitError(err error) of.State {
	if initErr, ok := errors.AsType[*of.ProviderInitError](err); ok && initErr.ErrorCode == of.ProviderFatalCode {
		return of.FatalState
	}
	return of.ErrorState
}
```

with two call sites in `InitWithContext`:

```diff
 				if err != nil {
 					l.LogAttrs(ctx, slog.LevelError, "initialization failed", slog.Any("error", err))
-					p.updateProviderState(name, of.ErrorState)
+					p.updateProviderState(name, stateFromInitError(err))
```

```diff
-		p.setStatus(of.ErrorState)
+		p.providerStatusLock.Lock()
+		aggregate := p.evaluateState()
+		p.providerStatusLock.Unlock()
+		p.setStatus(aggregate)
```

The second hunk is the one that's easy to miss. `errgroup` waits for every goroutine, so by the time `eg.Wait()` returns each provider has already recorded its own state and the aggregate is correct — and then `setStatus(of.ErrorState)` clobbers it back down. Both hunks are needed for FATAL to survive.

Checked with this applied: fatal init → FATAL, wrapped fatal → FATAL, non-fatal `ProviderInitError` → ERROR, plain error → ERROR, one fatal provider alongside a healthy one → FATAL.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.236.*

sahidvelji:
One more, small enough to fold in here rather than file separately: `stateFromEvent` falls through to a bare map lookup, so any event type outside the three in `eventTypeToState` yields the zero value `of.State("")`. That empty string is written into `providerStatus`, scores `stateValues[""] == 0` in `evaluateState` (i.e. READY), and fails the `== of.ReadyState` check in `Track`.

Concretely, a provider already in ERROR that emits an unrecognized event type is silently promoted:

```
after PROVIDER_ERROR:      Status() = "ERROR"
after unknown event type:  Status() = "READY"
```

Nothing in-tree emits such an event today, so this is latent rather than live — but `of.EventType` is an open string type and third-party providers are free to emit their own. It is the same class of bug as #544: a map miss scoring as READY.

Core's `stateFromEvent` defaults for exactly this reason:

```go
func stateFromEvent(event Event) State {
	if stateFn, ok := statesMap[event.EventType]; ok {
		return stateFn(event.ProviderEventDetails)
	}
	return NotReadyState // default
}
```

so the multi-provider equivalent would be:

```diff
 func stateFromEvent(e namedEvent) of.State {
 	if e.EventType == of.ProviderError && e.ErrorCode == of.ProviderFatalCode {
 		return of.FatalState
 	}
-	return eventTypeToState[e.EventType]
+	state, ok := eventTypeToState[e.EventType]
+	if !ok {
+		return of.NotReadyState
+	}
+	return state
 }
```

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.236.*

kripa-sindhu-007:
@sahidvelji Fixed both in 480058a. Fatal init errors now preserve FATAL, including wrapped errors, and unknown event types default to NOT_READY.
Added regression coverage for both; make test, make lint, and make e2e-test pass locally.

erka (approved):

sahidvelji (approved):

PR #581, opened 2026-09-13
build: manage dev tools and tasks with mise

sahidvelji:
## This PR

Proof of concept: replaces the `Makefile` with [mise](https://mise.jdx.dev) for both dev tools and tasks.

- `mise.toml` declares the toolchain — Go and golangci-lint in `[tools]`, and the single-purpose tools (mockgen, govulncheck, goreleaser, cyclonedx-gomod) on the tasks that run them — plus the tasks themselves; `mise.lock` pins exact versions and checksums
- `make test` / `e2e-test` / `lint` / `fix` / `docs` / `mockgen` become `mise run test` / `test:e2e` / `lint` / `fix` / `docs` / `mockgen`
- both workflows install tools with `jdx/mise-action` and then call the same tasks contributors run, replacing `actions/setup-go`, `golang/govulncheck-action`, `goreleaser-action` and `gh-gomod-generate-sbom`
- `CONTRIBUTING.md`, `.gitignore` and a dead `Makefile` entry in `renovate.json` updated to match

### Why mise?

Tool versions are currently spread across four places that have already drifted: Go in `go.mod`, golangci-lint and mockgen in the `Makefile`, and each CI action pinning its own. `govulncheck-action` is the clearest example — it runs on Go `stable` rather than our `go.mod` version, does a second redundant checkout, and installs `govulncheck@latest`. CI's test commands are a fifth copy that diverged from `make test` by adding coverage flags. mise collapses all of that into one file, one lockfile, and one set of task names that CI and contributors share.

It's also a safe bet rather than a novelty: ~34k stars, and ThoughtWorks moved it to **Adopt** in [Technology Radar Vol 34 (April 2026)](https://www.thoughtworks.com/radar/tools/mise), specifically for consolidating "tool and language versioning, environment variable management and task execution into a single high-performance, Rust-based tool". Several OpenFeature repos are polyglot, so a shared answer here could travel.

Nothing is locked in — the tasks are thin wrappers, so `mise.toml` doubles as documentation for anyone who'd rather run `go test` directly.

### Notes

- Coverage profiles moved into the test tasks, so CI no longer needs its own variant of the test command. `coverage-*.out` is now gitignored.
- Go's version now sits in `mise.toml` as well as `go.mod`. mise can read `go.mod`, but only its `toolchain` directive — reading the `go` directive is deprecated upstream.
- `cyclonedx-gomod` stays at v1.3.0 (matching the action it replaces) rather than jumping to v1.12.0, to keep SBOM output unchanged here.
- `mise run docs` uses `go doc -http`, which fetches pkgsite on demand, so there is no pkgsite pin to maintain.
- Tools declared on a task install the first time that task runs, so CI sets `MISE_LOCKED` rather than passing `--locked` to `mise install`; that covers task tools too. Renovate's mise manager reads `tasks.*.tools` as well as `[tools]`, so both kinds of pin stay maintained.

### How to test

`mise install && mise run ci`. `mise run mockgen` should produce no diff. The release path is verified locally with `mise exec goreleaser@2.18.1 go:github.com/CycloneDX/cyclonedx-gomod/cmd/cyclonedx-gomod@1.3.0 -- goreleaser release --snapshot --clean`, which exercises the SBOM hook finding `cyclonedx-gomod` on mise's PATH.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.236.*


mise.toml
```diff
@@ -0,0 +1,72 @@
...
+go = "1.26"
+golangci-lint = "2.13.1"
+goreleaser = "2.18.1"
+# Keep in sync with the go.uber.org/mock version in go.mod: mockgen stamps its
+# own version into the generated files.
+"go:go.uber.org/mock/mockgen" = "0.6.0"
+"go:golang.org/x/pkgsite/cmd/pkgsite" = "0.5.0"
+"go:golang.org/x/vuln/cmd/govulncheck" = "1.8.0"
+# Consumed by the goreleaser `sboms` hook, not invoked directly.
+"go:github.com/CycloneDX/cyclonedx-gomod/cmd/cyclonedx-gomod" = "1.3.0"
```
sahidvelji: Worth calling out: keeping these current is not extra maintenance work. Renovate ships a built-in [`mise` manager](https://docs.renovatebot.com/modules/manager/mise/) that is enabled by default under `config:recommended`, and its default file patterns already match `mise.toml`, so no change to `renovate.json` is needed.

It resolves short names through the mise registry and picks a datasource per backend, so every line in this block is covered:

| entry | backend | datasource |
| --- | --- | --- |
| `go` | `core` | Go releases |
| `golangci-lint`, `goreleaser` | `aqua` | GitHub tags |
| `go:...` (mockgen, pkgsite, govulncheck, cyclonedx-gomod) | `go` | proxy.golang.org |




*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.236.*

erka:
Great work @sahidvelji 

A few things:
- `mise` installs all the tools for every action, even when they aren't needed (like goreleaser for regular PR or gomock for every PR). If we could use `MISE_DISABLE_TOOLS` to reduce their number, it would be great
- could we use `go doc -http` instead of `pkgsite`? That would let us reduce the number of tools we need.

sahidvelji:
Both done in be53057.

`go doc -http` replaces pkgsite, so that's one less tool to pin.

Rather than `MISE_DISABLE_TOOLS`, I moved the single-purpose tools onto the tasks that run them, so a PR job installs only Go and golangci-lint. `MISE_DISABLE_TOOLS` didn't fit: it doesn't filter `mise install --include-task-tools`, it needs the full `go:go.uber.org/mock/mockgen`, and as a denylist it picks up any tool added later. Renovate reads `tasks.*.tools` too, so the pins stay maintained.

golangci-lint stays in `[tools]` since `lint` and `fix` both use it. CI now sets `MISE_LOCKED` so task tools stay locked; they install after mise-action's cache save, which I measured at ~2.5s cold.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.236.*

erka (approved):

Issue #596, opened 2026-09-18
[BUG] e2e gherkin suite asserts nothing: test-harness submodule points at the flagd suite, not the spec's


aepfli:
### Observed behavior

The `test-harness` git submodule exists solely so `e2e/evaluation_test.go` can read one Gherkin file:

https://github.com/open-feature/go-sdk/blob/main/e2e/evaluation_test.go#L28

```go
Paths: []string{"../test-harness/gherkin/evaluation.feature"},
```

That file is not the SDK suite. `open-feature/test-harness` was **renamed to `open-feature/flagd-testbed`** (the URL in `.gitmodules` still says `test-harness`; GitHub redirects it), and its `gherkin/` directory holds **flagd provider** scenarios:

```gherkin
@rpc @in-process @file
Feature: flagd evaluations

  Background:
    Given an option "cache" of type "CacheType" with value "disabled"
    And a stable flagd provider
```

None of those steps match go-sdk's step definitions. Because godog is not in strict mode, every scenario resolves to *undefined*, which godog reports as a pass:

```
30 scenarios (30 undefined)
205 steps (205 undefined)
--- PASS: TestEvaluation (0.05s)
ok  	github.com/open-feature/go-sdk/e2e	0.169s
```

**The e2e job is green and asserts nothing.** All ~40 step definitions in `e2e/evaluation_test.go` (788 lines) are dead code, and have been since June 2025.

### Expected Behavior

The e2e suite should run the SDK-level Gherkin, which has lived in the **spec** repo since August 2023 — [Appendix B: Gherkin Suites](https://github.com/open-feature/spec/blob/main/specification/appendix-b-gherkin-suites.md), `specification/assets/gherkin/`.

Every other SDK already does this:

| repo | submodule |
| --- | --- |
| java-sdk | `spec` -> `spec/specification/assets/gherkin` |
| js-sdk | `spec` |
| python-sdk | `spec` |
| dotnet-sdk | `spec` |
| ruby-sdk | `spec` |
| **go-sdk** | **`test-harness` (= flagd-testbed)** |
| php-sdk | `test-harness` — legitimate, its integration tests run against flagd |

java-sdk, for reference:

```java
@Suite
@IncludeEngines("cucumber")
@SelectDirectories("spec/specification/assets/gherkin")
@ExcludeTags({"deprecated", "reason-codes-cached", "async", "immutability", "evaluation-options"})
public class GherkinSpecTest {}
```

### Steps to reproduce

```
[...]

PR #598, opened 2026-09-18
chore: pin the test-harness submodule to its current release tag

aepfli:
Pins the test-harness submodule to the tag it already points at. Nothing moves.

Renovate reads `branch` in `.gitmodules` as the submodule's current version. Without it there is no version to compare against, so it can only chase the upstream default branch — see open-feature/flagd-testbed#398, and open-feature/community-tooling#126, which turns that path off entirely.

| submodule | commit | pin |
|---|---|---|
| `test-harness` | `b308c1b` | `v3.10.1` |

The symptom is open in this repo right now: #596 is an `update test-harness digest to 5022ee2` PR, pointing at testbed main rather than any release.

**Overlap:** #597 removes this submodule entirely. If that lands first, close this — pinning does not make removal harder. If this lands first, #597 drops the pin along with the submodule.

Also switches the submodule URL to the canonical `flagd-testbed`. The repo was renamed and the old URL only works through GitHub's redirect, which stops working if a repo with the old name is ever created again. Nothing moves: existing clones keep their cached URL and carry on, `git submodule sync` adopts the new one.

Refs: open-feature/community-tooling#125


erka:
@aepfli I am lost with #597. Why do we need this?

aepfli:
you are right, and there are two different changes, which are not compliant. we should use the `branch` proprty in the submodules, and this set of changes, to whom this pr belongs, was just adding the branch property everywhere with a script, where it makes sense. i will fold this into the other one - see https://github.com/open-feature/go-sdk/pull/597/changes/ded2ddde88c7135e3d0469882d366d6bec0cda4d

